PR #20, opened 2016-08-12
Fix variable bitrate detection for MP3 files

josepanguera:
Starting with Mutagen commit 65d5910c12d9a4a4e0c7de0316b494ac064828c9 the code
for detecting the VBR detection in nicotine is no longer working as it was
using a private interface.

In Mutagen commit 1ffd7dd97b145316b56ac7bd586a8ad8779d2b6d a public interface
to detect the VBR for MP3 files was introduced. Use it by default and fallback
to the old code in case an old version of Mutagen is in use.

### DIFF
--- a/pynicotine/metadata_mutagen.py
+++ b/pynicotine/metadata_mutagen.py
@@ -66,16 +66,15 @@ def processMusepack(audio):
 		"time": audio.info.length,
 	}
 def processMPEG(audio):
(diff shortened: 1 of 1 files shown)


gfarmerfr:
Good catch :)
I haven't rescan my shares for quite a while and so it wasn't trigger by last mutagen from pip for me :)

PR #44, opened 2017-02-04
Make nicotine work with FreeBSD

karolyi:
FreeBSD has python2 under /usr/local/bin, but using a /usr/bin/env sha-bang should fix this for all POSIX platforms.

### DIFF
--- a/nicotine
+++ b/nicotine
@@ -1,4 +1,4 @@
-#!/usr/bin/python2
+#!/usr/bin/env python2
 # -*- coding: utf-8 -*-
 #
 # COPYRIGHT (C) 2016-2017 Michael Labouebe <gfarmerfr@free.fr>


gfarmerfr:
Hi @karolyi,

I'm currently discussing this with our debian/ubuntu packager because they have the opposite requirement. See end of section 2.4.2 in the [Debian Python Packaging Policy](https://www.debian.org/doc/packaging-manuals/python-policy/ch-python.html#s-interpreter)

> Maintainers should not override the Debian Python interpreter using /usr/bin/env name. This is not advisable as it bypasses Debian's dependency checking and makes the package vulnerable to incomplete local installations of Python.

If we come to an agreement I will do it the [Python way](https://docs.python.org/2.7/tutorial/appendix.html#tut-scripts) so the source tree will work almost everywhere and will work out of the box with FreeBSD.

If I do that I must add a patch to the debian packaging to switch the shebang to comply to their requirement. Also the shebang of```setup.py``` should be modified too.

karolyi:
Hey @gfarmerfr, using the python way is totally fine. My approach would have been the same too, I was wondering why you specified python2 so strictly.

Should I modify the pull request that way, or will you solve it?

gfarmerfr:
I still need to learn how to create a debian quilt patch before apply the "python way" so I will solve it on my side in one commit to avoid breaking the buildbot :). No need to burden yourself with debian specific stuff :)

eLvErDe:
Hey,

There's a single binary with the bad shebang right ?
Just go for a sed in target dh_autoinstall, that s the proper way. Want me to do that tomorrow ?

Adam

On February 5, 2017 9:18:42 PM GMT+01:00, Michael Labouebe <notifications@github.com> wrote:
>I still need to learn how to create a debian quilt patch before apply
>the "python way" so I will solve it on my side in one commit to avoid
>breaking the buildbot :). No need to burden yourself with debian
>specific stuff :)
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/Nicotine-Plus/nicotine-plus/pull/44#issuecomment-277545975

-- 
Envoyé de mon téléphone Android avec K-9 Mail. Excusez la brièveté.

gfarmerfr:
Hi @eLvErDe,

There's only two scripts to modify :
```
nicotine
setup.py
```

I've gone ahead and put ```/usr/bin/env python2``` in both of them.
You can go ahead and change them to ```/usr/bin/python2``` via the ```debian/``` files when you've time.

gfarmerfr:
Hi @eLvErDe 

Did you found out a few minutes to look at how to do this via dh_autoinstall you were speaking of ?

eLvErDe:
Hi,

I just did a try:
Debhelper takes care of this automatically, no need to change anything 
it's already done in the generated binary package :-)

Adam.

gfarmerfr:
Ok great thx @eLvErDe  :)
I'll look with @kiplingw if the ppa build need any tweak.
@karolyi since the shebang change is upstreamed I'm clsing this PR.
Thx everyone.

PR #71, opened 2017-12-13
Updated polish language

mariachini:


Mutnick:
@mariachini thank you for your efforts. I can only spot check with translator but this looks good.

Along with the nicotine.po text file, there is an associated binary format nicotine.mo file which can be re-generated by running /pynicotine/languages/msgfmtall.py.  Would you like to do that and include that file with this commit?

If not, I can handle it no worries. Thanks again.

mariachini:
Added as a pull request.
I hope I did that correctly :)

Issue #99, opened 2019-11-19
Port to python3


dvzrv:
Will nicotine+ be ported to python3?

I see very little activity in the repository in the past months and would just like to highlight, that [python2 is soon EOL](https://pythonclock.org/).
This means, that many Linux distributions will drop all python2 based software shortly thereafter from their repositories.

If no port is attempted: Are there alternatives?

zniavre:
re-install python2 and make it running as usual ....

dvzrv:
> re-install python2 and make it running as usual ....

How very... helpful.

bernd-wechner:
> re-install python2 and make it running as usual ....

An option. But many of us are glad to see Python 2 go ...  and the question was if there are any alternatives. But yes the SoulseekQT client still works methinks.

http://www.soulseekqt.net/news/node/1

It's no Open Source, but it works.

The port to Python3 is a job indeed, but not a huge one (process is simple enough especially if there's a test suite that has good coverage already built, but test the existing code base, then [automate](https://docs.python.org/2/library/2to3.html) a translation, then the big job - mop up the rest, fix the bits that don't work no more no more). 

Alas I can't make time for it as I'm not using Nicotine anyhow precisely because it's seemed unmaintained and without a community supporting it for some years now ...  Living software doesn't use python 2, and won't in future. The EOL for python 2 simply catalyses EOL for a lot of unmaintained python 2 code - that is seen by some, as a sort of benefit).

thismachinechills:
I've been using this software for like 15 years and am willing to port the pynicotine portion of Nicotine+ over to Python 3. If anyone would like to participate, ping me on my personal site

lene:
I've done some work at https://github.com/lene/nicotine-plus/tree/python3. As soon as I can get the client to do something useful, e.g. connect to the server, I'll open a PR here.

dvzrv:
@lene that's great news. Thanks for your work on this!

kiplingw:
Just letting everyone know that I hope to have @lene's porting work merged as soon as possible. After that I'd like to ask everyone to please test as much as possible from the unstable PPA. If everything works properly we will release it as stable 2.0.0.

bitchkat:
[...]

PR #93, opened 2019-07-13
Various Flake8 fixes

r900:
Fixes the most annoying errors when running the source code through flake8.
All but one E999 errors are gone, I didn't look into how to fix this yet (`pynicotine/slskproto.py`, line 998).
AFAICT none of the changes should affect functionality, so far it runs nicely on my machine.

PR #106, opened 2020-02-14
Converting to Python3 (First steps)

lene:
See https://github.com/Nicotine-Plus/nicotine-plus/issues/99

This porting to Python3 turns out to be quite a big deal, and I promised I'd open a PR as soon as I see some progress, so here goes. I requested a merge to the develop branch, which may or may not be the right branch.
Major problems when porting to Python3:
* pygtk, which was based on Gtk2, is not available under Python3, so the UI needs to be ported to Gtk3.
* Handling of `bytes` has changed between Python 2 and 3 and is much stricter now.

I ported to the point that the UI shows up, looks correct to me, and the client logs in to server.slsknet.org. 
I'm opening the PR now, in this incomplete state, to get some feedback. Of course, commits are also welcome. If anyone wants to contribute code, maybe ping me first so we can avoid duplicating efforts.


kiplingw:
Thanks. Please let me know when you feel your PR is ready and conflicts are resolved.

kiplingw:
@lene, don't forget to update `debian/control` for any dependencies. Presumably several will have to be changed.

Maryse47:
> pygtk, which was based on Gtk2, is not available under Python3, so the UI needs to be ported to Gtk3.

Perhaps it's good opportunity for switching to Qt through [PyQt](https://riverbankcomputing.com/software/pyqt/intro) or [Pyside2](https://doc.qt.io/qtforpython/tutorials/qmlapp/qmlapplication.html) :smile:

kiplingw:
> > pygtk, which was based on Gtk2, is not available under Python3, so the UI needs to be ported to Gtk3.
> 
> Perhaps it's good opportunity for switching to Qt through [PyQt](https://riverbankcomputing.com/software/pyqt/intro) or [Pyside2](https://doc.qt.io/qtforpython/tutorials/qmlapp/qmlapplication.html) smile

Gtk3 is fine. There is no good reason to convert to Qt.

lene:
> > Perhaps it's good opportunity for switching to Qt through [PyQt](https://riverbankcomputing.com/software/pyqt/intro) or [Pyside2](https://doc.qt.io/qtforpython/tutorials/qmlapp/qmlapplication.html) smile
> 
> Gtk3 is fine. There is no good reason to convert to Qt.

I agree. Porting to Gtk3 is a bit of a pain, but it is nothing compared with rewriting the UI in an entirely different framework.

In the long run (i.e. probably never) I'd love to see an architecture that abstracts from the UI so that you can easily write different UI layers as you please, but currently the architecture is too entangled to realistically do that. Also, my reasons for wanting better separation between UI and core are mostly aesthetic, functionality is just fine as it is.

kiplingw:
@lene, I've just merged @r900 PR in #93. He made quite a few changes. Would you mind pulling and merging master into your branch? It looks like you've made a lot of progress and after it's merged and tested, I think we should be ready for a long awaited 1.4.3 stable release.

r900:
while i'd also rather see pyqt than gtk3 - why not see if we can make a clean(er?) separation of the logik and interface? i for one wouldn't mind being able to run it headless ;)

kiplingw:
> while i'd also rather see pyqt than gtk3 - why not see if we can make a clean(er?) separation of the logic and interface? i for one wouldn't mind being able to run it headless ;)

I don't think that's likely to happen. Most people would not have any need for running it headless. But if we had more contributors I'd be amenable to the separation in the same way Transmission has where they have a bunch of different frontends, including nCurses IIRC.

For now the energy should be focused on completing the migration to Gtk3 so we can release a 1.4.3 package on the PPA.

r900:
@kiplingw i fully agree, let's get the show on the road :)

lene:
> @lene, I've just merged @r900 PR in #93. He made quite a few changes. Would you mind pulling and merging master into your branch? It looks like you've made a lot of progress and after it's merged and tested, I think we should be ready for a long awaited 1.4.3 stable release.

I rebased on the current master, that went pretty well.

As to my progress, I'm currently stuck at the search function, which does not show any results in my version. Once I've solved that, and downloading files after that, I think the Python3 rewrite should be ready for some wider testing - which I'm sure will turn up a bunch of other issues, so maybe a stable release is still a bit in the future. Maybe merge the Python3 version into the unstable branch?

kiplingw:
> I rebased on the current master, that went pretty well.

Perfect!

> As to my progress, I'm currently stuck at the search function, which does not show any results in my version. Once I've solved that, and downloading files after that, I think the Python3 rewrite should be ready for some wider testing - which I'm sure will turn up a bunch of other issues, so maybe a stable release is still a bit in the future. Maybe merge the Python3 version into the unstable branch?

Right now GitHub is still saying the branch has conflicts, so you may have forgotten to push. Also, you need to update the source and binary dependencies in `debian/control`, as well as any sh-bangs to python3(1).

lene:
> Right now GitHub is still saying the branch has conflicts, so you may have forgotten to push. 
I made this PR on the develop branch actually, that's the cause for the conflicts. I switched the base branch to master, now it's conflicts-free.

> Also, you need to update the source and binary dependencies in `debian/control`, as well as any sh-bangs to python3(1).
Yes, thanks for reminding me. That's pretty trivial though (I think), Search and Download functionality is giving me more headache at the moment :)

docker/Dockerfile.python2
```diff
@@ -9,27 +9,52 @@
...
-RUN apt-get install -y nicotine xfce4-terminal
+RUN apt-get install -y nicotine xfce4-terminal less
 
 COPY docker/config /root/.nicotine
 
 ENV slskproto="/usr/lib/python2.7/dist-packages/pynicotine/slskproto.py"
 # run()
 RUN sed -i '/if connection in conns and len(conns\[connection\].ibuf) > 0:/a\\
 \t\t\t\t\tprint "********    ibuf: ", conns[connection].ibuf' $slskproto
+RUN sed -i '/self.process_server_input(conns\[server_socket\].ibuf)/a\\
```
kiplingw: @lene this looks great. If I understand this correctly, this is a unit test simulating communication with the server. If you get a chance, try and find a way to use this from DEP-8 as well. We have a basic stub in debian/tests/control already.

lene: > If I understand this correctly, this is a unit test simulating communication with the server. 

Sorry, no, this is humble printf-debugging. But actually I _did_ start o work on a mock server, check out https://github.com/lene/nicotine-plus/blob/test_suite/tests/mock_socket.py if you want to see what I did until now.
I'm torn between how to invest my limited time best at the moment. I'm still stuck at search (the client does not seem to receive any `FileSearchResult`s) and I can't decide whether I should do the debugging right (i.e. write a test suite and use it to zero in on the problem, which is a pretty massive time investment up front) or simple (i.e. printf debugging, which is ineffective and may well be more work in the end than writing a test suite). Whenever I have some time to work on nicotine+, I choose whatever I fell like at the moment, so I'm actually half-assing both approaches. Not super efficient but meh, whatever :)

> If you get a chance, try and find a way to use this from DEP-8 as well. We have a basic stub in debian/tests/control already.

That's nice, I had no idea you could do that! I'll check it out once I have a functional test suite. In the meantime I have set up a GitHub action to run the test suite on every push to my unfinished `test_suite` branch: https://github.com/lene/nicotine-plus/blob/test_suite/.github/workflows/pythonapp.yml

kiplingw:
Great work @lene. Looking forward to getting this merged and a new stable release out after everyone has tested it.

kiplingw:
Hey @lene. How are things coming along at your end?

pynicotine/slskmessages.py
```diff
@@ -420,40 +428,41 @@ class Login(ServerMessage):
...
+            pos, self.reason = self.getObject(message, bytes, pos)
 
         else:
-            pos, self.banner = self.getObject(message, types.StringType, pos)
+            pos, self.banner = self.getObject(message, bytes, pos)
         if len(message[pos:]) > 0:
             try:
                 import socket
-                pos, self.ip = pos+4, socket.inet_ntoa(self.strrev(message[pos:pos+4]))
+                pos, self.ip = pos+4, socket.inet_ntoa(message[pos+4:pos:-1])
```
toofar, line 456: I don't think these do quite the same thing:
```
> "1234567890"[6:2:-1]
'7654'
> 1234567890"[2:6][::-1]
'6543'

File: pynicotine/slskproto.py [hunk 3]
```diff
@@ -108,32 +113,33 @@ def __init__(self, conn = None, addr = None, ibuf = "", obuf = "", init = None):
...
 		self.sentbytes2 = 0
 		self.readbytes2 = 0
 
 
 class PeerConnectionInProgress:
 	""" As all p2p connect()s are non-blocking, this class is used to
-	hold data about a connection that is not yet established. msgObj is 
-	a message to be sent after the connection has been established.
+	hold data about a connection that is not yet established. msgObj is
+	a message "".join("%02x" % b for b in conns[i].obuf)to be sent after the connection has been established.
```
toofar, line 125: oops?

kiplingw:
Thanks @toofar. Looks like @lene is getting closer but still a few lose ends to clean up.

lene:
> Hey @lene. How are things coming along at your end?

I got stuck at debugging search results (either the server doesn't distribute the search query to other clients, or the other clients don't respond to it, I cannot determine which). Then life happened. I'll try to get back to it and hopefully find out more this or next week.

kiplingw:
Thanks @lene. Looking forward to merging your amazing PR.

pynicotine/slskproto.py
```diff
@@ -310,14 +316,14 @@ def __init__(self, ui_callback, queue, bindip, port, config, eventprocessor):
 			long = _("The range you specified for client connection ports was %(low)s-%(high)s, but none of these were usable. Increase and/or move the range and restart Nicotine+.") % {'low':portrange[0], 'high':portrange[1]}
 			if  portrange[0] < 1024:
 				long += "\n\n" + _("Note that part of your range lies below 1024, this is usually not allowed on most operating systems with the exception of Windows.")
-			self._ui_callback([PopupMessage(short, long)])
```
toofar, line 313: `long` is a string variable here, not the type

toofar:
Turns out that the issue of user info, user browse and search not working where due to string -> byte changes, so they appear to be working for me now. Also found that string.join() was used in places like "download containing folder". There is still one other "can't cat str and byte" error I've seen. Still trying to reproduce. Also the "Group by Users" and "Enable filters" check boxes on the search results page cause exceptions to be logged.

```patch
commit 18f421c82c9e972ea27769af7b0416c3c7e21174 (HEAD -> pr/106)
Author: Jimmy <jimmy@spalge.com>
Date:   Sat Apr 18 15:52:37 2020 +1200

    Don't encode paths in file transfer list as bytes.
    
    This is used to localise file paths and names in gtkgui/transferlist.py
    before adding them to the tree model, which expects strings.

diff --git a/pynicotine/transfers.py b/pynicotine/transfers.py
index 969e9e428fd1..cf4c957e1504 100644
--- a/pynicotine/transfers.py
+++ b/pynicotine/transfers.py
@@ -1946,8 +1946,4 @@ class Transfers:
         if user and user in config["server"]["userencoding"]:
             coding = config["server"]["userencoding"][user]
 
-        string = self.eventprocessor.decode(string, coding)
-        try:
-            return string.encode(locale.nl_langinfo(locale.CODESET))
-        except:
-            return string
+        self.eventprocessor.decode(string, coding)

commit 646b7572a42733636841bac0418b77f8f080d260
Author: Jimmy <jimmy@spalge.com>
Date:   Sat Apr 18 15:07:37 2020 +1200

    string.join() is gone in python3

diff --git a/pynicotine/gtkgui/search.py b/pynicotine/gtkgui/search.py
index 33bc9332c4b1..95cd09842611 100644
--- a/pynicotine/gtkgui/search.py
+++ b/pynicotine/gtkgui/search.py
@@ -37,7 +37,6 @@ import os
 import re
 import sre_constants
 import locale
-import string
 import random
 import _thread
 
@@ -1591,7 +1590,7 @@ class Search:
         for i in self.selected_results:
 
             user = i[0]
-            dir = string.join(i[1].split("\\")[:-1], "\\")
+            dir = "\\".join(i[1].split("\\")[:-1])
 
             if (user, dir) in folders:
                 continue
@@ -1618,7 +1617,7 @@ class Search:
         for i in self.selected_results:
 
             user = i[0]
-            dir = string.join(i[1].split("\\")[:-1], "\\")
+            dir = "\\".join(i[1].split("\\")[:-1])
 
             if (user, dir) in folders:
                 continue
@@ -1642,7 +1641,7 @@ class Search:
     def OnCopyDirURL(self, widget):
 
         user, path = self.selected_results[0][:2]
-        path = string.join(path.split("\\")[:-1], "\\") + "\\"
+        path = "\\".join(path.split("\\")[:-1]) + "\\"
 
         if path[:-1] != "/":
             path += "/"
diff --git a/pynicotine/gtkgui/transferlist.py b/pynicotine/gtkgui/transferlist.py
index 6b7e01a7773e..1f06aa565d74 100644
--- a/pynicotine/gtkgui/transferlist.py
+++ b/pynicotine/gtkgui/transferlist.py
@@ -32,7 +32,6 @@ gi.require_version('Gtk', '3.0')
 from gi.repository import Gtk as gtk
 from gi.repository import GObject as gobject
 
-import string
 from time import time
 from math import ceil
 from .utils import InitialiseColumns, int_sort_func, float_sort_func, HumanSize, HumanSpeed
@@ -590,7 +589,7 @@ class TransferList:
     def OnCopyDirURL(self, widget):
 
         i = self.selected_transfers[0]
-        path = string.join(i.filename.split("\\")[:-1], "\\") + "\\"
+        path = "\\".join(i.filename.split("\\")[:-1]) + "\\"
 
         if path[:-1] != "/":
             path += "/"
diff --git a/pynicotine/gtkgui/utils.py b/pynicotine/gtkgui/utils.py
index 2dfa7c0329ab..f1be1e830009 100644
--- a/pynicotine/gtkgui/utils.py
+++ b/pynicotine/gtkgui/utils.py
@@ -37,7 +37,6 @@ import time
 import locale
 import os
 import sys
-import string
 import re
 import types
 import urllib.request, urllib.parse, urllib.error
@@ -1422,7 +1421,7 @@ def _expand_alias(aliases, cmd):
                         last = int(args[1])
                     else:
                         last = len(cmd)
-                v = string.join(cmd[first:last+1])
+                v = " ".join(cmd[first:last+1])
                 if not v:
                     v = default
                 ret = ret + v
@@ -1435,7 +1434,7 @@ def _expand_alias(aliases, cmd):
                 i = i + len(arg) + 3
                 for j in range(len(cmd)-1, -1, -1):
                     arg = arg.replace("$%i" % j, cmd[j])
-                arg = arg.replace("$@", string.join(cmd[1:], " "))
+                arg = arg.replace("$@", " ".join(cmd[1:]))
 
                 import subprocess
 

commit edde69e1d12818ed083affe477c21e82da9f8cfd
Author: Jimmy <jimmy@spalge.com>
Date:   Sat Apr 18 14:55:44 2020 +1200

    Don't ord() results of bytearray.__getitem__()
    
    socket.recv() returns bytes, lets stick to dealing with them until we
    know we have something user presentable.
    
    * when manipulating connection.ibuf or something that will get added to
      it always use bytes
    * bytearray[0] etc returns an int, not a byte or a string, so no need to
      ord() it, I have verified most of these (on my machine) by running
      into them and seeing exceptions printed to stdout. I doubt I have
      excersized them all though.

diff --git a/pynicotine/slskmessages.py b/pynicotine/slskmessages.py
index f0522b1265c3..c839ffea52ed 100644
--- a/pynicotine/slskmessages.py
+++ b/pynicotine/slskmessages.py
@@ -444,7 +444,7 @@ class Login(ServerMessage):
         return message
 
     def parseNetworkMessage(self, message):
-        pos, self.success = 1, (ord(message[0]) if isinstance(message[0], str) else message[0])
+        pos, self.success = 1, message[0]
         if not self.success:
             pos, self.reason = self.getObject(message, bytes, pos)
 
@@ -573,7 +573,7 @@ class GetUserStatus(ServerMessage):
         pos, self.status = self.getObject(message, int, pos)
         # Exception handler is for Soulfind compatibility
         try:
-            pos, self.privileged = pos+1, ord(message[pos])
+            pos, self.privileged = pos+1, message[pos]
         except:
             pass
 
@@ -1689,11 +1689,11 @@ class UserInfoReply(PeerMessage):
     def parseNetworkMessage(self, message):
         pos, self.descr = self.getObject(message, bytes)
         pos, self.has_pic = pos+1, message[pos]
-        if ord(self.has_pic):
+        if self.has_pic:
             pos, self.pic = self.getObject(message, bytes, pos)
         pos, self.totalupl = self.getObject(message, int, pos)
         pos, self.queuesize = self.getObject(message, int, pos)
-        pos, self.slotsavail = pos+1, ord(message[pos])
+        pos, self.slotsavail = pos+1, message[pos]
 
         if len(message[pos:]) >= 4:
             pos, self.uploadallowed = self.getObject(message, int, pos)
@@ -1738,7 +1738,7 @@ class SharedFileList(PeerMessage):
             pos, nfiles = self.getObject(message, int, pos)
             files = []
             for j in range(nfiles):
-                pos, code = pos+1, ord(message[pos])
+                pos, code = pos+1, message[pos]
                 pos, name = self.getObject(message, bytes, pos)
                 pos, size = self.getObject(message, sizetype, pos, getsignedint=True, printerror=False)
                 if message[pos-1] == '\xff':
@@ -1832,7 +1832,7 @@ class FileSearchResult(PeerMessage):
         self.pos, nfiles = self.getObject(message, int, self.pos)
         shares = []
         for i in range(nfiles):
-            self.pos, code = self.pos+1, ord(message[self.pos])
+            self.pos, code = self.pos+1, message[self.pos]
             self.pos, name = self.getObject(message, bytes, self.pos)
             # suppressing errors with unpacking, can be caused by incorrect sizetype
             self.pos, size = self.getObject(message, sizetype, self.pos, printerror=False)
@@ -1846,7 +1846,7 @@ class FileSearchResult(PeerMessage):
                     attrs.append(attr)
             shares.append([code, name, size, ext, attrs])
         self.list = shares
-        self.pos, self.freeulslots = self.pos+1, ord(message[self.pos])
+        self.pos, self.freeulslots = self.pos+1, message[self.pos]
         self.pos, self.ulspeed = self.getObject(message, int, self.pos, getsignedint=1)
         self.pos, self.inqueue = self.getObject(message, int, self.pos)
 
@@ -1930,7 +1930,7 @@ class FolderContentsResponse(PeerMessage):
                 pos, nfiles = self.getObject(message, int, pos)
                 shares[folder][directory] = []
                 for j in range(nfiles):
-                    pos, code = pos+1, ord(message[pos])
+                    pos, code = pos+1, message[pos]
                     pos, name = self.getObject(message, bytes, pos, printerror=False)
                     pos, size = self.getObject(message, sizetype, pos, getsignedint=1, printerror=False)
                     pos, ext = self.getObject(message, bytes, pos, printerror=False)
@@ -2000,7 +2000,7 @@ class TransferResponse(PeerMessage):
 
     def parseNetworkMessage(self, message):
         pos, self.req = self.getObject(message, int)
-        pos, self.allowed = pos+1, ord(message[pos])
+        pos, self.allowed = pos+1, message[pos]
         if message[pos:] != "":
             if self.allowed:
                 pos, self.filesize = self.getObject(message, int, pos)
@@ -2207,7 +2207,7 @@ class SearchRequest(ServerMessage):
         pass
 
     def parseNetworkMessage(self, message):
-        pos, self.code = 1, ord(message[0])
+        pos, self.code = 1, message[0]
         pos, self.something = self.getObject(message, int, pos)
         pos, self.user = self.getObject(message, bytes, pos)
         pos, self.searchid = self.getObject(message, int, pos)
@@ -2225,7 +2225,7 @@ class UserPrivileged(ServerMessage):
 
     def parseNetworkMessage(self, message):
         pos, self.user = self.getObject(message, bytes, pos)
-        pos, self.privileged = pos+1, bool(ord(message[pos]))
+        pos, self.privileged = pos+1, bool(message[pos])
 
 
 class GivePrivileges(ServerMessage):
diff --git a/pynicotine/slskproto.py b/pynicotine/slskproto.py
index 4ead77238255..83a0eb6ff3af 100644
--- a/pynicotine/slskproto.py
+++ b/pynicotine/slskproto.py
@@ -791,7 +791,7 @@ class SlskProtoThread(threading.Thread):
 				break
 			elif conn.init is None:
 				# Unpack Peer Connections
-				if msgBuffer[4] == chr(0):
+				if msgBuffer[4] == 0:
 					msg = PierceFireWall(conn)
 					try:
 						msg.parseNetworkMessage(msgBuffer[5:msgsize+4])
@@ -800,7 +800,7 @@ class SlskProtoThread(threading.Thread):
 					else:
 						conn.piercefw = msg
 						msgs.append(msg)
-				elif msgBuffer[4] == chr(1):
+				elif msgBuffer[4] == 1:
 					msg = PeerInit(conn)
 					try:
 						msg.parseNetworkMessage(msgBuffer[5:msgsize+4])
@@ -810,7 +810,7 @@ class SlskProtoThread(threading.Thread):
 						conn.init = msg
 						msgs.append(msg)
 				elif conn.piercefw is None:
-					msgs.append(_("Unknown peer init code: %(type)i, message contents %(msgBuffer)s") %{'type':ord(msgBuffer[4]), 'msgBuffer':msgBuffer[5:msgsize+4].__repr__()})
+					msgs.append(_("Unknown peer init code: %(type)i, message contents %(msgBuffer)s") %{'type':msgBuffer[4], 'msgBuffer':msgBuffer[5:msgsize+4].__repr__()})
 					conn.conn.close()
 					self._ui_callback([ConnClose(conn.conn, conn.addr)])
 					conn.conn = None
@@ -870,7 +870,7 @@ class SlskProtoThread(threading.Thread):
 			if msgsize >= 0:
 				msgBuffer = msgBuffer[msgsize+4:]
 			else:
-				msgBuffer = ""
+				msgBuffer = b""
 		conn.ibuf = msgBuffer
 		return msgs, conn
 
@@ -885,7 +885,7 @@ class SlskProtoThread(threading.Thread):
 			msgsize = struct.unpack("<i", msgBuffer[:4])[0]
 			if msgsize + 4 > len(msgBuffer):
 				break
-			msgtype = ord(msgBuffer[4])
+			msgtype = msgBuffer[4]
 			if msgtype in self.distribclasses:
 				msg = self.distribclasses[msgtype](conn)
 				msg.parseNetworkMessage(msgBuffer[5:msgsize+4])
@@ -899,7 +899,7 @@ class SlskProtoThread(threading.Thread):
 			if msgsize >= 0:
 				msgBuffer = msgBuffer[msgsize+4:]
 			else:
-				msgBuffer = ""
+				msgBuffer = b""
 		conn.ibuf = msgBuffer
 		return msgs, conn
 

commit dd9adb2ed0414625308fee4355e52d5d70d3a36b
Author: Jimmy <jimmy@spalge.com>
Date:   Sat Apr 18 14:52:32 2020 +1200

    Return search results as strings.
    
    Userencode is used for localising filepath details in search.py, the
    relevant column types are all str, so don't return bytearrays.

diff --git a/pynicotine/pynicotine.py b/pynicotine/pynicotine.py
index 9ad93d593bd1..a3c59b986279 100644
--- a/pynicotine/pynicotine.py
+++ b/pynicotine/pynicotine.py
@@ -424,10 +424,7 @@ class NetworkEventProcessor:
             coding = config["server"]["userencoding"][user]
 
         string = self.decode(string, coding)
-        try:
-            return string.encode(locale.nl_langinfo(locale.CODESET))
-        except:
-            return string
+        return string
 
     def encode(self, str, networkenc=None):
 

```

kiplingw:
> Turns out that the issue of user info, user browse and search not working where due to string -> byte changes, so they appear to be working for me now. 

Well done @toofar.

lene:
Hi @toofar, thanks so much! If that gets search, browse and user info to work it is really exciting! Can I trouble you to open a PR with your changes against https://github.com/lene/nicotine-plus/tree/python3? That would make merging these changes into that branch easier.

pynicotine/slskproto.py
lene: Oops indeed!

pynicotine/slskmessages.py
```diff
@@ -420,40 +428,41 @@ class Login(ServerMessage):
...
     def __init__(self, username=None, passwd=None, version=None):
+        # TODO: use _str(). also needs to be changed downstream, so defer change for now.
         self.username = username
         self.passwd = passwd
         self.version = version
         self.ip = None
 
-    def makeNetworkMessage(self):
+    def __repr__(self):
+        return f"Login({self.username}, {self.passwd}, {self.version}, {self.ip})"
```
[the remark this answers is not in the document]
lene: You're absolutely right, I'm using this for printf debugging with a throwaway account where it is useful to verify the password is correct. Will remove it as soon as possible.

pynicotine/slskmessages.py
```diff
@@ -301,21 +313,21 @@ def getObject(self, message, type, start=0, getintasshort=0, getsignedint=0, pri
...
-            if type is types.IntType:
+            if type is int:
                 if getintasshort:
                     return intsize+start, struct.unpack("<H", message[start:start+struct.calcsize("<H")])[0]
                 elif getsignedint:
                     return intsize+start, struct.unpack("<i", message[start:start+intsize])[0]
                 else:
                     return intsize+start, struct.unpack("<I", message[start:start+intsize])[0]
-            elif type is types.LongType:
+            elif type is int:
```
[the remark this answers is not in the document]
lene: Well, dang. I'm not sure how to handle this as I have no idea what the protocol expects - I don't think the SLSK protocol is documented anywhere. So it is really hard for me to determine the original intention here.
However, since I only find calls to `getObject()` which request to get a long int made with `NetworkLongLongType` as `type` (I'll need to get rid of variables which are named the same as builtins btw, but I'll defer that to a different PR since this one is big and comlex enough already) I'm going to delete this block. `NetworkLongLongType` is handled below.

pynicotine/slskmessages.py
```diff
@@ -325,27 +337,25 @@ def getObject(self, message, type, start=0, getintasshort=0, getsignedint=0, pri
...
             # return start, None
 
     def packObject(self, object):
         """ Returns object (integer, long or string packed into a
         binary array."""
-        if type(object) is types.IntType:
+        if type(object) is int:
             return struct.pack("<i", object)
-        elif type(object) is types.LongType:
+        elif type(object) is int:
```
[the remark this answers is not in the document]
lene: I'm using `object.bit_length()` here to determine which should be used. Potentially there are problems with that approach since there can be `long` values which occupy less than 33 bit, but I don't see how else this can be done.

pynicotine/slskmessages.py
lene: Oops, indeed. Fixed in b0783e1.

lene:
@toofar, ignore the message above please, I was too impatient to continue and already added your changes to my branch manually.
They do not work yet so I have to put in another debugging session, but I think the changes in itself are solid. Thanks again!

lene:
Big thanks to @toofar and @droserasprout for your help. As of now, searching, user browse, downloading and user info work (for me anyway, testing and feedback would be welcome). I'd say the Python3 port is alpha quality now - it works as a Soulseek client, but there are still a lot of bugs. 

I've opened https://github.com/lene/nicotine-plus/issues/27, https://github.com/lene/nicotine-plus/issues/28, https://github.com/lene/nicotine-plus/issues/29 and https://github.com/lene/nicotine-plus/issues/30 for the bugs remaining from @toofar's list. https://github.com/lene/nicotine-plus/issues/18, https://github.com/lene/nicotine-plus/issues/25 and https://github.com/lene/nicotine-plus/issues/26 are also still on my radar. 

I'm not sure about the `docker` folder - it is very useful to me while developing, and it may be to others, that's why I'd leave it in there for now, but eventually it has to go, I suppose.

If anyone would like to help: feedback, testing and documenting bugs is now most important.

kiplingw:
Thanks @lene. In order to build a package for the unstable PPA for user testing, you have to update the dependencies. See `debian/control`.

kiplingw:
Accidentally closed ticket, but re-opened it. @lene, please see my inline comments. Thanks.

kiplingw:
Thanks @lene. Well done. 

What I recommend you explore next is DEP-8 continuous integration testing which is the industry standard across Debian based distros. There is currently a generic stub in `debian/tests/control`. If you add more tests in there then you can test them in `autopkgtest`. Like with the DPM, there is a bit of a learning curve, but it pays dividends later.

droserasprout:
Despite Nicotine does not use semantic versioning (which is sad) IMO changing major Python version and switching from PyGTK based GTK+2 interface to PyGObject based GTK+3 one does not fit to patch version. Shouldn't we bump to 2.0.0 with this MR merged?

kiplingw:
> Despite Nicotine does not use semantic versioning (which is sad) IMO changing major Python version and switching from PyGTK based GTK+2 interface to PyGObject based GTK+3 one does not fit to patch version. Shouldn't we bump to 2.0.0 with this MR merged?

Actually I agree.

lene:
@kiplingw, my branch is basically finished (except for one open issue, which at the moment  I cannot reproduce). I'd be happy to merge, the only open question is, what should the version number be? I think it's up to you as the maintainer to decide that :)

kiplingw:
Thanks @lene. I'd suggest keeping it at 1.4.3, and after a few weeks of user testing from the PPA we can bump it to 2.0.0. Also please make sure to add appropriate CI testing in `debian/tests/control`. If you need help or ideas, consider asking in `#debian-ci` or `#packaging` on OFTC via IRC. Once you've done that I will go through your PR with a fine tooth comb and hopefully sign off on it shortly.

kiplingw:
@lene, I asked around on OFTC in `#packaging` (`pabs` is very helpful). He recommended taking a look at [dogtail](https://gitlab.com/dogtail/dogtail), [gtk-test](https://github.com/gtk-rs/gtk-test), or the [built in](https://developer.gnome.org/gtk3/stable/gtk3-Testing.html) functions.

emilazy:
I'm excited to see this merged! I've successfully updated the NixOS package to use this branch locally and would like to send it upstream for incorporation into NixOS unstable as soon as possible; it'd be great to reduce the number of packages we have stuck on Python 2 / GTK 2 and help get more testing done.

Note that the pynotify dependency is tied to GTK 2, so will have to be dropped; [notify2](https://pypi.org/project/notify2/) looks like a pure-Python replacement with a similar API that uses the DBus notifications API directly.

kiplingw:
Thanks @emilazy. We should have it merged soon. We just need to get DEP-8 CI running. After that I will review the commits in detail.

toofar:
> add appropriate CI testing in debian/tests/control
> taking a look at dogtail, gtk-test, or the built in functions

@kiplingw, maybe it is already clear to @lene but it is not to me. What should an appropriate level of testing look like in this case? The current one just checks that it can print the version, but you have marked it as superficial. If that passing isn't enough to give you confidence in this change what would you like it to do? Run the program and open the about and settings windows or something? The autopkgtest readme doesn't say anything about what kind of graphical environment to expect but there is probably some way to make it work.

kiplingw:
@toofar, I'd suggest opening the UI, clicking some buttons, and closing it and verifying it can do all of that without crashing. Network connectivity won't be an option because most DEP-8 test environments don't have full network connectivity.

lene:
Okay, so my take on DEP-8 CI is this:
* It is a completely new feature that was not present in 1.4.2, so IMO it should not be part of this PR (though it's obviously very desirable to have).
* It is a field that's entirely new to me, and it entails quite a beast of a learning curve, so if I am to implement it, it will take its time.
* I checked out the links you gave above. Native Gtk testing is deprecated since Gtk 3.20, so it's not recommendable to use. gtk-test is written in and requires Rust, which I don't speak and would not recommend to add to nicotine+ as a dependency. dogtail is written in Python and looks quite usable, but its documentation consists of "UTSL" and little more - I haven't even gotten their most basic examples to work (see also https://gitlab.com/dogtail/dogtail/-/issues/6). So it would need some work too. Further search, while not exhaustive, has led me nowhere interesting yet. 

So, I'm happy to look into this further, but it would be really helpful if somebody with experience in the topic could assist me or take over. If it's up to me I'll probably go down the dogtail route, or write my own basic test framework if I cannot get dogtail to work.

emilazy:
I personally think that gating the merge on elaborate CI is might be a mistake; while it'd be great to have in general, the codebase is practically untested as it is and having core dependencies on unmaintained software like Python 2 and GTK+ 2 feels like a much bigger liability to me than having to manually test some basic startup tasks before release.

I'm worried that stalling on merging this PR until it's perfectly polished and brings the codebase and packaging in line with modern standards will have the opposite effect of leading to development languishing, since probably nobody is going to want to work on cleaning up things further while it still depends on an obsolete language version and toolkit and there's only been a couple handful of commits on master since 2017.

kiplingw:
> Okay, so my take on DEP-8 CI is this:
> 
>     * It is a completely new feature that was not present in 1.4.2, so IMO it should not be part of this PR (though it's obviously very desirable to have).

Actually I don't agree with you there. While I do agree this wasn't a past practice with this project, the best practice is any commits that make substantial changes to the code base should undergo CI and unit testing. In this case we never had any meaningful CI testing before and what we're trying to do now is simply make sure the newly ported Gtk+3 GUI opens and closes as expected.

>     * It is a field that's entirely new to me, and it entails quite a beast of a learning curve, so if I am to implement it, it will take its time.

You'd be surprised. Most DEP-8 scripts are not long and may only be a few lines. Others aren't scripts but single inline commands in `debian/tests/control`. Although it's new to you, you'll find it's the only way to do CI properly in a hermetically sealed chroot container that is guaranteed to closely resemble what a freshly installed distro would look like.

>    dogtail is written in Python and looks quite usable, but its documentation consists of "UTSL" and little more - I haven't even gotten their most basic examples to work (see also https://gitlab.com/dogtail/dogtail/-/issues/6). So it would need some work too. Further search, while not exhaustive, has led me nowhere interesting yet.

I think dogtail is probably your best bet. If you have issues with it I'll try to help you as soon as I have time. Don't forget to add it to the dependency list in `debian/tests/control`. Remember that we're not married to dogtail, per se. We just want some way of verifying the GUI loads and no exceptions thrown at runtime.

kiplingw:
> I personally think that gating the merge on elaborate CI is might be a mistake; while it'd be great to have in general, the codebase is practically untested as it is and having core dependencies on unmaintained software like Python 2 and GTK+ 2 feels like a much bigger liability to me than having to manually test some basic startup tasks before release.

In my experience the best time to write unit and CI tests is as the code base is being modified. You are correct that this was not done in the past with this project, but that's a policy I am changing going forward. It will make everyone's life easier.

> I'm worried that stalling on merging this PR until it's perfectly polished and brings the codebase and packaging in line with modern standards will have the opposite effect of leading to development languishing, since probably nobody is going to want to work on cleaning up things further while it still depends on an obsolete language version and toolkit and there's only been a couple handful of commits on master since 2017.

I wouldn't worry. If "stalling" means another week or two, it's not an issue. Even if I merged his PR today, we wouldn't be releasing a stable release until we've had at least a few weeks of user testing on the unstable PPA anyways.

lene:
I have opened https://github.com/lene/nicotine-plus/pull/62 to work on CI and UI tests - if anybody wants to help, please have a look at that PR so we can avoid duplicating work.

> Even if I merged his PR today

As I said before in a different thread, its "her", not "his". Please take note.

kiplingw:
@lene, please open a separate PR here in the main repository for DEP-8 testing.

lene:
@kiplingw Here you go: https://github.com/Nicotine-Plus/nicotine-plus/pull/121

kiplingw:
> @kiplingw Here you go: #121

Thank you.

brownbath:
I have no experience whatsoever to contribute on this project but I wanted to thank you all for your work. I use Nicotine daily and look forward to its integration with 20.04.

debian/changelog
```diff
@@ -1,6 +1,5 @@
-nicotine (1.4.2-1) eoan; urgency=medium
+nicotine (1.4.3) eoan; urgency=medium
```
kiplingw, line 1: Please don't modify the previous package changelog entry. Please add a new one via `dch -i`. The package version should probably be `1.4.3-1` The trailing 1 is the version of the packaging itself. So if we modify the Debian metadata or scripts in some way without modifying the "upstream" source, then we usually bump the package version.

pynicotine/gtkgui/about.py
```diff
@@ -57,24 +62,22 @@ def __init__(self, parent, nicotine):
 
         self.nicotine = nicotine
 
-        gtk.Dialog.__init__(self, _("About Nicotine+"), parent, gtk.DIALOG_MODAL)
+        gtk.Dialog.__init__(self, _("About Nicotine+"), parent, gtk.DialogFlags.MODAL)
```
kiplingw, line 65: @lene, feel free to add yourself as Gtk+ 3 / Python 3 porter or something similar. Your contribution deserves recognition.

pynicotine/gtkgui/settingswindow.py
```diff
@@ -144,7 +159,7 @@ def SetSettings(self, config):
 
         # We need to check if the frame has the upnppossible attribute
         # If UPnP port mapping is wanted, OnFirstConnect has been called
-        # and the attibute is set.
+        # and the attribute is set.
```
kiplingw, line 162: Good eye.

pynicotine/gtkgui/userlist.py
```diff
@@ -136,57 +151,36 @@ def __init__(self, frame):
         self.trusted = []
 
         for user in self.frame.np.config.sections["server"]["userlist"]:
+            username, comment, notify, privileged, trusted, last_seen, flag = user
```
kiplingw, line 154: Much cleaner. Well done.

kiplingw:
> I have no experience whatsoever to contribute on this project but I wanted to thank you all for your work. I use Nicotine daily and look forward to its integration with 20.04.

My pleasure @brownbath. We will hopefully merge this within the next few days. Make sure to add the unstable PPA if you haven't already. If the build server manages to build the package, you will see it automatically injected into the repository.

debian/changelog
lene: I used `dch`, now there is the value UNRELEASED in the distribution field, which I don't think is intended. But I don't think eoan is the correct value for it either, right? I leave it up to the you as the maintainer to decide the value.

lene:
> Almost ready @lene. If you can address the minor issues above I will merge.

Nice! Are you aware though that the CI in https://github.com/Nicotine-Plus/nicotine-plus/pull/122 is not fully working yet? I've been banging my head against the wall about getting dogtail to run within CI tests, without success yet.

emilazy:
For what it's worth, I tried getting the integration tests running in the build sandbox for NixOS, which should be a similar situation; I didn't quite manage to get it working yet due to NixOS-specific weirdness, but something like this should be close to working in a more normal CI environment:

```sh
xvfb-run -s '-screen 0 800x600x24' dbus-run-session \
  sh -c '
    at-spi-bus-launcher --launch-immediately &&
    python tests/integration/run_app.py nicotine
  '
```

kiplingw:
> For what it's worth, I tried getting the integration tests running in the build sandbox for NixOS, which should be a similar situation; I didn't quite manage to get it working yet due to NixOS-specific weirdness, but something like this should be close to working in a more normal CI environment:
> 
> ```shell
> xvfb-run -s '-screen 0 800x600x24' dbus-run-session \
>   sh -c '
>     at-spi-bus-launcher --launch-immediately &&
>     python tests/integration/run_app.py nicotine
>   '
> ```

@emilazy, can you show us the full steps you took with `autopkgtest(1)` in order to run the CI tests?

debian/changelog
kiplingw: It's ok, `eoan` or actually better `focal` is fine since the PPA's build recipe will automatically substitute it for all of the different releases (e.g. disco, eoan, focal, etc.).

emilazy:
I don't run Debian; the shell snippet I gave is an adaptation of an excerpt of the `checkPhase` for the updated `nicotine-plus` [Nix](https://nixos.org/nix/) derivation I'm working on. You'd know better than me how to try and integrate it with the Debian CI system, but I presume it would involve tweaking the [`Test-Command`](https://github.com/Nicotine-Plus/nicotine-plus/blob/75af3c7f39ac90241fdb1e58884b8f29a441968f/debian/tests/control#L16). I also haven't fully gotten it working (due to NixOS-specific details of the `at-spi2-core` packaging), so it might not work out of the box; it's just probably closer than what's there currently, as the accessibility bus has to be started for dogtail to be able to work.

kiplingw:
Unfortunately I don't know much about NixOS. But there may well be something that implements DEP-8 on it. On Debian based systems it's autopkgtest, but there may be some other effort underway for your distro.

kiplingw:
@lene, the addition you made to `pynicotine/utils.py` is causing a syntax error on my machine with Python 3.7.5.

```
$ ./debian/rules get-orig-source && debuild -S -sa
I: nicotine_1.4.3
# Preparing source for nicotine v1.4.3...
cd /home/kip/Projects/Nicotine+ \
&& python setup.py sdist --formats=gztar --dist-dir=../ \
&& mv -v ../nicotine-1.4.3.tar.gz ../nicotine_1.4.3.orig.tar.gz
  File "setup.py", line 124, in <module>
    from pynicotine.utils import version
  File "/home/kip/Projects/Nicotine+/pynicotine/utils.py", line 420
    print('*' * 8, *[arg[:200] if isinstance(arg, str) else arg for arg in args])
                   ^
SyntaxError: invalid syntax
```

kiplingw:
@emilazy, please don't prepare tarballs for your distro yet. @lene's PR is still very experimental and there are many things that still need to be fixed.

kiplingw:
@lene, it looks like the reason why the latter two DEP-8 tests are not being run is because the source tarball is not shipping the `tests` directory. Take a look:

Generate tarball:
`$ ./debian/rules get-orig-source`

Generate source package:
`$ debuild -S -sa`

Run CI tests in test bed:
```
autopkgtest ../nicotine_1.4.3-1.dsc
...
autopkgtest [16:26:13]: test unit-tests: pytest-3 tests/unit
autopkgtest [16:26:13]: test unit-tests: [-----------------------
ERROR: file not found: tests/unit
```

emilazy:
> Unfortunately I don't know much about NixOS. But there may well be something that implements DEP-8 on it. On Debian based systems it's autopkgtest, but there may be some other effort underway for your distro.

To clarify, I was trying to give assistance with getting dogtail working in CI in response to https://github.com/Nicotine-Plus/nicotine-plus/pull/106#issuecomment-623117663, rather than asking for help getting them working myself (the issue I ran into is a NixOS-specific quirk that shouldn't affect Debian-based CI). DEP-8 is a Debian-specific system, so I doubt any other distribution implements it.

> @emilazy, please don't prepare tarballs for your distro yet. @lene's PR is still very experimental and there are many things that still need to be fixed.

Alright; I'm a little surprised as I did [mention my intention to package this and help test](https://github.com/Nicotine-Plus/nicotine-plus/pull/106#issuecomment-620323974) earlier. To be clear, it's a PR to update the package in the unstable rolling-release version of NixOS, rather than offering packages for any stable NixOS releases. I figured that given the plan to test via the PPA, nixos-unstable is an appropriate venue to help test and shake out bugs, especially since we're trying to move as many packages away from Python 2 now that it's unmaintained. But I've marked it as a draft so that nobody merges it for now.

Do you know of any specific things that need fixing so I can help bring this port to a state where you're happy with it?

kiplingw:
> Do you know of any specific things that need fixing so I can help bring this port to a state where you're happy with it?

Making sure all DEP-8 tests work correctly in autopkgtest. Right now only the first of three do.

lene:
@kiplingw: I created a new MR, https://github.com/Nicotine-Plus/nicotine-plus/pull/124, for fixes. It's still WIP, but I'd value feedback whether everything needed (i.e. mainly the test folder) is found now and works without Syntax Errors.
@emilazy: thanks for your help and you are certainly on the right track with `dbus-run-session`, but it still doesn't work for me (the command simply hangs). I'm still trying though. (Although I have a crude workaround in mind if I cannot get dogtail to work: start nicotine under `Xvfb`, take a screenshot and compare the image with a reference one. Yeah, I know. Better ideas are appreciated ;-))

emilazy:
@lene Not sure, but the hanging might just be because at-spi-bus-launcher runs in the foreground; replacing the `&&` with `&` might help it get further.

lene:
@emilazy Thanks, that was certainly worth a shot, but no dice. If I run it like that, I get the same message as if not using `at-spi-bus-launcher` and `dbus-run-session` at all:
```
Dogtail requires that Assistive Technology support be enabled.
You can enable accessibility with sniff or by running:
'gsettings set org.gnome.desktop.interface toolkit-accessibility true'
```
If I throw in a `sleep 5` to give `at-spi-bus-launcher` ample time to do its stuff, that doesn't help either. I think it's a given that `at-spi-bus-launcher` just hangs (I also tried several variations, where I replaced any of the commands in that sequence with NOPs and that confirmed that.) `at-spi-bus-launcher` seems to just not work within containers/chroots (the result is the same whether in a docker container or `schroot`).

kiplingw:
@lene, when you get a chance please look into the latter two DEP-8 tests that you added which currently are not run. Once that works I will try letting the unstable PPA build the package.

lene:
I'm still working on it, but I have given up on dogtail. Now looking into https://robotframework.org, which seems promising, but has a bit of a learning curve. I'll be back as soon as I have results.

kiplingw:
Thanks for keeping us apprised.

zniavre:
so bug hunting/tracking is open ? :)

jtojnar:
Dogtail checks for toolkit-accessibility setting:

https://gitlab.com/dogtail/dogtail/-/blob/master/dogtail/utils.py#L266

In NixOS we enable it manually in the default memory GSettings backend:

https://github.com/NixOS/nixpkgs/blob/91da9ed1291e9fd94359d7134b07ab48a62c6a7d/pkgs/development/python-modules/dogtail/default.nix#L42-L47

emilazy:
@jtojnar The dogtail test was removed in favour of a simpler one based on Robot Framework, so thankfully those issues are no longer relevant (but for what it's worth, the main issue I had was getting `at-spi-bus-launcher` to work properly in the build sandbox).

Issue #113, opened 2020-04-13
images seem to be integrated from the launch directory if they have special names


kristopolous:
So as you can see below if you have a file named "empty.jpg" in the directory you launch the application from, it just drops it into the client as depicted.

When running strace on it, I can see it finds 2 images, the other one being in a dropdown menu.

These paths should probably be fixed. I can do this if you want, or anyone else, shouldn't be hard.

![fromhome](https://user-images.githubusercontent.com/231761/79169904-9ec36180-7da2-11ea-937c-c16ce85d1b37.png)

kristopolous:
wait, didn't realize there's a 1.4 branch ... ignore this for now.

kristopolous:
problem persists in 1.4.1 ... let me try master

kristopolous:
yep. bug is still real. ok, needs to be fixed.
![persistence](https://user-images.githubusercontent.com/231761/79170320-e0a0d780-7da3-11ea-81d5-fa2a3d38ae07.png)

kiplingw:
Thanks @kristopolous. PR welcome. I should be able to merge that as soon as you're ready. Thanks for finding that.

kristopolous:
@kiplingw ok, thanks, feel free to close this as per whatever you policies are

kristopolous:
hey as an aside, that was pretty fast ... do you just follow your email closely or is there some kind of special notification app you have for your phone? I could use something like that

kiplingw:
Click the "Watch" button up at the top [here](https://github.com/Nicotine-Plus/nicotine-plus/). :)

kristopolous:
ah yeah, I have a zoo of about 200 of them ... 95% of the conversations I don't have time to get into and will resolve without my participation, problem is I can't tell what 95% that is. I guess it's time for a new tool.

PR #114, opened 2020-04-15
Fixes loading of images from the launch directory (#113)

kristopolous:
The ui/icontheme which defaults as the empty string, was being evaluated
as to whether it existed (as opposed to was non-empty). Thus the code
always ran, and the path for a theme was constructed using the default
empty string ("") + the image path for the icon.

This meant if there was path-local resolvable images of the same name
and no theme, they would get dropped into the interface as if it was
a theme.

The fix is to use `dict.__getitem__'s` shortform of `dict.get` which
evaluates the empty string as a bottom value, as intended.

This bug appears to have been introduced in 40a0d921 when the icontheme
was first implemented in 2006.

### DIFF
--- a/pynicotine/gtkgui/frame.py
+++ b/pynicotine/gtkgui/frame.py
@@ -794,7 +794,7 @@ class NicotineFrame:
             "plugin"
         ]
 
-        if "icontheme" in self.np.config.sections["ui"]:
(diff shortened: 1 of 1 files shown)


kiplingw:
Thanks @kristopolous!

PR #117, opened 2020-04-22
Added test data for upload and search

lene:


lene:
Oops, my mistake again :sweat: (I need to start paying attention when creating a PR supposedly to my own branch, GitHub always sets this branch as target by default)

PR #124, opened 2020-05-04
WIP: Python3 fixes

lene:
Fixed packaging process and some bugs.
Integration test still does not work (due to dogtail/at-spi/dbus foo), but packaging should include all necessary files to run all tests.


debian/control
```diff
@@ -6,7 +6,8 @@ Standards-Version: 4.3.0.3
 Build-Depends:
     debhelper-compat (= 11),
     dh-python,
-    python3-all,
+    python3 (>= 3.6),
+    python3-pytest,
     python-gobject-2,
```
kiplingw, line 11: Should this be replaced or removed? I think this is deprecated.

lene: `autopkgtest` failed with `python3-all`, but also without a python interpreter, this version is the one I got to run. `python3-pytest` is needed because apparently the tests are run during the `debuild -S -sa` process.

lene: BTW I think I can cut down on the dependencies listed here, both `Build-Depends` and `Depends`, I'll go through them one at a time. But first, integration tests.

kiplingw: Sorry, I meant `python-gobject-2`. Isn't that deprecated and for Python 2?

kiplingw:
Hey @lene, question still pending in the comments above when you get a chance.

lene:
Hey, I finally got some (very basic) test with the Robot Framework running. It just starts the program and checks it doesn't crash after a few seconds, but it should be something someone can build upon. Someone who is more into that whole topic than me, I mean.
Also removed some obsolete and redundant dependencies.

kiplingw:
Thanks @lene. I'm still trying to get my autopkgtest to work since my test image seems to have broken. But on another note it looks like there are some Gtk issues since your monolithic PR.

```
  File "/usr/bin/nicotine", line 276, in <module>
    run()
  File "/usr/bin/nicotine", line 252, in run
    app = frame.MainApp(data_dir, config, plugins, trayicon, hidden, bindip, port)
  File "/usr/lib/python3/dist-packages/pynicotine/gtkgui/frame.py", line 3776, in __init__
    self.frame = NicotineFrame(data_dir, config, plugins, trayicon, start_hidden, bindip, port)
  File "/usr/lib/python3/dist-packages/pynicotine/gtkgui/frame.py", line 521, in __init__
    self.UpdateColours(1)
  File "/usr/lib/python3/dist-packages/pynicotine/gtkgui/frame.py", line 1456, in UpdateColours
    self.tag_log.set_property("foreground-rgba", colour)
TypeError: argument self: Expected Gdk.RGBA, but got gi.overrides.Gdk.Color
```

lene:
Thanks for the bug report! That's an interesting one, I could only reproduce it by setting the config variable `ui.chatremote`. I suspect there will be many similar issues hiding behind diverse config settings and usage patterns.
How should we proceed with this kind of bugs? Open a PR for every one of them or a collective PR for all that get reported? I have opened https://github.com/Nicotine-Plus/nicotine-plus/pull/125 with a fix for this one, and noted it as a collective PR, but I'm not sure this is the right way to proceed.

kiplingw:
No problem. I'm curious though why they're not being caught at your end. After you build and install the package with `sbuild` or `dpkg-buildpackage`, it runs without any issues for you? I'm building and running it on `eoan`.

```
Package: gobject-introspection
Version: 1.62.0-1

Package: gir1.2-gtk-3.0
Version: 3.24.12-1ubuntu1

Package: python3-gi
Version: 3.34.0-1
```

If you can replicate on your end it's probably better to submit the fix as all one PR since it's really the same issue with what appears to be simple incorrect Gtk API usage.

kiplingw:
1.4.3 pushed to unstable PPA. Package awaiting testing for `Eoan`, `Focal`, and `Groovy`.

zniavre:
good morning trying using last 1.4.3 got huge amount of info with terminal 

https://pastebin.com/fNgSEWqH

nicotine is working quite well search dl/up chatroom private room.

in settings windows i hav only see spellchecker missing (and notify thingy)

statusbar on main window looks oversized ( progressbar is misplaced in a kind of second row) 

what a great step !!! thank you very much !

kiplingw:
Thanks @zniavre . I'm also seeing the Gtk+ warnings. 

But I'm also seeing a lot of socket related messages since @lene's PR being dumped to the user visible log. All of the ones I saw appeared to be innocuous which suggests to me debugging messages may have been unintentionally logged. Perhaps an API change made certain types of exceptions that otherwise would have been quietly ignored are now being caught in an exception handler intended for more significant issues.

```
...
00:16:19 Connection closed by peer: {'addr': None, 'username': '<removed>', 'conn': <socket.socket [closed] fd=-1, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0>, 'msgs': [], 'token': 907, 'init': <pynicotine.slskmessages.PeerInit object at 0x7f4de21afe10>, 'conntimer': <Timer(Thread-1070, stopped 139971776214784)>, 'tryaddr': None}
...
```

lene:
> No problem. I'm curious though why they're not being caught at your end. After you build and install the package with `sbuild` or `dpkg-buildpackage`, it runs without any issues for you? I'm building and running it on `eoan`.

I suppose that's because `ui.chatremote` is unset by default. Makes me wonder why it crashes in your case though. Does the crash happen after a fresh build in the chroot for you, or only if you run it on your day-to-day system?

FWIW, here's how I build the package and run autopkgtest:
```
$ rm ../nicotine_1.4.3*
$ ./debian/rules get-orig-source
$ debuild -S -sa -us -uc
$ autopkgtest ../nicotine_1.4.3-1.dsc -- schroot nicotine-amd64
```
> If you can replicate on your end it's probably better to submit the fix as all one PR since it's really the same issue with what appears to be simple incorrect Gtk API usage.

I'm not sure what you're saying here... the fix for this bug was https://github.com/Nicotine-Plus/nicotine-plus/pull/125 and it's already merged, are you referring to that?

lene:
I'll look into the Gtk warnings and the "Connection closed by peer" messages though. I suppose they are innocuous enough, but there's no need to spam.

kiplingw:
> I suppose that's because `ui.chatremote` is unset by default. Makes me wonder why it crashes in your case though. Does the crash happen after a fresh build in the chroot for you, or only if you run it on your day-to-day system?

It wasn't really a crash, per se. It was more like an apparent hang. Although the message loop may well have still been running, even though repaints weren't happening. If that was the case then it would likely have still passed during CI even if I was having the same problem without being able to see it running.

> FWIW, here's how I build the package and run autopkgtest:
> 
> ```
> $ rm ../nicotine_1.4.3*
> $ ./debian/rules get-orig-source
> $ debuild -S -sa -us -uc
> $ autopkgtest ../nicotine_1.4.3-1.dsc -- schroot nicotine-amd64
> ```

Yup. That looks correct. I'd also recommend, if you aren't already, using a QEMU based container for more realistic hermetically sealed testing.

> I'm not sure what you're saying here... the fix for this bug was #125 and it's already merged, are you referring to that?

Sorry I thought you meant the fix was required in multiple locations. But since your last PR for #125 the UI appears to not hang and renders correctly.

lene:
> > I'm not sure what you're saying here... the fix for this bug was #125 and it's already merged, are you referring to that?
> 
> Sorry I thought you meant the fix was required in multiple locations. But since your last PR for #125 the UI appears to not hang and renders correctly.

Sorry for being unclear - I meant that I expect other bugs of a similar kind which we haven't found yet since they require specific settings or execution paths to turn up.

PR #157, opened 2020-05-22
Ship plugins system-wide

mathiascode:
Would be nice to have the plugins included by default, and potentially detect issues in them.

### DIFF
--- a/plugins/README
+++ b/plugins/README
@@ -1,20 +1,14 @@
 README Concerning plugins for N+
 
 
-FOR EVERYBODY
-
-The plugin system is in its infant stage, a lot of things aren't implemented yet
-and a few things might be buggy. The provided plugins should be regarded as
-documentation and shouldn't be loaded unless you're willing to break things.
-
-
 FOR END USERS
 
-In order to get the plugins working you need to copy them to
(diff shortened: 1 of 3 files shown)


plugins/README
```diff
@@ -1,20 +1,14 @@
...
-
-
 FOR END USERS
 
-In order to get the plugins working you need to copy them to
-~/.nicotine/plugins, creating this directory if it doesn't exist yet. Plugins
-will be loaded upon startup, and can be reloaded with /reload. This reload
-command does not always completely reload the plugins due to the way Python
-works, in some cases a restart of N+ is needed to get new versions working.
+In order to load your own plugins you need to copy them to
```
kiplingw, line 6: This actually should really be relevant more for developers than end users. End users should never have to do this. The plugins should ship within the deb and be correctly installed into the FHS.

kiplingw: When I think of vanilla "end users" I don't normally think of people writing their own plugins.

mathiascode: Good point.

kiplingw:
@mathiascode, I'm not sure if you received the Launchpad builder's [build log failure](https://launchpadlibrarian.net/480829943/buildlog.txt.gz), but it appears this commit broke it.

Issue #154, opened 2020-05-21
1.4.3 Linux - Hidding Tabs - Always Opens Now Under Buddy List


0pLuS0:
Hello,

In Slackware 14.2 with these Deps;

python3-3.7.2
pygobject3-python3-3.18.2
mutagen-1.44.0
gtk+3-3.18.9

I pulled from this Branch;

```git clone -b chatfixes https://github.com/mathiascode/nicotine-plus.git```

When I hide the 'Chat Rooms' and 'Private Chat', every time I start Nicotine+ it now opens up under the 'Buddy List' tab.

I'm hoping, even with my older DEPS, it might still open under the last tab I left it on.

```I've attached a screen shot showing it opened up under the 'Buddy List'```;

![Nicotine_tabs](https://user-images.githubusercontent.com/36190223/82604545-03dd5480-9ba4-11ea-9350-7fd4ae7ba6d3.jpg)

THANKS

mathiascode:
I'll look into this.

0pLuS0:
This now works, I also mentioned it in #169 

THANKS

mathiascode:
Nice to hear! You're welcome. :)

PR #169, opened 2020-05-23
Add option to restore previous tab on startup

mathiascode:
If disabled, the first tab is now selected instead of the last one.

Fixes https://github.com/Nicotine-Plus/nicotine-plus/issues/154


0pLuS0:
This is working for me now!

THANKS :)

PR #191, opened 2020-05-25
Add Soulseek protocol doc from old wiki

mathiascode:
Converted from https://www.nicotine-plus.org/wiki/SoulseekProtocol


kiplingw:
Mat, I'm not sure if you're in the mood or have time for it, but it would be great if we also had a finite state machine, UML sequence, or some other diagram demonstrating handshake, transactions, etc.

mathiascode:
I'm too unfamiliar with the protocol at this point to do it, and I also don't feel like investing much time in that right now.

Added a slightly newer revision from https://www.museek-plus.org/wiki/SoulseekProtocol. The last edit was 10 years ago, so some things are probably out-of-date.

PR #193, opened 2020-05-25
Remove files that helped during development but are now useless

lene:


kiplingw:
Good call @lene. Can you rebase?

lene:
Done!

PR #196, opened 2020-05-25
gir1.2-notify-0.7 dependency

lene:
This dependency was introduced somewhere but not added to debian/control. Fixing this.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -26,6 +26,7 @@ Depends:
     miniupnpc (>= 1.9),
     gobject-introspection,
     gir1.2-gtk-3.0,
+    gir1.2-notify-0.7,
     python3 (>= 3.5),
     python3-dbus,
     python3-geoip (>= 1.3.2),
--- a/test/integration/nicotine.robot
+++ b/test/integration/nicotine.robot
@@ -11,5 +11,5 @@ Library           nicotine_library.py
 
 Run nicotine as shell command
     Running nicotine starts a process    nicotine    3
-    Result should be    0
+    Result should be    ${True}
 


test/integration/nicotine.robot
```diff
@@ -11,5 +11,5 @@ Library           nicotine_library.py
 
 Run nicotine as shell command
     Running nicotine starts a process    nicotine    3
-    Result should be    0
+    Result should be    ${True}
```
kiplingw, line 14: Thanks @lene. Ignore my incompetence. I'm not very familiar with the Robot Framework.

kiplingw:
Thanks @lene. When I was first introduced to Debian's `autopkgtest` it all seemed like overkill for testing. Then it took enough situations like this to realize where it shines because your development environment always contains all kinds of side effects and alterations that's impossible to control for when trying to replicate an end user's environment if you don't have a hermetically sealed container.

debian/control
```diff
@@ -26,6 +26,7 @@ Depends:
     miniupnpc (>= 1.9),
     gobject-introspection,
     gir1.2-gtk-3.0,
+    gir1.2-notify-0.7,
```
kiplingw, line 29: Should we add this to README.md as well? I'm not sure if this is a Debian and derivative specific nomenclature or if this package is more generally known by this name in the wild.

lene: Probably a good idea! I know nothing about this package, but it looks awfully specific.

lene:
> Thanks @lene. When I was first introduced to Debian's `autopkgtest` it all seemed like overkill for testing. Then it took enough situations like this to realize where it shines because your development environment always contains all kinds of side effects and alterations that's impossible to control for when trying to replicate an end user's environment if you don't have a hermetically sealed container.

Yes, I must admit I'm a convert! I'll still moan about what a pain it is to set up though.

kiplingw:
It is definitely a pain to setup the first time. But after that it's pretty easy to set them up for different releases and different architectures like `arm64`, `ppc64el`, etc.

debian/control
mathiascode: At least on Manjaro/Arch, everything seems to be bundled in one "libnotify" package.

kiplingw: I'll add it anyways for safe measure.

Issue #192, opened 2020-05-25
userlist for myself does not display files number


zniavre:
in userlist and buddylist for myself it displays 0 file , i delete my old config file to make new configuration but it still does not display real number of my share files (it should display at least 6000 +/-)

thank you.

zniavre:
i hav confirmation that a friend using slskQT can see my number of files in the user list and buddylist 

but here with last commit of nic+ it does not appear

mathiascode:
Looks like the file count is incorrectly displayed under the directory count for some reason.

PR #200, opened 2020-05-26
Don't use incorrect type when getting objects

mathiascode:
These were supposed to be parsed as long, not int.
Fixes https://github.com/Nicotine-Plus/nicotine-plus/issues/192


zniavre:
it works ! 

Thank you

Issue #77, opened 2018-04-03
 select ValueError: filedescriptor out of range in select()


0pLuS0:
When doing a search in Linux in 1.4.1 at the terminal, I noticed the terminal filled up with these messages;

18:50:56 select ValueError: filedescriptor out of range in select()
18:50:56 select ValueError: filedescriptor out of range in select()
18:50:56 select ValueError: filedescriptor out of range in select()
18:50:56 select ValueError: filedescriptor out of range in select()
18:50:56 select ValueError: filedescriptor out of range in select()
18:50:56 select ValueError: filedescriptor out of range in select()
18:50:56 select ValueError: filedescriptor out of range in select()
18:50:56 select ValueError: filedescriptor out of range in select()
18:50:56 select ValueError: filedescriptor out of range in select()
18:50:56 select ValueError: filedescriptor out of range in select()
18:50:56 select ValueError: filedescriptor out of range in select()
18:50:56 select ValueError: filedescriptor out of range in select()
18:50:56 select ValueError: filedescriptor out of range in select()
18:50:56 select ValueError: filedescriptor out of range in select()
18:50:56 select ValueError: filedescriptor out of range in select()

By the way, it never stops, it just floods the terminal like crazy...

Hmm

PR #217, opened 2020-05-31
Use Python selectors module instead of select()

mathiascode:
This way we automatically use the best syscall for the platform (epoll, poll etc.), thus bypassing the file descriptor limit of select().

Fixes https://github.com/Nicotine-Plus/nicotine-plus/issues/77 (this happened when there were many search results, and thus too many connections)

### DIFF
--- a/pynicotine/gtkgui/downloads.py
+++ b/pynicotine/gtkgui/downloads.py
@@ -90,7 +90,6 @@ class Downloads(TransferList):
         cols = frame.DownloadList.get_columns()
 
         try:
-            print(len(cols))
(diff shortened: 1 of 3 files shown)


kiplingw:
Looks like some tests failed. Perhaps it's still early and needs some more testing?

mathiascode:
No idea what Github Actions has against epoll...

kiplingw:
Me neither. That is a bit weird. `epoll(7)` usage seems to be so common place, I don't know why the test beds would trip up over them. Something probably worth looking into if I had more time.

PR #229, opened 2020-06-03
Add checks for settings window when updating config

mathiascode:
Fixes https://github.com/Nicotine-Plus/nicotine-plus/issues/228

PR #122, opened 2020-04-30
DEP-8 integration testing

lene:
DEP-8 CI separated out. Depends on https://github.com/Nicotine-Plus/nicotine-plus/pull/106 to actually do anything.

PR #234, opened 2020-06-03
CI check (don't merge)

mathiascode:


test/integration/nicotine.robot
```diff
@@ -11,5 +11,5 @@ Library           nicotine_library.py
 
 Run nicotine as shell command
     Running nicotine starts a process    nicotine    3
-    Result should be    ${True}
+    Result should be    ${0}
```
kiplingw, line 14: Hey @mathiascode. This was originally my proposed fix too the last time we had a similar issue in #122. See what @lene said [here](https://github.com/Nicotine-Plus/nicotine-plus/pull/122#issuecomment-633637570).

mathiascode: I changed the approach a bit, since I didn't get Popen to exit on failure. We start Nicotine+ using subprocess.call with a timeout. If the program is still running when the timeout is reached, the test was successful, and the result is set to True.

kiplingw:
Got it. Do you need me to merge?

mathiascode:
Nah, changes are already in https://github.com/Nicotine-Plus/nicotine-plus/pull/229.

PR #238, opened 2020-06-04
Run autopkgtest on all available Ubuntu images

mathiascode:


mathiascode:
Welp, that will need some tinkering. I'll look into it later.

.github/workflows/tests.yml
```diff
@@ -10,6 +10,9 @@ jobs:
       matrix:
         os: [ubuntu-16.04, ubuntu-18.04, ubuntu-20.04]
         python: [3.5, 3.6, 3.7, 3.8]
```
kiplingw, line 12: I actually think 3.6 isn't available on `Focal` either.

mathiascode: Doesn't seem like any Python version will install on the 20.04 image yet. It's a preview after all, so I'll disable it for now.

mathiascode: autopkgtest was an unknown command on 16.04 for some reason, passed on 18.04 and 20.04.

kiplingw:
Looks good. Let me know when ready.

mathiascode:
Need to look into why autopkgtest isn't a valid command on Xenial first.

kiplingw:
The binary was called something different back then. See adt-* binaries. You want to call the appropriate one that runs it within a container if possible.

.github/workflows/autopkgtest.yml
```diff
@@ -80,7 +83,8 @@ jobs:
         run: |
           set +e
           NICOTINE_VERSION=$(<NICOTINE_VERSION/NICOTINE_VERSION)
-          sudo autopkgtest ./${DEBUILD_SOURCE}/nicotine_${NICOTINE_VERSION}_source.changes -- null
+          sudo autopkgtest ./${DEBUILD_SOURCE}/nicotine_${NICOTINE_VERSION}_source.changes -- null || \
```
kiplingw, line 86: This works, but it would be more readable for people to use an `if` statement.

.github/workflows/tests.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-16.04, ubuntu-18.04, ubuntu-latest]
+        os: [ubuntu-16.04, ubuntu-18.04]
```
kiplingw, line 11: Is there a reason why ubuntu-latest was removed?

mathiascode: ubuntu-latest is just an alias for ubuntu-18.04 at the moment.

PR #250, opened 2020-06-06
Add appdata file

mathiascode:
Can be tweaked later if desired.


kiplingw:
For anyone else who would like to contribute, the official XDG specification is available [here](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html).

PR #251, opened 2020-06-07
Ship AppStream metainfo file with packages

mathiascode:


mathiascode:
As with everything else, there seems to be conflicting information about whether to use *.appdata.xml or *.metainfo.xml. Sticking with appdata, since most GUI applications are using it, and Fedora wants it for GUI applications.

kiplingw:
I think either will work, but I'd suggest using `*.metainfo.xml`. It appears as though the former may be [deprecated](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location).

PR #282, opened 2020-06-13
Fix setting process title without procname

toofar:
I'm not sure how the c interface works but it was setting the process title to "n" for me. When passing it as a list of bytes like this it is correctly set to "nicotine".

Also not sure why I've ran into this on two machines (finally updated the one on my seedbox) and no-one else has reported it.

Locales shouldn't be an issue as it doesn't look like the string passed is being localized.

### DIFF
--- a/nicotine
+++ b/nicotine
@@ -120,7 +120,7 @@ def renameprocess(newname, debug=False):
         import ctypes
         # GNU/Linux style
         libc = ctypes.CDLL('libc.so.6')
-        libc.prctl(15, newname, 0, 0, 0)
+        libc.prctl(15, ctypes.c_char_p(newname.encode('ascii')), 0, 0, 0)
     except Exception:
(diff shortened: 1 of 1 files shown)


kiplingw:
Thanks @toofar. This is what I'm seeing without your PR:

```
$ ps aux | grep nicotine
kip       212740  0.0  0.0   6380  2364 pts/0    S+   18:12   0:00 grep --color=auto nicotine
kip      3201042  1.2  0.3 4436208 117504 ?      Sl   Jun11  14:56 /usr/bin/python3 /usr/bin/nicotine
```

Where would a user look to see the incorrect process name?

toofar:
Just in top, ps reports the command line correctly for me the same as in your example.

kiplingw:
I can't replicate with htop(1), but I could with top(1).

PR #289, opened 2020-06-15
Remove Windows-specific upnp code

mathiascode:
The only environment we support on Windows nowadays is MinGW, which doesn't need these workarounds.

### DIFF
--- a/pynicotine/upnp.py
+++ b/pynicotine/upnp.py
@@ -17,10 +17,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import platform
 import re
 import socket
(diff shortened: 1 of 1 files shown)


mathiascode:
@kiplingw On an unrelated note, do you think you could give me access to manage issues in this repo? I feel bad pinging you for every little thing. :P

kiplingw:
I was actually already trying to figure out how to do that a while ago but couldn't see an option in the project settings.

mathiascode:
I think one way to do it would be to create a new team, and add the triage permission for the repo in that team.

kiplingw:
Done. Ok hopefully that works now.

mathiascode:
That did it. Thanks!

PR #296, opened 2020-06-15
Detect install prefix when running

mathiascode:
Uses /app for Flatpak instead of /usr.


mathiascode:
This PR is ready for merge.

Issue #335, opened 2020-06-29
Question: group by folders vs group by users 


zniavre:
the new feature is nice but instead of replacement can we hav the choice to use this one ? 
im much acclimated to group by users display
maybe this new feature is not mature enough to be the only way displaying list ? 

sorry it's not a real complain but a kind a feature request

mathiascode:
I'll check if it's feasible to add the old feature back.

zniavre:
some information to not be considered negative guy : 
1/ it missing on empty line general information > 
![Capture d’écran du 2020-06-29 11-47-38](https://user-images.githubusercontent.com/2409578/86262787-a126a200-bbc0-11ea-8455-507e7030641f.png)

2/ with upload tabs im experiencing strange behavior please let me  show you with 4 screenshots 
![Capture d’écran du 2020-07-01 17-28-21](https://user-images.githubusercontent.com/2409578/86262981-da5f1200-bbc0-11ea-9066-fc9225637e4f.png)
![Capture d’écran du 2020-07-01 17-28-30](https://user-images.githubusercontent.com/2409578/86262987-dc28d580-bbc0-11ea-8a5e-6da2797ebe9d.png)
![Capture d’écran du 2020-07-01 17-28-41](https://user-images.githubusercontent.com/2409578/86262993-dd5a0280-bbc0-11ea-92f9-2aaf884b6cd3.png)
![Capture d’écran du 2020-07-01 17-28-55](https://user-images.githubusercontent.com/2409578/86263000-dfbc5c80-bbc0-11ea-8551-72ad7688ea52.png)

2 users but only one grouping ( you can see how better are screenshot than my explaination)  

hav a nice day; be good and well 

><((((*>

mathiascode:
Regarding the second issue (wrong user grouping), it was fixed in https://github.com/Nicotine-Plus/nicotine-plus/pull/339.

As for the first issue, I left the user rows empty due to performance reasons, but there is probably some reasonable way the row info can be brought back.

mathiascode:
The user row info issue is resolved in PR https://github.com/Nicotine-Plus/nicotine-plus/pull/337

zniavre:
Thank you !! well done

mathiascode:
You're welcome! :)

Issue #34, opened 2016-12-14
Downloads tab hanging when adding a lot of files


gfarmerfr:
Extracted from https://github.com/Nicotine-Plus/nicotine-plus/issues/32

When you have a lot (about 3000 or more) files added to download, the program hangs for several seconds when doing anything, if the Downloads tab is visible. I suspect that it is trying to sort every item each tick of the GUI main loop.

mathiascode:
In addition to already added downloads, adding a lot of new downloads also lags the program hard, since the transfers file is opened and closed for every new transfer.

mathiascode:
I almost have a solution ready for this issue.

mathiascode:
There was a quite severe bug in the current code. When asking the server for a user's IP address, we did this for every download in the transfer list instead of once, resulting in some nasty connection spam.

PR #337, opened 2020-06-29
Improve transfer list performance

mathiascode:
Not quite ready yet, but we're getting there. This PR massively improves the performance of the transfer list. There are a lot of optimizations, some of which are:
- Avoid unnecessary looping through the download and upload lists, which degrades the performance as the lists grow. This was the main reason for performance loss.
- Reduce translation requests of status strings when updating the transfer list
- Don't spam the Soulseek server with user IP requests for every transfer in the list
- Save the download list to file every few seconds instead of saving it after updating each transfer

Many thanks to the py-spy profiler!

Fixes https://github.com/Nicotine-Plus/nicotine-plus/issues/34
Fixes https://github.com/Nicotine-Plus/nicotine-plus/issues/335


mathiascode:
Essentially done, but I need to do more testing to make sure I didn't break anything.

mathiascode:
More work can be done in the future, but at least the program doesn't freeze indefinitely anymore. Ready to merge.

mathiascode:
Finally done (I hope).

PR #353, opened 2020-07-04
Don't reset buddy list when moving it

mathiascode:
User statuses would reset if you changed the position of the buddy list. The code comment mentions "Reinitialize the userlist to avoid an error that freeze the UI on recent GTK versions", but this doesn't seem to be the case anymore.

Also some minor cleanup, and removal of buttons that were oddly placed.


kiplingw:
Looks like it may still need some work, Mat.

mathiascode:
"E302 expected 2 blank lines, found 1" oh well

PR #359, opened 2020-07-04
Add option to change leech detector message

mathiascode:
I'd imagine this is a feature people want.

Also some code tweaks to immediately enable the plugin properties button when enabling a plugin.


zniavre:
hello i do not know if closed i can continue  speaking about this one.
is that possible to expand or resize the entry area in the plugin properties window ? it's quite short to write a long message, even the panda is sad to be unreadable 

thank you :o)

Issue #343, opened 2020-07-02
Can't find anything from Wu-tang


hboetes:
I just searched for `Legend Of The Wu-Tang` which gave no results. Then I searched for `Legend Of The Wu Tang` — without the dash — and I did get results. What's going on here?

mathiascode:
I don't think there's much that can be done here. The same thing happens in the official SoulseekQt client.

hboetes:
Would it be an idea to replace - with spaces automatically? Maybe even replace some other characters like / and .

mathiascode:
I'm not sure if it's a good idea, considering how much the results seem to change depending on which characters you include.

hboetes:
Right, it's used to modify search results. https://www.slsknet.org/news/node/2855 as in search for everything `wu` but omit everything with `tang`. Rather sloppy coding if you'd ask me (on the slsk server side) since there is a clear difference between `search -omit` and `jay-z`, and any decent coder would fix that. _Sigh_ closed source. :-(

Perhaps warn users if you'd search for `[A-z]-[A-z]`

mathiascode:
@hboetes Could you test this PR? https://github.com/Nicotine-Plus/nicotine-plus/pull/368

git clone -b searchfilter https://github.com/mathiascode/nicotine-plus

zniavre:
since the first day of this bug i can't reproduce it:  with or without dash
it found many files...

mathiascode:
@zniavre Non-SoulseekQt clients send results if you include special characters, but SoulseekQt does not. If you search for something with a dash or hyphen, you should notice that there are no results from SoulseekQt clients (their folders start with @@).

hboetes:
I just searched for wu-tang with your branch and repo and I got results (a few albums) then I searched with the master branch and repo and I got different results( a few loose tracks).

I like your version better. :-)

mathiascode:
You can toggle the character filtering in settings -> searches, if desired. Searches without special characters in them should essentially be the same as searches with them, just with the addition of results from SoulseekQt clients.

mathiascode:
Excluding words by placing - in front of them should work as long as they are clearly separated words.

hboetes:
Yes, I like a lot. Could you sort search results by speed by default? :-)

mathiascode:
I will defer the sorting change or now, since I had issues where the list would jump around a lot if you started scrolling when elements were still added.

hboetes:
[...]

PR #368, opened 2020-07-07
Search result improvements

mathiascode:
- Add option to remove special characters from search terms, enabled by default. Closes https://github.com/Nicotine-Plus/nicotine-plus/issues/343
- Add functional search result word filtering (put "-" in front of a word), like the official client has
- Filter out inexact or incorrect search results client-side, like the official client does

Don't merge yet.


mathiascode:
Ready

PR #379, opened 2020-07-09
Add miniupnp to required dependencies

mathiascode:
- List miniupnp as a required dependency, and enable UPnP by default (like SoulseekQt) to cut down on potential connection issues among new users
- Unlink UPnP and direct connection checkboxes, since UPnP doesn't guarantee that direct connections are possible (e.g. ports blocked on ISP level). UPnP will still help if there are many clients on the same network.


kiplingw:
This is good. This makes source level consistent with what we already require in `debian/control`.

PR #392, opened 2020-07-14
Update README.md

hboetes:
Add instructions for building from the git repository.


mathiascode:
Thanks! Could you update https://github.com/Nicotine-Plus/nicotine-plus/blob/master/doc/RUNFROMGIT.md with more details instead, and add a link to it in the readme?

hboetes:
Good idea. I'm on it.

PR #394, opened 2020-07-14
Fix issue in update checker

mathiascode:
The update checker would display the whole json response instead of the version number.

### DIFF
--- a/.github/workflows/autopkgtest.yml
+++ b/.github/workflows/autopkgtest.yml
@@ -76,6 +76,8 @@ jobs:
           name: debuild-source
       - name: Install autopkgtest dependencies
         run: |
+          sudo apt update
+          sudo apt upgrade -y
           sudo apt install autopkgtest
       - name: Run autopkgtest
         run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/autopkgtest.yml
```diff
@@ -76,6 +76,7 @@ jobs:
           name: debuild-source
       - name: Install autopkgtest dependencies
         run: |
+          sudo apt update
```
kiplingw, line 79: Also add a sudo apt upgrade -y after this.

.github/workflows/tests.yml
```diff
@@ -18,7 +18,8 @@ jobs:
           python-version: ${{ matrix.python }}
       - name: Install dependencies
         run: |
-          sudo apt-get -y install miniupnpc gobject-introspection gir1.2-appindicator3-0.1 gir1.2-gsound-1.0 gir1.2-gtk-3.0 \
+          sudo apt update
```
kiplingw, line 21: Probably same thing here too @mathiascode. `sudo apt upgrade -y`. Although this CI stuff really should be in a separate PR.

PR #404, opened 2020-07-17
Change layout to improve rendering on main website's NEWS.html

hboetes:


hboetes:
The introduction is rendered as a code block, it should be a normal paragraph.

https://nicotine-plus.github.io/nicotine-plus/NEWS.html

kiplingw:
Rather than removing that block, can you simply change how it is rendered?

hboetes:
Please check the diff again, I didn't remove any text, I removed the leading spaces so it will be indented like a paragraph.

kiplingw:
Ah yes, you are right.

PR #415, opened 2020-07-21
Remove Unicode-related Python 2 leftovers

mathiascode:
Yay to less platform-specific code. :)

Pretty sure none of this is needed in Python 3 anymore. I didn't detect any issues when testing in Windows.

The config is now saved as utf-8 on Windows, to fix problems when using non-latin characters.


mathiascode:
Closing for now, will reopen later.

mathiascode:
Should be ready.

PR #420, opened 2020-07-22
Clean up and move a few files

mathiascode:
- Remove Docker files, which are becoming out of date. They can be resurrected from the git history if anyone needs them.
- Reorganize a few files
- Remove old memory tracker script. Can be resurrected if anyone wants it, but there are probably better options out there.


kiplingw:
Thank you for your attention to detail.

PR #425, opened 2020-07-24
Convert about-dialogs to ui files

mathiascode:


pynicotine/gtkgui/ui/about/about.ui
```diff
@@ -0,0 +1,187 @@
...
+    <property name="comments" translatable="yes">A graphical client for the Soulseek peer-to-peer system
+Based on code from Nicotine and PySoulSeek
+
+This product includes IP2Location LITE data, available from:
+https://www.ip2location.com</property>
+    <property name="website">https://nicotine-plus.org</property>
+    <property name="license_type">GTK_LICENSE_GPL_3_0_ONLY</property>
+    <property name="authors"># Nicotine+ MAINTAINERS
+
+### Active
```
kiplingw, line 20: @mathiascode, can you have this loaded at runtime from disk so we don't have redundant copies of this text?

mathiascode: I had to revert this change. The dialog elements are destroyed when you close the dialogs.

PR #426, opened 2020-07-24
Improve chat room and user browse UI

mathiascode:
An attempt to reduce the number of vertical elements in the user browse UI to leave more room for files, and just make everything tidier in general.
s1
s2


zniavre:
close button of about window is not active ...

mathiascode:
Can you elaborate?

mathiascode:
I've fixed the issue where the escape key doesn't close the about dialogs, if that's what you're referring to. I'll create a PR shortly.

PR #441, opened 2020-07-29
Simplify buddy list code

mathiascode:
- Reduce the number of lists used to store data
- Improve code consistency and readability
- Disable a few popup items for offline users

### DIFF
--- a/pynicotine/gtkgui/chatrooms.py
+++ b/pynicotine/gtkgui/chatrooms.py
@@ -728,7 +728,7 @@ class RoomsControl:
                 clist += self.rooms
 
             if config["buddies"]:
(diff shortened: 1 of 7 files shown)


kiplingw:
Thanks @mathiascode, but like the other recent PRs, this seems to break in-tree and CI tests.

mathiascode:
Something is wrong with CI atm, dependencies fail to install. I'll try again in a while.

kiplingw:
Yeah, strange. Looks like it may be an issue with the test machine:

```
mysql_upgrade: Got error: 2002: Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2) while connecting to the MySQL server
```

PR #442, opened 2020-07-30
Improve user share performance

mathiascode:
- Large improvements to performance when scanning user share
- Reduced memory usage
- Improve user browse response performance by using a byte array. On a share with 1.5 million files, this change cuts down the startup time of Nicotine+ from indefinite (hanging) to ~4 seconds.


mathiascode:
Should be ready.

PR #446, opened 2020-07-31
Automate Windows installer generation

mathiascode:


.github/workflows/packaging.yml
```diff
@@ -0,0 +1,70 @@
+name: Nicotine+ packaging CI
```
kiplingw, line 1: @mathiascode, it's probably not a good idea to describe this as merely packaging since what we're really trying to do is test the installed artefacts from the built packages. So I encourage you to revert back to the previous file name.

PR #448, opened 2020-07-31
Minor documentation-related changes

mathiascode:
- Use less hard-coded file names
- Removed debian/docs, since setup.py should already handle documentation files


setup.py
```diff
@@ -102,8 +69,29 @@
     )
 )
 
-# Translation files
-mo_dirs = [x for x in glob.glob(os.path.join("languages", "*")) if isdir(x)]
+# Documentation
+docfiles = glob.glob("[!404.md]*.md") + glob.glob(os.path.join("doc", "*.md"))
```
kiplingw, line 73: Thanks @mathiascode. This is better if the build environment finds the documentation. The `debian/docs` is really only for upstream packages that ship documentation in non-standard locations. I haven't tested the built package, but you're confident the documentation still ships correctly?

mathiascode:
I will check this soon.

mathiascode:
I checked, there is no difference in the doc output with these changes.

PR #449, opened 2020-07-31
Windows installer improvements

mathiascode:
Cuts the installer size down from ~40 MB to ~25 MB.


files/windows/dependencies.sh
```diff
@@ -2,6 +2,7 @@
 
 pacman --noconfirm -S --needed \
   git \
+  upx \
```
kiplingw, line 5: ;)

mathiascode: This is why I'm happy the Windows packaging process is finally automatic. :)

PR #450, opened 2020-07-31
Untested MacOS support

mathiascode:
I can't confirm that these packages actually work, since I have no means of testing them at the moment, but they are available if any MacOS users want to help us out.


mathiascode:
@droserasprout in case you're interested

Issue #454, opened 2020-08-02
progress bar stuck at 100%


hboetes:
During a rescan of my collection I noticed the progress bar got stuck at 100%, where it would usually disappear, and I found this trace message in the console:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/shelve.py", line 111, in __getitem__
    value = self.cache[key]
KeyError: 'Klassiek\\Jean Sibelius\\Two Serenades for Violin and Orchestra, Op.69'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/pynicotine/gtkgui/frame.py", line 2039, in _RescanFinished
    self.np.shares.sendNumSharedFoldersFiles()
  File "/usr/local/lib/python3.7/site-packages/pynicotine/shares.py", line 105, in sendNumSharedFoldersFiles
    sharedfiles = sum([len(x) for x in list(conf["transfers"][shared_db].values())])
  File "/usr/local/lib/python3.7/_collections_abc.py", line 762, in __iter__
    yield self._mapping[key]
  File "/usr/local/lib/python3.7/shelve.py", line 113, in __getitem__
    f = BytesIO(self.dict[key.encode(self.keyencoding)])
  File "/usr/local/lib/python3.7/shelve.py", line 70, in closed
    raise ValueError('invalid operation on closed shelf')
ValueError: invalid operation on closed shelf

```
This problem is repeatable. I can't say I see anything odd in that directory.

mathiascode:
Thanks, I can reproduce the issue. Doesn't happen consistently.

PR #455, opened 2020-08-02
Further file scanning optimizations

mathiascode:
- Send number of files and folders a bit later, to ensure that the shares db is open. Should fix https://github.com/Nicotine-Plus/nicotine-plus/issues/454
- Remove pointless creation of list
- No need to sync shelves, as we're not in writeback mode
- Compress file share response in a new thread, to get consistent startup times no matter how large your file share is
- Remove unneeded code
- A whole lot of changes according to profiling results

### DIFF
--- a/pynicotine/config.py
+++ b/pynicotine/config.py
@@ -967,9 +967,7 @@ class Config:
             (fileindex, "bfileindex", "buddyfileindex.db")
         ]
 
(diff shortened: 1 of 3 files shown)


mathiascode:
Marking this as ready. I have another upcoming PR with more optimizations.

PR #463, opened 2020-08-04
Use TagLib instead of Mutagen

mathiascode:
Mutagen is fairly annoying to maintain, and massively bogs down the file share scanning process due to being a pure Python implementation. TagLib is a C++ library with equivalent features, and has a good Python wrapper we can use.

When rebuilding a file share, this change reduces the scanning time by 70-80%, bringing us to SoulseekQt speeds.


hboetes:
I just tested everything on OpenBSD. This is a great improvement. Thanks @mathiascode !

hboetes:
Can we do a 2.1.0 now? :-)

mathiascode:
Not quite yet, still have a few things to do.

PR #472, opened 2020-08-05
Tray icon improvements

mathiascode:
- Add support for Ayatana indicators
- Don't hardcode tray icon, use system icon theme when possible


debian/control
```diff
@@ -30,7 +30,7 @@ Depends:
     python3-taglib,
     xdg-utils
 Recommends:
-    gir1.2-appindicator3-0.1,
+    gir1.2-ayatanaappindicator3-0.1 | gir1.2-appindicator3-0.1,
```
kiplingw, line 33: Very good, @mathiascode. You're almost ready to become an official Debian developer.

mathiascode: Not ready to become a Debian user though. :]

PR #491, opened 2020-08-07
Use dashes instead of underscores in tray icon names

mathiascode:
Let's not be the odd ones out here.
https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html

The icon names for custom tray icons (trayicon_x) will stay the same for compatibility reasons.


pynicotine/gtkgui/tray.py
```diff
@@ -177,7 +177,7 @@ def Load(self):
             if self.appindicator is not None:
                 trayicon = self.appindicator.Indicator.new(
                     "Nicotine+",
-                    "org.nicotine_plus.Nicotine_connect",
+                    "",
```
mathiascode, line 180: Removed because it's set at startup in frame.py anyway.

PR #500, opened 2020-08-09
Settings dialog modernization

mathiascode:
- Replace most deprecated elements in the settings dialog
- Reorganize a few settings items to make them more easily discoverable
- Improve the UI


mathiascode:
PyInstaller 4.0 was released a few hours ago, which seems to break the Windows package generation for some reason. Looking into it at some later point.

hboetes:
I've just done some loose checks on all the menus, they all work as expected. It looks clean and I can't find options that don't work. I didn't check each menu in detail.

mathiascode:
@hboetes Thanks, that issue is fixed now.

PR #522, opened 2020-08-11
Make distributed searches more reliable

mathiascode:
Previously, it could take a long time to find a working parent connection to send us search requests, since in the worst case scenario we connected to a potential parent, waited for a response or timeout, and then proceeded to the next parent in the list. I noticed that SoulseekQt connects to all 10 potential parents at once, and adapted the same behavior in Nicotine+. As a result, distributed searches should be up and running within a minute of starting Nicotine+.

### DIFF
--- a/pynicotine/pynicotine.py
+++ b/pynicotine/pynicotine.py
@@ -1724,10 +1724,11 @@ class NetworkEventProcessor:
                     """ We previously attempted to connect to all potential parents. Since we now
                     have a parent, stop connecting to the others. """
 
-                    if i.conn is not None and i.conn != msg.conn.conn:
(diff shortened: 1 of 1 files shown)


mathiascode:
@zniavre This should hopefully fix the issues you've been having where your files don't appear in search results.

hboetes:
> @zniavre This should hopefully fix the issues you've been having where your files don't appear in search results.

Didn't we disable finding your own files?

mathiascode:
Others haven't been able to find zniavre's files recently.

mathiascode:
Fixed an issue that prevented us from finding a new parent once our current one disconnected.

hboetes:
I have tested this PR and I haven't found any issues.

Issue #535, opened 2020-08-13
Warning: unknown object type <class 'bool'> in message <class 'pynicotine.slskmessages.FileSearchResult'>


zniavre:
Nicotine+ version:  2.1.0.dev1.
Operating System/Distribution:20.04

### Describe the bug
messages in log area
https://pastebin.com/UG3vEw0s

### Expected behavior
less messages

PR #537, opened 2020-08-13
Make necessary conversion from bool to int

mathiascode:
Fixes https://github.com/Nicotine-Plus/nicotine-plus/issues/535

@zniavre You may have to remove your shares db and rescan.


hboetes:
I just tried and my shares still worked after applying the PR.

PR #548, opened 2020-08-14
Fix issue where we don't get a new parent after previous disconnected

mathiascode:
By the time our old parent disconnects, the connection stored in parentconn has most likely become out of sync with the one in the peer conn list, which leads to Nicotine+ never asking the server for a new parent. Revert to the old way of getting our parent connection.


hboetes:
For anyone interested after a random amount of time you'd stop receiving search requests from others and nobody would download anything from you. Like the Dutch would say: "That's not how we are married."

PR #555, opened 2020-08-15
Make "send to player" feature on uploads page functional

mathiascode:
Seems like this was broken in 1.4.x too.


zniavre:
if send to player is usefull in download list , is it to upload list since it's already in our share folder do we really need to listen what we are uploading  ?

mathiascode:
Maybe if someone wants to be reminded of what exactly they're sharing? Not sure what the reason was for adding it.

PR #556, opened 2020-08-15
Replace update checker with website redirect

mathiascode:
The update checker is quite fragile and "dumb" at the moment (doesn't tell the user how to update Nic+). Make the menu item open the website instead, which can easily be updated for many years to come, and doesn't rely on GitHub's API.


kiplingw:
Hey @mathiascode. I'm not sure that this is a good idea. I think the solution would be to refactor the code so that it is more robust. What we have is a useful feature, just not robust enough in its implementation.

PR #563, opened 2020-08-17
autopkgtest corrections

mathiascode:
- Run autopkgtest on current code instead of master
- Pybuild sets the http proxy to 127.0.0.1:9, which conflicts with our tests that need internet access. Disable tests as part of the build process, since they're defined to run later in debian/tests/control anyway.


debian/rules
```diff
@@ -35,6 +35,9 @@ SOURCE_DIR  = $(abspath $(PACKAGE_DIR)/../)
 #  embedded directly in the Python source...
 VERSION_UPSTREAM_DISTUTILS = $(shell sed -n 's/^version\s*=\s*"\(\S*\)"$$/\1/p' $(SOURCE_DIR)/pynicotine/utils.py)
 
+# Disable tests in pybuild, we run them later...
+export PYBUILD_DISABLE=test
```
kiplingw, line 39: Maybe it would be safer to disable only those tests that depend on network connectivity?

mathiascode: We can probably do that.

mathiascode:
Not sure if someone's abusing one of GitHub's macOS test servers or something else is going on, but it seems to hit GitHub's rate limit at times when trying to fetch the latest version.

mathiascode:
https://github.com/actions/virtual-environments/issues/602

I'll go ahead and disable the version test on the macOS workflow, if that's ok? @kiplingw

kiplingw:
Sure, but please keep an eye on what's happening with the OS X environment in case someone discovers a workaround.

mathiascode:
Will do.

PR #604, opened 2020-08-28
Avoid potential performance bottleneck, log improvements

mathiascode:
- If the max socket limit is too low (e.g. default select() syscall on Windows, not multiselect), it is possible for the overflow connection killer to be called over and over again, impacting performance negatively and increasing memory usage. Since this piece of code is not accessed if you're using a modern system with alternatives like epoll and kqueue, I've opted to remove it.
- Somewhat smarter behavior for the selectors module code to avoid temporary dicts and lists
- Removed logfacility history, since it's not used for anything
- Avoid log callbacks if a debug level is not enabled in the log view


mathiascode:
Might have further ways to improve this, marking as draft.

PR #610, opened 2020-08-29
Temporary fix for Windows build not launching

mathiascode:
At some point we should do some launch tests on Windows and macOS packages too, to detect issues like this sooner.


mathiascode:
https://github.com/pyinstaller/pyinstaller/issues/5129

PR #623, opened 2020-09-01
Use TCP keepalive to monitor server connection

mathiascode:
Alternative solution to the previous ping system, which never worked as intended for me at least.


mathiascode:
> I think you may have meant to say "DEPRECATED"?

Good point. It would seem like the word "depreciated" was incorrectly chosen for deprecated messages many years ago. I'll replace all instances of the word.

PR #624, opened 2020-09-02
Use more common translation folder structure

mathiascode:
Changes the folder structure for .po files to a more common one, and generates .mo files on build time.

Since we already have a Launchpad team, I was thinking about setting up the translations page on https://translations.launchpad.net/nicotine+, to reduce the contribution barrier for regular users. Opinions?


kiplingw:
I think that's a great idea. You should already have permission.

PR #636, opened 2020-09-06
Drop urllib.requests import

mathiascode:
For the small number of http-related features we need, importing urllib.request is overkill. Since version checking is not done often, we can also cut down on startup import time a bit by importing http.client on demand.


kiplingw (changes requested):
@mathiascode, do we need to update our list of build or runtime dependencies?

mathiascode:
That shouldn't be necessary, urllib is just a higher level interface that uses http.client underneath.

PR #647, opened 2020-09-09
Fix issue regarding 100% CPU usage in networking code

mathiascode:
Due to a minor error in the input processing code, we would hit a "ValueError: Invalid file descriptor: -1" over and over, ending the loop prematurely and causing large CPU usage due to a lack of cooldowns in the exception handling code.


hboetes:
I'm particularly affected by this problem and currently test-driving this PR.

PR #658, opened 2020-09-12
Update README.md

hboetes:
Add instructions for setting up Debian derivatives. Add software-properties-common package, which is not installed by default on Debian.


hboetes:
Thanks for the improvements.

Issue #443, opened 2020-07-30
a separate program for database scanning


hboetes:
Would it be possible to have a separate program for database scanning?

This has the benefit the database scanner can be called from the CLI:
- to create cron jobs.
- to check the output for errors.
- to be able to restart nicotine without interrupting the scanning process.
- reduce load on front-end process

Another solution might be to add CLI options to nicotine.

mathiascode:
This would ideally be done by separating the whole scanning process into a new subprocess (with the multiprocessing module), and adding a CLI flag to start the scanning process separately from the rest of the program. I'd like to avoid creating completely new programs, mainly to avoid a situation like the Museek+ project, which is a nightmare to get started on (from both a user and a developer's perspective) and maintain due to the amount of abandoned junk included in the repository.

hboetes:
The current code is working so well I think we can drop this issue. I'm perfectly fine with closing it.

PR #691, opened 2020-10-02
Basic unit test for shares and CLI flag for scanning

mathiascode:
Partial fix for https://github.com/Nicotine-Plus/nicotine-plus/issues/443. The rescan flag is not documented yet, since it doesn't work well if Nicotine+ is already open (for now, you will have to quit Nicotine+, run `nicotine --rescan` and wait for it to complete, then start Nicotine+ again).

Also added some unit tests for shares.


debian/rules
```diff
@@ -35,9 +35,10 @@ SOURCE_DIR  = $(abspath $(PACKAGE_DIR)/../)
 #  embedded directly in the Python source...
 VERSION_UPSTREAM_DISTUTILS = $(shell sed -n 's/^version\s*=\s*"\(\S*\)"$$/\1/p' $(SOURCE_DIR)/pynicotine/utils.py)
 
+# Disable shares test, since shares are not saved for some reason...
```
mathiascode, line 38: @kiplingw Do you have any idea why this fails in autopkgtest? The dbs/shelves for shares are seemingly empty after a scan.

mathiascode: Seems to work when the tests run the second time, in debian/tests/control.

kiplingw: No idea. Where is the log? Possibly a race condition.

kiplingw:
Fails in [Groovy](https://launchpadlibrarian.net/500431509/buildlog_ubuntu-groovy-amd64.nicotine_2.1.2-202010030202~ubuntu20.10.1_BUILDING.txt.gz). At least I think this is the guilty PR.

Issue #531, opened 2020-08-12
running from source - missing reqs


starenka:
Can we please state required python libs to run from source? They're not stated in setup.py/requirement file. I run it now on using current pygobject & mutagen

```
mutagen~=1.45.1
PyGObject~=3.36.1
```
Is there a reason why the reqs are not in setup.py and therefore would be installed when doing `python setup.py install`?

mathiascode:
Main reasons: the native package manager of the OS is recommended for deps, nobody has asked for a requirements.txt file previously, and it's one less dependency list to keep up to date.

https://github.com/Nicotine-Plus/nicotine-plus/blob/master/doc/DEPENDENCIES.md
https://github.com/Nicotine-Plus/nicotine-plus/blob/2.0.1/doc/DEPENDENCIES.md

starenka:
I don't think encouraging people to `sudo python install` and using global site-packages is good idea (in general). The virtualenv approach should be suggested and thus you need those requirements to install in virtualenv. I don't know how packaging python app works (for linux/macos etc), but stating those reqs in setup.py/requirements.txt is standard python way to make these reqs installed in the env.

mathiascode:
Thanks for the heads up about virtualenv, sounds like a good approach. I'll look into it closer at some point.

starenka:
it basically boils down to:

```
git clone https://github.com/Nicotine-Plus/nicotine-plus.git /tmp/nicotine-plus
cd /tmp/nicotine-plus
python3 -m venv .env
# (now either use `source .env/bin/activate` which will allow you to run python and pip without the full path or use full paths for pip and python - will show you the latter one as it is more universal)
/tmp/nicotine-plus/.env/bin/pip install -r requirements.txt #if requiements not in setup.py
/tmp/nicotine-plus/.env/bin/python setup.py install
```

to run just: `/tmp/nicotine-plus/.env/bin/python tmp/nicotine-plus/nicotine`
to update, just pull from the repo and redo all steps from the `git clone`

starenka:
Nice! Cheers!

PR #692, opened 2020-10-02
Drop setuptools dependency

mathiascode:
The only reason I opted to use setuptools previously was due to `find_packages`. However, the `single-version-externally-managed` option, which we need to install files to the correct folders on the system, has caused a lot of pain so far, as it's not recognized by pip. Drop the setuptools dependency in favor of distutils, and use our own method of finding the correct packages and package data.

Pretty sure the introduction of setuptools also causes an issue where Nicotine+ is installed to /usr/lib/python3.X/dist-packages instead of /usr/lib/python3/dist-packages.

Fixes https://github.com/Nicotine-Plus/nicotine-plus/issues/531

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 
 # Ignore venv files
 venv/*
+.env/
 
 # Ignore packaging directories and files
 install.txt
--- a/debian/control
+++ b/debian/control
@@ -12,8 +12,7 @@ Build-Depends:
     python3 (>= 3.5),
     python3-taglib,
     python3-pep8-naming,
-    python3-pytest,
-    python3-setuptools
+    python3-pytest
 Vcs-Git: https://github.com/Nicotine-Plus/nicotine-plus.git
 Vcs-browser: https://github.com/Nicotine-Plus/nicotine-plus
 X-Python-Version: >= 3.5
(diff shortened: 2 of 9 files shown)


kiplingw:
Thanks @mathiascode. But isn't setuptools the successor to distutils?

mathiascode:
That's correct, and distutils will most likely be removed from the standard libraries in [a few years](https://www.python.org/dev/peps/pep-0632/). However, since issues [like this](https://github.com/pypa/distutils/issues/17) have occurred recently, and the whole situation is a bit uncertain due to patches by Debian and other distros right now, I'm more inclined to stick with distutils for now, and look into setuptools or completely different build systems when the time is right.

kiplingw:
Hmm, I'm not sure what to suggest. On the one had I don't want us to be dependant on a broken tool. On the other I don't think migrating back to a tool that is slated for removal is helpful either. We're not in any rush, so I'm going to leave this open for a week for open discussions and time for you to also think about it.

kiplingw:
@mathiascode, can you resolve your conflicts and verify CI?

mathiascode:
Done.

kiplingw:
Hmm, still breaking on Windows.

mathiascode:
Strange, I'll retry at some point tomorrow.

mathiascode:
Alright, got it sorted out.

PR #694, opened 2020-10-03
Fix Launchpad build

mathiascode:


kiplingw:
Thanks @mathiascode. Remember to always test every PR in autopkgtest. That will reduce the likelihood of these kinds of problems.

PR #701, opened 2020-10-04
Remove MediaDefender bots from default block list

mathiascode:
MediaDefender is defunct since 2009.


kiplingw:
Should we add something else in its place?

mathiascode:
Not that I know of.

PR #714, opened 2020-10-07
about.ui: use correct license

mathiascode:
The license should be GPL-3.0-or-later, not GPL-3.0-only.


kiplingw:
Excellent attention to detail, @mathiascode. Almost time for me to write you a letter of reference.

PR #719, opened 2020-10-09
Debian: install to /usr/lib/python3/dist-packages

mathiascode:
This seems to be the way to go to install Nicotine+ to the more generic /usr/lib/python3/dist-packages folder, instead of /usr/lib/python3.X/dist-packages. We want to avoid installing files to a minor version folder, as it breaks our deb packages on Debian installations if the minor Python 3 version doesn't match, and almost every other Python-based package doesn't install its files here.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -22,6 +22,8 @@ Rules-Requires-Root: no
 Package: nicotine
 Architecture: all
 Depends:
+    ${python3:Depends},
+    ${misc:Depends},
     python3-miniupnpc | miniupnpc (>= 1.9),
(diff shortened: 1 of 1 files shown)


kiplingw:
@mathiascode, if you haven't read [this](https://www.debian.org/doc/packaging-manuals/python-policy/) already, you will find it helpful.

mathiascode:
Thanks, I had read parts of it earlier, but looked through the whole thing now. I believe everything should work according to the guidelines once we've merged this PR.

PR #722, opened 2020-10-10
Production-ready macOS builds

mathiascode:
While these macOS builds aren't perfect yet (missing notification bubble support, spell checking, and a working tray icon), I think they're usable enough to be deemed stable.


grr:
i tried both the stable and unstable versions of the app on macOS Mojave (10.14.6) and the app wouldn't launch. no errors were found in the console, so not sure what the issue is. i had no problem, however, building and running the binary from the git repo.

mathiascode:
Likely because the macOS builds are created on Catalina (10.15), which sadly is the oldest image available on GitHub Actions.

grr:
According to the following links, support for version 10.14, 10.15, and 11.0 of the SDK is officially supported: 

* https://github.com/actions/virtual-environments/issues/461
* https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md#installed-sdks

godzfire:
Apologies for bumping an old thread, but I have 10.14 Mojave and can't update any further for 32bit application reasons.

I know that an actual prebuilt binary exists for 10.15, but nothing anymore for 10.14. I tried to use Homebrew as was suggested to create it, and everything runs and installs okay and it says that nicotine-plus installed, but there's no icon in my applications folder. I'm not sure what I'm doing wrong.

It also sounds like if I compiled the program myself on my 10.14 system it would work, but I have absolutely no idea how to do that. I'm very sorry but it's like trying to understand another language to me. 

Could I please get some assistance, or does someone have a built 10.14 version they'd be willing to share?

PR #724, opened 2020-10-10
Reddit plugin: drop feedparser dependency

mathiascode:
Requiring an external dependency to use a simple plugin like this is overkill.


mathiascode:
This would preferably be merged before Nic+ 2.1.2 is released, to deliver a functional plugin out of the box.

PR #734, opened 2020-10-12
Update AUTHORS.md

hboetes:
Add myself to the Authors


mathiascode:
Can you add the same text in pynicotine/gtkgui/ui/about/about.ui too?

PR #745, opened 2020-10-14
Launch file manager and audio player without external deps

mathiascode:
Cleaner solution than executing the xdg-open command.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -30,8 +30,7 @@ Depends:
     python3 (>= 3.5),
     python3-dbus,
     python3-gi,
-    python3-taglib,
-    xdg-utils
+    python3-taglib
 Recommends:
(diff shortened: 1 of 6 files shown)


debian/control
```diff
@@ -30,8 +30,7 @@ Depends:
     python3 (>= 3.5),
     python3-dbus,
     python3-gi,
-    python3-taglib,
-    xdg-utils
+    python3-taglib
```
kiplingw, line 33: Was this intentional in this PR? It doesn't really seem to have anything to do with the rest of your commits?

kiplingw (changes requested):

debian/control
mathiascode: It's intentional. The only tool we used in the xdg-utils package is xdg-open, which is now replaced with GIO's launch_default_for_uri method.

PR #754, opened 2020-10-18
Improvements on italian localization

M0Rf30:
As in title. There are some general improvements above all the translated strings and every single line is localized


kiplingw:
Thank you @M0Rf30.

PR #758, opened 2020-10-21
Create FAQ.md

hboetes:
Since certain problems occur more often I think it's a good idea to start maintaining a FAQ.

### DIFF
--- a/doc/FAQ.md
+++ b/doc/FAQ.md
@@ -1,0 +1,7 @@
+# FAQ
+
+This file contains frequently asked questions and their answers.
+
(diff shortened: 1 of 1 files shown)


mathiascode:
Don't merge yet, this might be a regression in Nicotine+.

mathiascode:
Since https://github.com/Nicotine-Plus/nicotine-plus/pull/760 was merged, I think we can close this.

hboetes:
I think we could use a faq, but not with these contents. I'll simply keep it in the back of my mind. But sure, let's close this one until I really have something to put in it.

PR #760, opened 2020-10-23
Attempt to send latin-1 encoded strings, fall back to utf-8

mathiascode:
This allows us to get more search results from Soulseek NS users, and reduces the number of download errors from these clients.


kiplingw:
Failure on unit test for Windows.

mathiascode:
Some unrelated error from the unit test for the update checker. Trying again later.

mathiascode:
Seems fine now. Looks like the issue was sorted out after the weekend.

PR #796, opened 2020-10-31
Notification and window improvements

mathiascode:
- Consolidate duplicate efforts for changing tab hilite icons into one function, "request_tab_icon"
- Remove Nicotine+ version number from window title. It takes up space and is not relevant information for everyday use.
- Don't modify window icon when new messages arrive, as this change is not guaranteed to work across desktop environments. Rely on other notification methods instead, such as window urgency, tray icon, notification bubbles and window title notifications.
- Code cleanups


kiplingw (changes requested):
Fails in lint.

PR #820, opened 2020-11-06
Bake UPnP functionality into Nicotine+

mathiascode:
- Since the miniupnpc dependency is a PITA to deal with due to missing Python bindings on certain systems, and the UPnP logic we need is fairly simple and short, I've opted to bundle a modified UPnP library instead.
- As one of the comments mentioned, unless port mappings are regularly renewed, they expire after 7 days on routers that use the IGDv2 protocol. Set the lease time to 24 hours to avoid leaving port mappings behind for too long, and regularly renew these when Nicotine+ is active.
- Automatically add port mapping if the user enables UPnP in settings, instead of requiring a restart.


kiplingw (changes requested):
Well done @mathiascode. Two things to think about.

1. Try running in Autopkgtest, as you should be with all PRs. Although it won't likely be able to find an NAT in there that it can forward, you'll at least be able to verify you got all the package and module dependencies correctly.

2. I'd suggest adding a configuration to allow the user to set what frequency the port mapping is renewed. A Gtk spinner for number of hours would probably be appropriate. Give it some thought.

mathiascode:
Thanks!
1. I can confirm that everything works as expected in Autopkgtest. I've made sure that we only use modules from the Python standard library.
2. This is a good idea, I'll add an option for that in the preferences window.

If you have time, could you check that this PR works as expected on your end? Although everything works well with my router, and the code follows the IGD spec, it wouldn't hurt to test on other router models.

kiplingw:
If you add the configuration options I can test through the _unstable_ PPA. It's an unstable PPA for a reason, so don't worry. Thanks Mathias.

mathiascode:
Alright, I've now added a configuration option.

mathiascode:
Let me check one thing though...

kiplingw:
> Alright, I've now added a configuration option.

Looks good!

Screenshot_2020-11-08_00-06-19

PR #826, opened 2020-11-08
Add 'Now Playing Sender' plugin

mathiascode:
This was requested in the IRC channel, and I thought it would be a good opportunity for me to get used to the plugin system.

The plugin currently contains an issue where the song is sent twice, which I haven't been able to resolve yet, but otherwise it works fine.


mathiascode:
Works as expected now. No more duplicate messages.

PR #834, opened 2020-11-11
Explicitly hide tray icon before exiting Nicotine+

mathiascode:
On Windows, it seems like the tray icon doesn't disappear until you move the cursor over the tray area. Hiding the tray icon before quitting Nicotine+ fixes this.


kiplingw:
So sad the system tray API is still so broken on Windows more than a decade after I abandoned that ridiculous OS.

PR #856, opened 2020-11-20
Update README.md

hboetes:
Add funny screenshot which explains where the name comes from.


mathiascode (changes requested):
Reserving for later, need to check something.

PR #860, opened 2020-11-21
Now Playing: port to GDBus

mathiascode:
GDBus seems to be the recommended alternative for interacting with dbus nowadays, and we eliminate yet another dependency.


pynicotine/nowplaying.py
```diff
@@ -198,9 +201,16 @@ def mpris(self, player):
                 log.add_warning(_("Auto-detected MPRIS player: %s."), player)
```
kiplingw, line 201: @mathiascode, is there not another method on the logger for informational messages as an alternative to relying on a warning?

mathiascode: Good catch, this should be a normal log message (add) instead of a warning (add_warning).

PR #872, opened 2020-11-26
Bug fixes for peer connections and transfers (part 1)

mathiascode:
This PR is the result of several days of debugging connectivity and transfer issues (many thanks to @alekksander for helping me test).

Some of the changes include:
- Properly close and clean up peer connections immediately after disconnecting from the server. Thanks to this, SoulseekQt clients are informed about us logging out, and the files they're downloading from the Nicotine+ client won't be permanently stuck on queued after we log in again.
- Don't attempt to send any QueueFailed messages after disconnecting from the server.
- Avoid adding two PeerConnection objects for a user, if we receive a direct and indirect connection attempt at the same time. This fixes issues where uploads would be stuck on "Cannot connect" due to messages being sent on the previous, obsolete connection, and never going anywhere.
- Avoid sending duplicate TransferRequest when attempting to upload files in certain cases.
- Prevent 0 byte file uploads from being stuck on "Initializing transfer"
- Fix minor argument inconsistencies for abort download function
- More comprehensive connection debug logging

### DIFF
--- a/pynicotine/gtkgui/downloads.py
+++ b/pynicotine/gtkgui/downloads.py
@@ -248,7 +248,7 @@ class Downloads(TransferList):
             elif key in ("R", "r"):
                 self.on_retry_transfer(widget)
             elif key == "Delete":
-                self.on_abort_transfer(widget, True, True)
+                self.on_abort_transfer(widget, clear=True)
             else:
                 # No key match, continue event
                 return False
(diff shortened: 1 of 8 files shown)


kiplingw:
Well done @mathiascode. If this works, I'm sure this is a major improvement that will benefit many users.

PR #906, opened 2020-12-16
Add initial documentation for maintainers

mathiascode:
Initial work on documentation for Nicotine+ maintainers, to increase the bus factor and help myself remember things.


mathiascode:
@kiplingw thoughts so far?

kiplingw:
Hmm, some overlap with `ReadMe.md`. Specifically on the unit testing and CI stuff. Maybe all maintainer related stuff should be moved into your new file?

mathiascode:
That's true, perhaps this file could be intended for both maintainers and developers/code contributors? Some of the information would certainly be important for someone thinking about diving into the code.

kiplingw:
I agree. I just don't want redundant information that has to be maintained in two places. Maybe you could move everything maintainer related into the new file, and leave a note in the `ReadMe.md` where maintainers should be looking?

mathiascode:
Sounds good to me.

mathiascode:
Updated. Anything you would like me to cover in the file?

kiplingw:
Looks good. Also make sure the Debian toolchain detects all the documentation files correctly and installs them wherever they should go. I think this is somewhere in `/usr/share/...`.

mathiascode:
Thanks. This is taken care of by distutils: https://github.com/Nicotine-Plus/nicotine-plus/blob/master/setup.py#L87-L103

PR #916, opened 2020-12-21
Introduce new headerbar-based design

mathiascode:
This PR introduces a more compact design utilizing headerbars.


mathiascode:
I'll add an option to enable server-side decorations (traditional titlebars) too, doesn't require that much effort to turn a header bar into a "toolbar".

mathiascode:
Initial reception of the design was good on IRC, and I consider this stable enough for general testing.

Issue #944, opened 2021-01-13
Feature Request: Clear filters button


redactedscribe:
### Describe the solution you'd like
A button to clear the filters. Currently, you can toggle the filter from being in effect by toggling "Filter Results" (which is useful), however to clear the filters, say 3 of the 5, one has to go through each and delete each value manually.  The button should both clear and apply the filter (to reveal all results).

### Additional context
I've noticed Windows' Nicotine+ `2.2.2` has a "Clear all search attempts" button, whereas `2.3.0.dev1` on Linux does not. Perhaps it was removed? I have no experience with the Linux version prior to this (I just noticed the button is available via the Preferences). Nevertheless, a button like this would also be useful to empty the search result filters.

redactedscribe:
Ideally, the inclusion of an "Apply"/"Set Filter" button to apply the filter would be useful too (rather than having to explain pressing Enter is how to do so).

PR #945, opened 2021-01-13
Rename "Search Filters" to "Search Result Filters" and improve help

redactedscribe:
Attempted to improve clarity of the filters' purpose by renaming Search Filters to Result Filters. Also tried to reword parts of the help file on the topic. Think I replaced all the strings referencing "search filters" that affect the UI. Open to feedback.

### DIFF
--- a/pynicotine/gtkgui/ui/dialogs/searchfilters.ui
+++ b/pynicotine/gtkgui/ui/dialogs/searchfilters.ui
@@ -2,7 +2,7 @@
 <interface>
   <requires lib="gtk+" version="3.18"/>
   <object class="GtkDialog" id="AboutSearchFiltersDialog">
-    <property name="title" translatable="yes">Search Filters</property>
+    <property name="title" translatable="yes">Result Filters</property>
     <property name="modal">1</property>
     <property name="window-position">center-on-parent</property>
     <signal name="delete-event" handler="on_hide" swapped="no"/>
@@ -17,8 +17,9 @@
         <child>
           <object class="GtkLabel">
             <property name="visible">1</property>
-            <property name="label" translatable="yes">You can use this to refine which results are displayed. The full results
-from the server are always available if you clear all the search terms.
+            <property name="label" translatable="yes">Result filters are used to refine which search results are displayed.
+
+To set the filter, press Enter after typing or clearing your terms. The filter applies immediately to the already received results, and to any more that are returned thereafer. The full results from the server are always available if you clear the filter of all terms. You do not need to perform another search to apply a different filter.
 
 You can filter by:
 
@@ -45,11 +46,7 @@ You can filter by:
 
 - Free slot:
       Show only those results from users which have at least
-      one upload slot free.
-
-- To set the filter, press Enter. This will apply to any existing results, and any more that are returned.
-- To filter in a different way, just set the relevant terms.
-- You do not need to do another search to apply a different filter.</property>
+      one upload slot free.</property>
             <property name="selectable">1</property>
           </object>
         </child>
--- a/pynicotine/gtkgui/ui/dialogs/searchfilters.ui
+++ b/pynicotine/gtkgui/ui/dialogs/searchfilters.ui
@@ -19,34 +19,51 @@
             <property name="visible">1</property>
             <property name="label" translatable="yes">Search result filters are used to refine which search results are displayed.
 
-To set the filter, press Enter after typing or clearing your terms. The filter applies immediately to the already received results, and to any more that are returned thereafer. The full results from the server are always available if you clear the filter of all terms. You do not need to perform another search to apply a different filter.
+Each list of search results has its own filter which can be revealed by toggling
+the Filter Results button. A filter is made up of multiple fields, all of which
+are applied when pressing Enter in any one of its fields. Filtering is applied
+immediately to results already received, and also to those yet to arrive. To
+view the full results again, simply clear the filter of all terms and re-apply
+it. As the name suggests, a search result filter cannot expand your original
+search, it can only narrow it down. To broaden or change your search terms,
+perform a new search.
 
 You can filter by:
 
 - Included text:
-      Files are shown if they contain this text.
-      Case is insensitive, but word order is important.
-      &apos;Spears Brittany&apos; will not show any &apos;Brittany Spears&apos;
+      - Files and folders containing this text will be shown.
+      - Case is insensitive, but word order is important:
+        &apos;Spears Brittany&apos; will not show any &apos;Brittany Spears&apos;
 
 - Excluded text:
-      As above, but files will not be displayed if the text matches
+      - As above, but files and folders are filtered out if the text matches.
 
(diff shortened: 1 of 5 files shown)


pynicotine/gtkgui/ui/dialogs/searchfilters.ui
```diff
@@ -2,7 +2,7 @@
 <interface>
   <requires lib="gtk+" version="3.18"/>
   <object class="GtkDialog" id="AboutSearchFiltersDialog">
-    <property name="title" translatable="yes">Search Filters</property>
+    <property name="title" translatable="yes">Result Filters</property>
```
mathiascode, line 5: I feel naming this "Search Result Filters" would make it easier to understand the context at a quick glance, mainly in the main menus.

pynicotine/gtkgui/ui/dialogs/searchfilters.ui
```diff
@@ -17,8 +17,9 @@
         <child>
           <object class="GtkLabel">
             <property name="visible">1</property>
-            <property name="label" translatable="yes">You can use this to refine which results are displayed. The full results
-from the server are always available if you clear all the search terms.
+            <property name="label" translatable="yes">Result filters are used to refine which search results are displayed.
+
+To set the filter, press Enter after typing or clearing your terms. The filter applies immediately to the already received results, and to any more that are returned thereafer. The full results from the server are always available if you clear the filter of all terms. You do not need to perform another search to apply a different filter.
```
mathiascode, line 22:
```suggestion
To set the filter, press Enter after typing or clearing your terms. The filter
applies immediately to the already received results, and to any more that are
returned thereafer. The full results from the server are always available if
you clear the filter of all terms. You do not need to perform another search to
apply a different filter.
```

pynicotine/gtkgui/ui/dialogs/searchfilters.ui
redactedscribe: I was thinking of this too, but though it may be too long. But I'm okay with this compromise.

redactedscribe:
I'll get to this in the coming hours.

pynicotine/gtkgui/ui/dialogs/searchfilters.ui
mathiascode: It's roughly the same length as the other help menu items, so it shouldn't be an issue.

pynicotine/gtkgui/ui/dialogs/searchfilters.ui
redactedscribe: By the way, the section on Size says:
```
k or m to specify kilo or megabytes. &gt;10M will find files larger than 10 megabytes. &lt;4000k will find files smaller than 4000k.
```
Should these be in binary byte types instead? kibi/mebi? If so, it calls into question the shorthand notation of `k` and `m` stated above. Though wrong, perhaps those should be left as-is, rather than requiring typing of `kib`/`KiB` for example, just because it's simpler.

redactedscribe:
I'll improve some of the Help now.

redactedscribe:
Some notes:

Sorting by Folder (or Filename or Size) in the search results only sorts the (in this case) folders per user:
```
userA: f6 f5 f3
userB: f4 f2 f1
```
will sort to
```
userA: f3 f5 f6
userB: f1 f2 f4
```
instead of a global:
```
userB: f1 f2
userA: f3
userB: f4
userA: f5 f6
```

Actually, I just realized this is because of the result grouping mode in use. "Ungrouped" works as I'd typically expect. A note of this could be made in a help section at some point, even just hinting that the grouping mode affects column sorting.

---

The Search Filters help reads that using `<` or `>` matches files less/greater than the specified value, but in fact it is less/greater than _or equal to_. I think the behavior of returning `<=` results is more user-friendly so I just reworded the help to be accurate. Just wanted to mention this in case you were unaware. However, similar to my `k`/`m` comment further up, it's not accurate to type `<` since it actually produces `<=`. But again, for simplicity's sake this may be worth overlooking.

---

The help button on the Search Filters section is round, when the help buttons on chat rooms or private messages is square.

---

Transfer Statistics may suit View better than Help. For example, using another divider and prepending it to the View menu.

pynicotine/gtkgui/ui/dialogs/searchfilters.ui
redactedscribe: Noting that `b` also works for bytes.

redactedscribe:
Beyond rewording the help and making it accurate, a Country section was added too.

Upon testing it now, the popup is beginning to get quite tall. If you have the inspiration, you could do something similar to the other help popups and allow the text to reflow and the window to be scrolled.

redactedscribe:
In case you were wondering, I couldn't find any other apostrophes that were not `&apos;`s in `*.ui` files. Used `[a-z]+'([a-z][a-z]|[a-z])`.

mathiascode:
> Some notes:
> 
> Sorting by Folder (or Filename or Size) in the search results only sorts the (in this case) folders per user:
> 
> ```
> userA: f6 f5 f3
> userB: f4 f2 f1
> ```
> 
> will sort to
> 
> ```
> userA: f3 f5 f6
> userB: f1 f2 f4
> ```
> 
> instead of a global:
> 
> ```
> userB: f1 f2
> userA: f3
> userB: f4
> userA: f5 f6
> ```
> 
> Actually, I just realized this is because of the result grouping mode in use. "Ungrouped" works as I'd typically expect. A note of this could be made in a help section at some point, even just hinting that the grouping mode affects column sorting.

Perhaps the fast configure dialog could be extended with a simple UI section that mentions this, and allows you to choose the list mode? I'd like to avoid separate help sections when possible, as 1. they tend to become outdated at some point (like the now removed Nicotine+ manual that was 10+ years out of date), and 2. my previous experience from other projects has shown that people hardly read such sections/faqs/wikis/messages etc.

> The Search Filters help reads that using `<` or `>` matches files less/greater than the specified value, but in fact it is less/greater than _or equal to_. I think the behavior of returning `<=` results is more user-friendly so I just reworded the help to be accurate. Just wanted to mention this in case you were unaware. However, similar to my `k`/`m` comment further up, it's not accurate to type `<` since it actually produces `<=`. But again, for simplicity's sake this may be worth overlooking.

I'll check what's going on in the source code, but it's probably not the best idea to drastically change behavior at this point.

> The help button on the Search Filters section is round, when the help buttons on chat rooms or private messages is square.

I originally made it round to give some visual contrast to the numerous square buttons present on the same page, but things have been simplified since. I'll probably revert this when implementing https://github.com/Nicotine-Plus/nicotine-plus/issues/944.

> Transfer Statistics may suit View better than Help. For example, using another divider and prepending it to the View menu.

Decided to move it to the modes section.

mathiascode:
> Upon testing it now, the popup is beginning to get quite tall. If you have the inspiration, you could do something similar to the other help popups and allow the text to reflow and the window to be scrolled.

Will do.

PR #958, opened 2021-01-20
Debian: remove deprecated get-orig-source target

mathiascode:
- Remove the "get-orig-source" target, as it is deprecated. Add updated commands to PACKAGING.md utilizing Python's sdist and "mk-origtargz" from devscripts. https://wiki.debian.org/onlyjob/get-orig-source
- Add Debian packaging workflow to GitHub Actions, to allow for easier inspection of logs and built packages.
- Remove old "parallel" build option. This seems to be from a time when we used to ship .png files, nowadays it doesn't seem to make a noticeable difference.

@kiplingw Since you are the author of a large part of the "Debianization", and have more experience with this process, I would appreciate your feedback on these changes.

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -1,8 +1,51 @@
-name: Packaging (Windows and macOS)
+name: Packaging (Debian, Windows and macOS)
 
 on: [push, pull_request]
 
 jobs:
+
+  debian:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+
+      - name: Install dependencies
+        run: |
+          BUILD_DEPS=$(perl -ne 'next if /^#/; $p=(s/^Build-Depends:\s*/ / or (/^ / and $p)); s/,|\n|\([^)]+\)//mg; print if $p' < debian/control)
+
+          sudo apt update
+          sudo apt -y install $BUILD_DEPS build-essential devscripts sbuild
+
+      - name: Build Debian package
+        run: |
+          # Generate tarball
+          python setup.py sdist
+          mk-origtargz dist/nicotine-plus-*.tar.gz
+
+          # Build source package
+          debuild -S -sa -us -uc
+
+          # Build binary
+          debuild -b -sa -us -uc
+
+      - name: Prepare artifacts
+        run: |
+          mkdir ../build/
+          cp -Lr ../nicotine_* ../build/
+
+      - name: Archive artifacts
+        uses: actions/upload-artifact@v1
+        with:
+          name: debian-package
+          path: ../build/
 
   windows:
     runs-on: windows-latest
@@ -42,13 +85,13 @@ jobs:
           cp -r packaging/windows/Nicotine*.exe installer/
 
       - name: Archive installer artifacts
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v2
         with:
           name: windows-${{ matrix.arch }}-installer
           path: installer
 
       - name: Archive package artifacts
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v2
         with:
           name: windows-${{ matrix.arch }}-package
           path: dist
@@ -80,7 +123,7 @@ jobs:
           packaging/macos/create-dmg.sh
 
       - name: Archive installer artifacts
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v2
         with:
           name: macos-installer
           path: dmg
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -19,10 +19,9 @@ jobs:
 
       - name: Install dependencies
         run: |
-          BUILD_DEPS=$(perl -ne 'next if /^#/; $p=(s/^Build-Depends:\s*/ / or (/^ / and $p)); s/,|\n|\([^)]+\)//mg; print if $p' < debian/control)
-
+          sudo sed -Ei 's/^# deb-src /deb-src /' /etc/apt/sources.list
           sudo apt update
-          sudo apt -y install $BUILD_DEPS build-essential devscripts sbuild
+          sudo apt build-dep .
 
       - name: Build Debian package
         run: |
--- a/debian/control
+++ b/debian/control
@@ -1,9 +1,10 @@
 Source: nicotine
 Section: net
 Priority: optional
-Maintainer: Kip Warner <kip@thevertigo.com>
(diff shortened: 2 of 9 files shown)


debian/control
```diff
@@ -1,9 +1,10 @@
 Source: nicotine
 Section: net
 Priority: optional
-Maintainer: Kip Warner <kip@thevertigo.com>
+Maintainer: Nicotine+ Team <nicotine-team@lists.launchpad.net>
 Standards-Version: 3.9.7
```
kiplingw, line 5: This should be bumped.

.github/workflows/packaging.yml
```diff
@@ -1,9 +1,52 @@
-name: Packaging (Windows and macOS)
+name: Packaging (Debian, Windows and macOS)
 
 on: [push, pull_request]
 
 jobs:
 
+  debian:
+    runs-on: ubuntu-20.04
```
kiplingw, line 8: Is 20.10 supported?

debian/copyright
```diff
@@ -1,10 +1,10 @@
 Format: http://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 Upstream-Name: nicotine
-Upstream-Contact: Kip Warner <kip@thevertigo.com>
+Upstream-Contact: Nicotine+ Team <nicotine-team@lists.launchpad.net>
 Source: https://github.com/Nicotine-Plus/nicotine-plus/
 
 Files: *
-Copyright: Copyright (C) 2020 Nicotine+ Contributors
+Copyright: Copyright (C) 2020 Nicotine+ Team
```
kiplingw, line 7: Copyright notice should be a range of Start-2021. Probably a good idea to update that everywhere.

debian/rules
```diff
@@ -1,61 +1,18 @@
...
 # 	Nicotine+
+#	Copyright (C) 2021 Nicotine+ Team
 #	Copyright (C) 2016 Kip Warner. Released under the GPLv3 or later.
 #
 
 # Output every command that modifies files on the build system...
 DH_VERBOSE = 1
 export DH_OPTIONS=-v
 
-# Set flags for how the package will be built. Note separated with white
```
kiplingw, line 11: I am fine in principal with removing the `get-orig-source` target, but I'd like it if you could verify with a DPM expert (not me) whether this is appropriate here. The reason is that the deprecation notice for the target I think contemplates official Debian packages where downstream packaging and upstream source are in separate locations. So you may want to look into what the best practise is for unofficial packages.

doc/PACKAGING.md
```diff
@@ -26,8 +26,8 @@ Unstable and stable PPAs are already provided for pre-compiled packages, as desc
 Start by generating the "upstream" tarball:
 
 ```console
-cd nicotine_source
-./debian/rules get-orig-source
+python setup.py sdist
+mk-origtargz dist/nicotine-plus-*.tar.gz
```
kiplingw, line 30: Good idea to have this here since the process is changing.

kiplingw (changes requested):

.github/workflows/packaging.yml
mathiascode: Only LTS Ubuntu releases are supported.

debian/control
mathiascode: I'll do this sometime in March, before end of standard support for Xenial.

debian/copyright
mathiascode: If this is a requirement in Debian copyright files, I'll change it here. My understanding is that specifying the start year alone is usually enough, and I'd like to avoid needless updates elsewhere.

kiplingw: It's actually not a Debian specific requirement, but a legal requirement for all the source. A range is [needed](https://epiphany.law/articles/copyright/copyright-dates-single-year-or-range) too because there have been various releases over the years. Fortunately the following shell snippet will help you to update them all in a single command:

```bash
find -type f -name "*.py" -execdir sed -i 's/Copyright (C) 2020 Nicotine+ Contributors/Copyright (C) first_release-2021 Nicotine+ Team/g' {} \;
```

debian/rules
mathiascode: Do you know any acquaintance or other person who could verify this?

I'd personally prefer to diverge as little as possible from a future downstream Debianization, minimize version parsing and hardcoding, and keep things simple in general.

kiplingw: > Do you know any acquaintance or other person who could verify this?
> 
> I'd personally prefer to diverge as little as possible from a future downstream Debianization, minimize version parsing and hardcoding, and keep things simple in general.

Best guy to ask is the one and only DPM guru I ask whenever I have questions, @pabs3.

kiplingw: Also ask in #packaging on irc.oftc.net. That's where I usually find him and others.

pabs3:
I would use the GitHub tarballs and `git archive` and rather than the
PyPI tarballs and `setup.py sdist`, but either one works.

Instead of manually parsing Build-Depends, you can get apt to install
both build-essential and the build-deps for the current dir in one cmd:

apt-get build-dep .

Not sure if GitHub actions supports multiple architectures but either
way I think I would split building the source package from building the
binary packages.

The addition of the appstream build-dep doesn't seem to be needed?

I recommend an underscore in the sbuild invocation:

sbuild ../nicotine_*.dsc

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

mathiascode:
> I would use the GitHub tarballs and `git archive` and rather than the PyPI tarballs and `setup.py sdist`, but either one works.

I think I'll keep using the sdist tarballs, since this makes it easier to verify that the source distributions contain all necessary files to build and run Nicotine+. These also go up on PyPi later for installation via pip.

> Instead of manually parsing Build-Depends, you can get apt to install both build-essential and the build-deps for the current dir in one cmd: apt-get build-dep .

Excellent, thanks! This manual parsing is something I wanted to get rid of previously.

> Not sure if GitHub actions supports multiple architectures but either way I think I would split building the source package from building the binary packages.

As our current binary packages are universal for all architectures and distribution releases, I don't think it's worth changing this right now. The PPA will continue to be the preferred way of installing Nicotine+, while the GitHub actions workflow mostly exists for verification of the whole process and its outcome.

I don't think multiple architectures are supported on GitHub Actions, unless you start using your own runners.

> The addition of the appstream build-dep doesn't seem to be needed?

Forgot to mention this in the PR, but it seems to be needed to generate translated metainfo files.
`msgfmt: cannot locate ITS rules for files/org.nicotine_plus.Nicotine.metainfo.xml.in`

> I recommend an underscore in the sbuild invocation: sbuild ../nicotine_*.dsc

I'll change this.

Thank you for chipping in, I appreciate it.

debian/rules
mathiascode: Thanks! I'll keep this in mind for later.

mathiascode:
I've applied the suggested changes. If everyone is happy, I'll merge this.

kiplingw (approved):

PR #959, opened 2021-01-21
Use PyPi for version checker instead of GitHub

mathiascode:
Now that Nicotine+ is available on PyPi, it seems like a more reliable source for checking the latest Nicotine+ version long-term wise.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -86,5 +86,4 @@ jobs:
 
       - name: Unit tests
         run: |
-          # Disable version test due to macOS machines hitting GitHub's rate limits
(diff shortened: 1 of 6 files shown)


kiplingw:
@mathiascode, remember to check to make sure its working probably with `uscan(1)`. This is how Debian is notified that we have a newer upstream release.

mathiascode:
Using PyPi instead of GitHub in the Debian watch file is done in this commit: https://github.com/Nicotine-Plus/nicotine-plus/pull/958/commits/d9e079c4ed8a09090ce9e03fc1bafed68c7e5d9c

Tested and working.

Issue #956, opened 2021-01-19
The queue limits the number of simultaneous uploads.


82d:
Nicotine+ 2.2.2:  
Operating System/Distribution: Arch  Linux

For example, I want to set conditions so that my files can be downloaded simultaneously by many users at a low speed. I set the speed in the settings, remove all queue restrictions. In practice, only two users are automatically downloaded, while the rest are waiting in line.If I select the others in the queue and click Continue, they continue to download, but only up to one file, and then they will queue again. How can I remove or expand the queue?

82d:
in the beta version, something seems to have improved. and now it automatically gives it to seven people at once, but no more. I would like more than seven if possible.

mathiascode:
There shouldn't be any limit, but I will review the PR for potential errors soon.

PR #960, opened 2021-01-21
Optimize file uploads

mathiascode:
- Remove a lot of needless calls to check_upload_queue, and check the upload queue every five seconds instead of each time a part of a file is uploaded.
- If a maximum upload speed is set in the preferences, and a limit for queuing files isn't set, multiple files are now uploaded even when the maximum upload speed is reached. Closes https://github.com/Nicotine-Plus/nicotine-plus/issues/956.


mathiascode:
Since #1017 is merged, I will look into reviving this PR.

PR #995, opened 2021-02-12
Implement user join/leave events

mendelgusmao:


mathiascode:
Thanks! I'll look into these events a bit closer soon.

mathiascode:
I'd prefer to keep the existing join and leave notifications local only, and add new notifications for users, e.g. UserJoinChatroomNotification and UserLeaveChatroomNotification. There's also a current issue in Nicotine+ where the chatroom leave notification is triggered as expected (https://github.com/Nicotine-Plus/nicotine-plus/blob/master/pynicotine/gtkgui/chatrooms.py#L1423), but the join notification isn't assigned anywhere.

mendelgusmao:
Alright, I'll implement new notification triggers and update the PR.

mathiascode (approved):

PR #1000, opened 2021-02-20
Store chosen away state to config and load it upon connection

mendelgusmao:


mendelgusmao:
**DO NOT MERGE**

During tests I found a very odd behavior: when **_apply_away_state()** is called in **on_connect()**, Nicotine "freezes" with **Connected to server server.slsknet.org:..., logging in...** on status.

I tested with 3.0.0+cherry-pick of this commit. Something similar is happening randomly in pure origin/master.

pynicotine/gtkgui/frame.py
```diff
@@ -905,15 +906,22 @@ def on_connect(self, *args):
             self.np.servertimer.cancel()
             self.np.servertimer = None
 
+        self.away = self.np.config.sections["server"]["away"]
```
mathiascode, line 909: It would be better to set this at https://github.com/Nicotine-Plus/nicotine-plus/blob/master/pynicotine/pynicotine.py#L1050, since the UI is initialized with the away status at that point. A SetStatus message is also sent there.

mendelgusmao: I agree!

But I think it's still no good to put it in login(), because it would make the "chosen away state" to be set at every login, conflicting with the real away state.

I moved it to NicotineFrame initialization.

pynicotine/gtkgui/frame.py
```diff
@@ -905,14 +909,20 @@ def on_connect(self, *args):
             self.np.servertimer.cancel()
             self.np.servertimer = None
 
+        if self.away:
+            self._apply_away_state(False)
```
mathiascode, line 913: As far as I can tell, this call is now redundant, and also conflicts with the "in progress" status set earlier in the function (...). The away state is already set in init_interface, called after connecting and logging in.

mendelgusmao: That's true. Fixed in 1f09f60

mathiascode (approved):
Thanks for the contributions!

mendelgusmao:
Thanks!

PR #1005, opened 2021-02-21
WIP: Use Deluge port checker, show result in a separate icon

droserasprout:
It seems like "Check port status" button in Settings/Server frame opens a dead link for some time. This PR replaces this link with call to API provided by Deluge infrastructure. Depending on was this call successful or not OK or warning sing appears next to the button.Code was also poorly ripped from Deluge GTK client. This operation needs to be non-UI-blocking before this PR could be merged and some tests would also be great. But before continuing to digging into this I would ask community if it's a good idea to depend on Deluge's API? Maybe there are more suitable solutions for this issue?

### DIFF
--- a/pynicotine/gtkgui/settingswindow.py
+++ b/pynicotine/gtkgui/settingswindow.py
@@ -39,7 +39,6 @@ from pynicotine.gtkgui.dialogs import save_file
 from pynicotine.gtkgui.utils import FileChooserButton
 from pynicotine.gtkgui.utils import initialise_columns
 from pynicotine.gtkgui.utils import load_ui_elements
-from pynicotine.gtkgui.utils import open_uri
 from pynicotine.gtkgui.utils import update_widget_visuals
 from pynicotine.logfacility import log
 from pynicotine.utils import unescape
@@ -167,7 +166,6 @@ class ServerFrame(BuildFrame):
         except Exception as exc:
             log.add(_('Error testing listen port: %s' % exc))
             return False
-
 
     def on_toggle_upnp(self, widget, *args):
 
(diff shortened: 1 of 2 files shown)


mathiascode:
Does the link not load at all? It still works on my end, and also seems to be used in the official client.

I would prefer a port checking system that doesn't send you to a website though (and preferably with https traffic, unlike the slsknet.org one). Not sure if it's a good idea to rely on Deluge's servers, but I'll check what other solutions exist out there.

droserasprout:
That's very strange. I've checked this link once again and yes, it's working now! My 2-3 previous attempts at different dates including yesterday evening have all led to 404 error.

GoProCodes:
i agree very good code yes https is black i know peple hyes

GoProCodes:
i agree very good code yes https is black i know peple hyes

GoProCodes:
i agree very good code yes https is black i know peple hyes

GoProCodes:
i agree very good code yes https is black i know peple hyes

GoProCodes:
i agree very good code yes https is black i know peple hyes

GoProCodes:
i agree very good code yes https is black i know peple hyes

mathiascode:
I'll close this PR for now. If the https-only issue becomes more widespread, we can consider looking into alternatives, or perhaps attempt to convince Soulseek maintainers to use a proper SSL certificate on the tools.slsknet.org subdomain.

mathiascode:
I enabled HTTPS-only mode in Firefox, and it seems to offer the option to continue to a http-only site. The 404 error only seems to appear if you've previously attempted to access the https-variant of tools.slsknet.org, and bypassed the SSL warning.

PR #1018, opened 2021-03-09
Set correct search column for interests (un)recommendations treeviews

timovwb:
When typing in the recommendations and unrecommendations treeviews in the interests tab, the quicksearch will search in the score column. With this change it'll search in the actual item column.


mathiascode:
Good catch!

PR #1336, opened 2021-04-14
Stop using blocking dialog functions in preparation for GTK 4

mathiascode:
GtkDialog.run() has been removed in GTK 4. Replace it with GtkDialog.show() and the "response" signal.

See https://developer.gnome.org/gtk4/stable/ch41s02.html#id-1.7.4.4.81


mathiascode:
Confirmed working on GTK 3.18, the oldest version we support.

PR #1340, opened 2021-04-16
Minor change to Test Replier files

Inso-m-niaC:


pynicotine/plugins/examplars/testreplier/PLUGININFO
```diff
@@ -1,4 +1,4 @@
 Version = "2008-10-29r00"
```
mathiascode, line 1: Increase this.

PR #1363, opened 2021-04-29
Use argparse module to parse command line arguments

mathiascode:
- Use the argparse module from the stdlib to simplify our command line argument parsing
- Modify our exit code handling to fall in line with other popular Python programs


mathiascode:
Tested on Ubuntu 16.04, Arch, Windows and macOS.

PR #1370, opened 2021-05-08
Use pytest for integration tests

mathiascode:
This PR simplifies our testing process by moving the integration tests to pytest, allowing for startup tests on Windows, macOS and other distributions in the process, if desired. It also fixes an issue where the Ubuntu unit test workflows used the system Python version instead of versions requested in actions/setup-python.


kiplingw:
@mathiascode, is all CI testing no longer performed via DEP-8 anymore? Or is it still, you've just changed how it's invoked during autopkgtest invocation? Hopefully only the latter.

mathiascode:
> @mathiascode, is all CI testing no longer performed via DEP-8 anymore? Or is it still, you've just changed how it's invoked during autopkgtest invocation? Hopefully only the latter.

Testing still works the same. The main difference is that the startup test is also invoked before/during the build process (since running Nicotine+ directly from a folder is desired functionality) in addition to post-installation with autopkgtest.

kiplingw:
Ok that's fine. You just don't want to remove the DEP-8 CI testing because that is the only way to properly test installed artifacts.

Issue #400, opened 2020-07-16
proxy settings


danrobi11:
Hello 
Running new 2.0.0 flatpak version
Why theres no proxy settings?
This is a file-sharing app, it needs proxy settings!

mathiascode:
I don't think this is something we want to support inside Nicotine+ itself, since (to my knowledge) SOCKS5 does not support port forwarding. A VPN that supports port forwarding is recommended: https://www.reddit.com/r/VPNTorrents/comments/oqnnrq/list_of_vpns_that_allow_portforwarding_2021/

PR #1390, opened 2021-05-17
Update list of active contributors

mathiascode:
Since almost a year has passed now, it's time to update the list of active contributors.

@lene Thanks again for your hard work on the Python 3 port! You're always welcome back if you ever feel like working on Nicotine+ again. :)

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -15,11 +15,6 @@ Kip Warner
 - Debianization
 - [kip(at)thevertigo(dot)com]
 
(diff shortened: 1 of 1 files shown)


lene:
Thanks! :) I've been eyeing #400, which would be a really useful improvement IMHO, so unless someone else does it I might be back one day.

mathiascode:
Feel free to tackle that issue, it's not on my radar anytime soon.

PR #1391, opened 2021-05-18
Update Italian localization

M0Rf30:
As in title.
Regards

### DIFF
--- a/po/it.po
+++ b/po/it.po
@@ -2,8 +2,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2021-05-10 00:13+0200\n"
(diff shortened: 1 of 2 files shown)


mathiascode:
Thanks! One question, how do you generate the nicotine.pot file?

M0Rf30:
As you previously suggested with:
`python3 setup.py update_pot`

mathiascode:
That's fine. I suppose some subtle system differences change the order of the strings somehow.

mathiascode:
The strings should be sorted in the latest commit. Can you check if `python3 setup.py update_pot` yields the same result as the current `nicotine.pot` file?

M0Rf30:
Ok I fetched latest commits, update the pot file and here the diff after a test commit
```
-"POT-Creation-Date: 2021-05-18 18:07+0300\n"
+"POT-Creation-Date: 2021-05-18 17:17+0200\n"
```
So it should be ok

PR #1394, opened 2021-05-20
Change freenode link to OFTC

mathiascode:
This change isn't set in stone yet, but you'll likely find me on OFTC if/when we migrate away from freenode.

Context: https://www.kline.sh/

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,16 +124,16 @@ Unstable macOS installers are generated after every commit to the master branch,
 - [Download Installer](https://github.com/Nicotine-Plus/nicotine-plus/actions?query=branch%3Amaster+event%3Apush+is%3Asuccess+workflow%3A%22Packaging%22)
 
 ## Building from git (Unstable)
+
 For more experienced users and developers who want to test the latest and greatest changes in Nicotine+, building from git is described in [RUNFROMGIT.md](doc/RUNFROMGIT.md). Also read the next section about getting involved.
 
 # Getting Involved
-Please come and join us in the `#nicotine+` channel on Freenode!
 
 If you'd like to contribute, you have a couple of options to get started:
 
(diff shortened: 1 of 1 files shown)


kiplingw:
@mathiascode, let's hold the phone on this. I want to ask around to determine whether OFTC or Libera.chat servers are more appropriate.

mathiascode:
Sure, sounds good to me.

kiplingw:
@mathiascode, if you want to re-open this PR or start a new one, the trend I'm seeing seems to be to migrate to Libera.Chat after looking and asking around. My guess is because that's where all the Freenode staff moved to.

mathiascode:
Done in https://github.com/Nicotine-Plus/nicotine-plus/commit/0dc369989065f681400cce8676f9a0dfef3e4a0d

Could you unrestrict the freenode channel for some time, and update the topic to mention the migration? The current ban-kicks don't provide any information about the migration (except for "You are banned"), which confuses people.

kiplingw:
Sorry, I didn't realize that. @pabs3, I thought I had already done this, but I'll try again.

mathiascode:
Screenshot from 2021-05-21 21-13-15

This is on HexChat.

kiplingw:
It looks like it might be [hardcoded](https://github.com/hexchat/hexchat/blob/90c91d6c9aa048eff8f8f8f888d37a21fd714522/src/common/textevents.in#L16).

pabs3:
The method I suggested is indeed not ideal, since you only get the
message the first time you join the channel. It might be better to set
chanserv guard on, set the topic and entrymsg and then set the channel
as moderated so that people cannot talk on it.

An alternative solution would be to forward/redirect the channel to
another channel like ##moved_to_libera or similar.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

PR #1431, opened 2021-06-19
Add option to show locked/privately shared files in search results

mathiascode:
For the sake of transparency, I've added the option to show privately shared/locked files from other clients in search results. Since this is a controversial Soulseek feature, these files will not appear in results unless explicitly enabled in the search preferences.

For more information, see: https://groups.google.com/g/soulseek-discussion/c/vAnknGtihxg/m/pU2ItYSMAQAJ

### DIFF
--- a/doc/SLSKPROTOCOL.md
+++ b/doc/SLSKPROTOCOL.md
@@ -1,6 +1,6 @@
 # Soulseek Protocol Documentation
 
-Last updated on June 18, 2021
+Last updated on June 20, 2021
 
 As the official Soulseek client and server is proprietary software, this documentation has been compiled thanks to years of reverse engineering efforts. To preserve the health of the Soulseek network, please do not modify the protocol in ways that negatively impact the network.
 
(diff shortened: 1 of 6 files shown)


BachoSeven:
Is there any visual indication that the locked files are such from the search interface?

mathiascode:
Such files are currently prefixed with "[PRIVATE FILE]". A padlock icon would probably look better, but I had issues getting it working for now.

BachoSeven:
Hei, I thought I'd mention it here first but you might want me to post this as a separate issue:
two times out of two, while searching for a file which I know to be locked and/or interacting with an user whom I know has all locked files(`CaptainNoFuture`), I've had the log window flooded with this messages:
```
22:06:19 Peer init message type 4 size 4 contents bytearray(b'\x04\x00\x00\x00') unknown
```
So I assumed it's related to this...

mathiascode:
@BachoSeven Is this still happening on the latest master?

mathiascode:
I've pushed more changes since my last comment.

BachoSeven:
Sorry, I'll try this out now

BachoSeven:
It seems to be fixed, I successfully found the locked file(which shows up as expected with `[PRIVATE FILE]`, and browsing the user doesn't flood the logs anymore. Thanks for the fix

BachoSeven:
btw, browsing the user doesn't show the private/locked files, which is intentional right? not sure if there's a use case for wanting to turn that on through an option as well, but it might be useful for someone, just throwing the idea out there.

mathiascode:
> btw, browsing the user doesn't show the private/locked files, which is intentional right? not sure if there's a use case for wanting to turn that on through an option as well, but it might be useful for someone, just throwing the idea out there.

I've now implemented this.

PR #1451, opened 2021-07-15
nowplaying.py: Add hours to get_length_time()

Inso-m-niaC:


Inso-m-niaC:
I am pretty sure that the original if check was not needed. Anything other than an integer would raise a TypeError when divided by the caller.

mathiascode (approved):

PR #1464, opened 2021-07-31
Allow aliased commands to pass arguments

Inso-m-niaC:


Inso-m-niaC:
I didn't account for plain text being the value of the alias..

PR #1524, opened 2021-09-03
Improve layout of Interests tab

### LINKED ISSUE #1523: Show Similar Users button disappears off window edge due to widget alignment issues (Interests tab)
Nicotine+ version: 3.1.2.dev1  •  GTK 3.22.11
Operating System/Distribution: MX18, MX19, others

### Describe the bug
As title.

UI bug always present when the viewport is narrow such as when side-by-side (see screenshot below)

### Expected behavior

The Show Similar Users button should be aligned to the left, so that it is always visible.

Also note all three panels on the Interests user interface have strange alignment problems that make it unusable unless maximized. The left-hand 'My Interests' panel does not need to be resizable.

It would perhaps be more appropriate to place the button in the left-hand 'My Interests' panel

1. Goto Interests tab
2. Unmaximize the main window set to narrow width
3. Adjust the width of the panels
- Many alignment problems are evident

### Additional context
This UI bug is an accessibility problem.

_The mouse cursor indicates where the missing button should be:_
nicotine-plus Interests tab layout problems on small screens Show Similar Users button disappears off edge
Implementing a fix for this widget alignment issue would probably need issue #1522 to be addressed at the same time (Edit: Done, unrelated to this issue).

slook:
1. Relocate 'Show users with similar interests' button before label so doesn't disappear off the screen on narrow viewports [fixes #1523].
2. Remove all instances of depreciated "xalign" property (note this causes labels to be central which looks okay here)
3. Relocate 'Personal' refresh button before label so doesn't disappear on narrow viewports, helps keep label alignment central.
4. Relocate 'Add something you like...' entry above list to reduce confusion as it was close to the dislikes list before.
5. Relocate 'Add something you dislike...' entry above list to suit new layout.
6. Rename caption for left panel to "Edit My Interests" to make the purpose of the Interests tab more obvious.
7. Reduce width of 'My Interests' and 'Similar Users' boxes by -40px so Recommendations box is still usable on narrow viewport.

### DIFF
--- a/pynicotine/gtkgui/ui/interests.ui
+++ b/pynicotine/gtkgui/ui/interests.ui
@@ -10,18 +10,17 @@
         <property name="hexpand">1</property>
         <child>
           <object class="GtkBox" id="LikesDislikes">
-            <property name="width-request">240</property>
+            <property name="width-request">200</property>
             <property name="visible">1</property>
             <property name="orientation">vertical</property>
             <child>
               <object class="GtkLabel">
                 <property name="visible">1</property>
-                <property name="xalign">0</property>
                 <property name="margin-start">12</property>
                 <property name="margin-end">12</property>
                 <property name="margin-top">12</property>
                 <property name="margin-bottom">12</property>
-                <property name="label" translatable="yes">My Interests</property>
+                <property name="label" translatable="yes">Edit My Interests</property>
                 <attributes>
                   <attribute name="weight" value="bold"/>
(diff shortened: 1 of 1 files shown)


slook:
Of course, the refresh button isn't needed anyway, as loading from the server is almost instant...

@mathiascode has a much better idea for a responsive layout see issue 1523, so this pull request isn't needed anymore. s.

PR #1541, opened 2021-09-04
Add Listenbrainz /now command integration

Nachtalb:
Add `/now` integration for [Listenbrainz](http://listenbrainz.org/). 

Sadly I cannot test it myself as I cannot run N+ from source atm as outlined in [this comment](https://github.com/Nachtalb/more-upload-stats/issues/1#issuecomment-912893610). **If someone else could take over from here that'd be mighty fine.**

The main integration of getting the song should basically work tho as I have tested it in a small plugin I wrote: https://github.com/Nachtalb/listenbrainz_now/blob/3bc9a23dc7454f04c2d14bf935b7da263fe502a6/listenbrainz_now/__init__.py#L29-L51

I'd still like to contribute as much as possible ^_____^  As soon as I am able to test the stuff myself I'll flood you with PRs or something...

During the po update, I get the following error. As such, I can't update the locals myself. There is a way to get around it, by adding an encoding comment at the top of the files. But then comes the next and the next and so on... I don't know where the problem lies (maybe I am missing some libs?)
```
running update_pot
xgettext: Non-ASCII string at pynicotine/geoip/geoip.py:36.
          Please specify the source encoding through --from-code or through a comment
          as specified in http://www.python.org/peps/pep-0263.html.
```

### DIFF
--- a/po/nicotine.pot
+++ b/po/nicotine.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2021-09-04 20:37+0300\n"
+"POT-Creation-Date: 2021-09-06 21:13+0200\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
@@ -1178,7 +1178,7 @@ msgstr ""
 msgid "You have been added to a private room: %(room)s"
 msgstr ""
 
-#: pynicotine/gtkgui/chatrooms.py:487 pynicotine/gtkgui/interests.py:104
+#: pynicotine/gtkgui/chatrooms.py:487 pynicotine/gtkgui/interests.py:105
 #: pynicotine/gtkgui/transferlist.py:142 pynicotine/gtkgui/userlist.py:78
 msgid "Status"
 msgstr ""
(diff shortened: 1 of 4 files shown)


pynicotine/nowplaying.py
```diff
@@ -259,6 +261,41 @@ def mpris(self, player):
...
+        """ Function to get the currently playing song listenbrainz api """
+
+        import json
+
+        if not username:
+            log.add_important_error(_('listenbrainz: Please provide your listenbrainz username'))
+            return
+
+        try:
+            response = http_request('https', 'api.listenbrainz.org', f'/1/user/{username}/playing-now',
```
slook, line 274: It doesn't seem like a good idea to have URL's hard-coded into the core plugin handler, especially if third-party plugin developers need to do this. We should consider a better way of hooking data into the required entries from the /plugins folder.

pynicotine/gtkgui/ui/settings/nowplaying.ui
```diff
@@ -49,6 +49,14 @@
                 <signal name="toggled" handler="update_now_playing_info" swapped="no"/>
               </object>
             </child>
+            <child>
+              <object class="GtkRadioButton" id="NP_listenbrainz">
```
slook, line 53: While this is a good suggestion for a data provider, many other providers may be desired in the future. We should provide a hook for /plugins/ so this can be facilitated without needing to modify the core.

pynicotine/gtkgui/settingswindow.py
```diff
@@ -2463,6 +2463,8 @@ def get_player(self):
             player = "lastfm"
         elif self.NP_mpris.get_active():
             player = "mpris"
+        elif self.NP_listenbrainz.get_active():
+            player = "listenbrainz"
```
slook, line 2467: While this is a good option for users, there may be many other options desired in the future. We should provide a /plugins/ option so that third-party developers don't need to modify the core.

slook (commented):
This is a good suggestion for new integration. However, I think it's time to consider if requiring third-party developers to make core changes is a viable long-term solution. 

My instinct tells me that we need to provide better hooks so that all of the proposed changes are made possible by placing a new file in the /plugins/ folder, but I do not know myself how this could be accomplished.

In conclusion, some further thought it needed to extend core functionality, in order to constrain third-party plugin development and updates to the /plugins/ folder only. We would welcome further suggestions @Nachtalb

slook:
Additionally, it has been bought to my attention:
_"it needs the f strings editing. if the minimum py version is 3.5, u cant use f strings, they didnt come 'til 3.6"_

mathiascode:
> During the po update, I get the following error. As such, I can't update the locals myself. There is a way to get around it, by adding an encoding comment at the top of the files. But then comes the next and the next and so on... I don't know where the problem lies (maybe I am missing some libs?)
> 
> ```
> running update_pot
> xgettext: Non-ASCII string at pynicotine/geoip/geoip.py:36.
>           Please specify the source encoding through --from-code or through a comment
>           as specified in http://www.python.org/peps/pep-0263.html.
> ```

In setup.py, does anything happen if you change:

`os.system("xgettext -o po/nicotine.pot " + " ".join(files))`

to

`os.system("xgettext --from-code=utf-8 -o po/nicotine.pot " + " ".join(files))`

I'm fine with these changes. Perhaps the whole now playing functionality would be better suited as a core plugin in the future, but that's long term.

pynicotine/nowplaying.py
mathiascode: Remove this f-string for Python 3.5 compatibility.

Nachtalb:
@mathiascode Yes that worked thanks :+1:

po/nicotine.pot
```diff
@@ -33,6 +33,10 @@ msgstr ""
 msgid "Graphical client for the Soulseek peer-to-peer network"
 msgstr ""
 
+#: data/org.nicotine_plus.Nicotine.desktop.in:8
+msgid "org.nicotine_plus.Nicotine"
```
mathiascode, line 37: The icon name shouldn't be here, but I'm not sure what caused this.

mathiascode: Probably this: https://savannah.gnu.org/support/?108887

Nachtalb: Thanks! Upgraded to 0.21 now and it's gone.

Nachtalb:
Btw is there a way to run these checks locally? When I use flake8 locally I don't get the same result :/

mathiascode (approved):

PR #1546, opened 2021-09-05
Catch threading exceptions in plugins

Nachtalb:
⚠️  ATM this only catches exceptions during command execution of a plugin. 

This should be looked at in a way of how to catch exceptions rather than actually merging it. I have tested in on 3.10 and 3.5 so it should work in the intermediate versions as well.  

I'd like some feedback on where you guys want the handling. One way would be to create to set a global thread exception hook in order to catch all any time or if you want it case wise. IMHO one to rule them all would make more sense. Because as of now when an error in a thread is raised N+ freezes up completely and has to be force killed.

### DIFF
--- a/pynicotine/pluginsystem.py
+++ b/pynicotine/pluginsystem.py
@@ -22,6 +22,7 @@
 
 import os
 import sys
+import threading
 
 from ast import literal_eval
 from time import time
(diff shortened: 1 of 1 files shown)


Nachtalb:
@mathiascode Could you give us your thought on this. Where it should be placed, project-wide hook etc.

mathiascode:
I haven't had time to look into this yet, but we have a global threading hook that shows the critical error dialog: https://github.com/nicotine-plus/nicotine-plus/blob/288b8b86715c7312ba7be1814e72abaac4ff9af7/pynicotine/gtkgui/frame.py#L93

Nachtalb:
Hmm, this does only wortk 3.8  and up tho. And errors in threads created by plugins crashes the whole program while other plugin errors do not.  :/

mathiascode:
I've added global plugin exception logging here: https://github.com/nicotine-plus/nicotine-plus/commit/45e5f12afe59bf5ffc9fc4f67cca6d47219ef49a

Let me know how it works for you. The threading excepthook could eventually be extended to cover Python 3.7 and lower.

slook:
I don't know if further effort on this is warranted, since the logged exceptions enable plugin developers to ensure that plugins are compatible with older systems that cannot catch thread exceptions.

Any user who is affected by a buggy plugin can workaround the issue by disabling the plugin, and reporting the issue to the plugin developer (which they are more likely to do if the application crashes completely).

mathiascode:
Implemented in b18220d9e1b7be5e3dea2375d1e0749d13c21b2f

PR #1549, opened 2021-09-06
Fix settings for plugins with uppercase letters

Nachtalb:
> As we can see above, the API is pretty straightforward. The only bit of magic involves the DEFAULT section which provides default values for all other sections. **Note also that keys in sections are case-insensitive and stored in lowercase**.
> ~ https://docs.python.org/3/library/configparser.html#quick-start

The default python config parser saves keys case insensitive - meaning plugins with a folder name like `MyPlugin` is saved as `myplugin`. When we read the config tho we used the key `MyPlugin` and thus thought the no plugin settings exist and overwrote them.


mathiascode (approved):
Good catch!

PR #1550, opened 2021-09-06
Properly clean up modules in case plugin fails to load

Nachtalb:
Up until now if the plugin failed to load the imported modules were still loaded. This may make it impossible to re enable the plugin after the plugin code has been fixed / changed.

### DIFF
--- a/pynicotine/pluginsystem.py
+++ b/pynicotine/pluginsystem.py
@@ -165,6 +165,7 @@ class PluginHandler:
 
         except Exception:
             from traceback import format_exc
+            self.unload_plugin(plugin_name)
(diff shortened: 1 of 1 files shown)


pynicotine/pluginsystem.py
```diff
@@ -200,20 +219,8 @@ def disable_plugin(self, plugin_name):
                 self.core.privatechats.CMDS.remove('/' + trigger + ' ')
 
             self.update_completions(plugin)
+            self.unload_plugin(plugin_name)
 
-            # Remove references to relative modules
```
mathiascode, line 204: Perhaps this code could be moved into a "finally" clause instead?

PR #1551, opened 2021-09-07
Prevent crashing with malformed or missing PLUGININFO data

Nachtalb:
ATM with malformed PLGUININFO data the plugin will be loaded upon start. But if the name was missing it crashed upon opening the plugin list and if other data was missing it crashed upon clicking on the plugin.

### DIFF
--- a/pynicotine/gtkgui/settingswindow.py
+++ b/pynicotine/gtkgui/settingswindow.py
@@ -2972,7 +2972,7 @@ class PluginsFrame(BuildFrame):
             return
 
         plugin_info = self.frame.np.pluginhandler.get_plugin_info(self.selected_plugin)
-        dialog = self.PluginPreferencesDialog(self, plugin_info["Name"])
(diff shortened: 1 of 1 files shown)


mathiascode:
I added some defaults in https://github.com/nicotine-plus/nicotine-plus/commit/7d012691cbb97584b3e1ee7cb57842a3d0ada35c, but you could probably replace "Example Plugin" with the technical plugin_name. This old code needs more work, since it lacks any kind of exception handling (file read error, invalid values etc).

Nachtalb:
I thought about providing a default dict. But I think it's a bit counterintuitive because it doesn't represent the real data. It would surely decrease the code lines touched. But like this, we wouldn't actually know if the data is from the plugin or from ourselves. 😕 I am not really sure...

mathiascode:
We can go with this approach, if you remove the default values from pluginsystem.py at the same time.

PR #1558, opened 2021-09-08
Only match clickable URLs in youtube_info

Inso-m-niaC:


Inso-m-niaC:
closing. redoing the regex.

PR #1578, opened 2021-09-22
Improve logging for pluginsystem

slook:
Remove some debug items from standard log, and change to Load / Unload terminology, because Enabled / Disabled was ambiguous (especially at shutdown).

### DIFF
--- a/pynicotine/pluginsystem.py
+++ b/pynicotine/pluginsystem.py
@@ -329,7 +329,7 @@ class PluginHandler:
                 continue
 
             log.add_debug(_("Plugin %(module)s returned something weird, '%(value)s', ignoring"),
-                    {'module': module, 'value': str(return_value)})
+                          {'module': module, 'value': str(return_value)})
 
         return False
 
@@ -369,7 +369,7 @@ class PluginHandler:
                 continue
 
             log.add_debug(_("Plugin %(module)s returned something weird, '%(value)s', ignoring"),
-                    {'module': module, 'value': return_value})
+                          {'module': module, 'value': return_value})
 
         return args
 
--- a/pynicotine/pluginsystem.py
+++ b/pynicotine/pluginsystem.py
@@ -286,7 +286,7 @@ class PluginHandler:
                     plugin.settings[key] = customsettings[key]
 
                 else:
-                    log.add_debug(_("Stored setting '%(key)s' is no longer present in the '%(name)s' plugin"), {'key': key, 'name': plugin.__name__})
+                    log.add_debug(_("Stored setting '%(key)s' is no longer present in the '%(name)s' plugin"), {'key': key, 'name': plugin_name})
 
         except KeyError:
             log.add_debug("No stored settings found for %s", (plugin.__name__,))
--- a/pynicotine/pluginsystem.py
+++ b/pynicotine/pluginsystem.py
@@ -44,7 +44,7 @@ class PluginHandler:
         self.core = core
         self.config = config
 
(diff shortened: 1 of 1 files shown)


pynicotine/pluginsystem.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, core, config):
         self.core = core
         self.config = config
 
-        log.add_debug("Loading plugin handler")
+        log.add("Loading plugin system")
```
slook, line 47: Ideally, this log line would only appear if Preferences > 'Enable logging' is True

pynicotine/pluginsystem.py
```diff
@@ -261,7 +261,7 @@ def load_enabled(self):
             return
 
         to_enable = self.config.sections["plugins"]["enabled"]
-        log.add_debug("Loading plugin(s): %s" % ', '.join(to_enable))
```
mathiascode, line 264: I understand what you're trying to do, but I interpreted this as loaded & enabled for a split second.

pynicotine/pluginsystem.py
mathiascode: You can probably move it after line 262.

pynicotine/pluginsystem.py
slook: _"Loaded" appears after each individual plugin init, so "Loading" status is not relevant in the debug log.
This is outcome of my changes:_
```
00:48:36 Loading plugin system
00:48:36 [Misc] Enabled plugin(s): leech_detector, multipaste, port_checker, testreplier, youtube_noembed
00:48:36 [Misc] Stored setting 'num_downloads' is no longer present in the 'Leech Detector' plugin
00:48:36 Leech Detector: Complain to leechers after tolerating 20 downloads, require users have minimum 10 files in 1 shared public folders.
00:48:36 Loaded plugin Leech Detector
00:48:36 Loaded plugin Multi Paste
00:48:36 Loaded plugin Port Checker
00:48:36 Loaded plugin Test Replier
00:48:36 Loaded plugin YouTube Noembed
```

pynicotine/pluginsystem.py
slook: Thanks, good idea. Actually I moved it just above as this seems to make sense. See comment below.

pynicotine/pluginsystem.py
```diff
@@ -286,10 +286,10 @@ def plugin_settings(self, plugin_name, plugin):
                     plugin.settings[key] = customsettings[key]
 
                 else:
-                    log.add(_("Stored setting '%(name)s' is no longer present in the plugin"), {'name': key})
+                    log.add_debug(_("Stored setting '%(key)s' is no longer present in the '%(name)s' plugin"), {'key': key, 'name': plugin.__name__})
```
Inso-m-niaC, line 289: This line is 149 characters long.

I'll be honest, i would prefer it if keys were deleted here instead. The only thing that would trigger this is if a developer modified the plugin and removed some plugin setting(s).

slook: Okay thanks, I will split the lines.

That is a good point, hopefully developers wont do this as it would make a mess of the user's config. However In the event of a key deletion, I think this log entry would even more desirable to keep it here (with slightly different wording).

Inso-m-niaC: I don't follow your logic here i'm afraid.

mathiascode:
Still a few lint errors to correct.

mathiascode (approved):

Issue #1565, opened 2021-09-15
Leech Detector is bugging people who have folders shared already.


Tenome:
Nicotine+ version:  
Operating System/Distribution:
3.1.1  •  GTK 3.24.30

### Describe the bug
Leech detector PMs people who are already sharing a folder (not sure how common it is, only one person has complained so far). Maybe the plugin is impatient and doesn't wait long enough if the user has a large share folder that doesn't immediately load?

mathiascode:
The leech detector plugin currently checks the number of shared files and folders sent from the server (visible in room user lists and user info), but this is unreliable due to bugs in SoulseekQt clients where they don't send the correct info to the server. We'll probably have to read the values from the user browse responses instead.

slook:
> not sure how common it is

This is fairly common. I believe it is possible that this happens when the plugin user has specified one or both of the "Least number of shared folders / files" parameters to a non-default value (more than 1). It would be useful if the plugin included/appended those parameters within the default message, but it doesn't.

> We'll probably have to read the values from the user browse responses instead.

I'm not sure that downloading users would want to receive notifications that "x user is browsing your shared files" from this plugin, because that would making it difficult to determine/log if it was a human browsing or not.

mathiascode:
> I'm not sure that downloading users would want to receive notifications that "x user is browsing your shared files" from this plugin, because that would making it difficult to determine/log if it was a human browsing or not.

It's too late for that. SoulseekQt also has a leech detector feature that has been using this method for years now.

slook:
Perhaps it is acceptable only in the error condition where a value of 0 is returned, by implementing the method as a fallback procedure.

> the number of shared files and folders sent from the server ... is unreliable due to bugs in SoulseekQt clients

Perhaps the Leech Detector plugin could be made able to workaround the known server bug, by giving the requesting requesting client the benefit of the doubt (and only send the message after verifying the 0 value with User Browse fallback method).

Tenome:
> I'm not sure that downloading users would want to receive notifications that "x user is browsing your shared files" from this plugin, because that would making it difficult to determine/log if it was a human browsing or not.

Is that not what the N+ script does already? I get those notifications in the bottom activity log fairly often when I'm browsing. I was assuming that I was triggering their leech detector plugin as well (I've never gotten a leech PM myself by accident either).

mathiascode:
> Is that not what the N+ script does already?

SoulseekQt does this.

[...]

PR #1579, opened 2021-09-22
Leech Detector: Add maximum tolerance threshold

slook:
Add 'Maximum number of leeched downloads to tolerate before complaining' option (default=20) using upload_finished_notification hook

Add 'Notify leechers about the required minimums' option (default=True)

Add 'Only send message if server reports non-zero numbers that are definitely correct' (default=True) to workaround #1565 until User Browse count is implemented.

Prevent complaining to users in Buddy List if 'Queue limits do not apply to users in buddy list' option is set in Main Preferences > Upload


slook:
Sorry to confuse the pr, I'm still trying to figure out the best way of creating a branch patch.

PR #1581, opened 2021-09-24
Various improvements to plugin system

mathiascode:
- Remove the `__name__` BasePlugin class attribute. Automatically set `internal_name` (plugin folder name) and `human_name` (human-readable name) attributes when the plugin loads.
- Remove arguments from BasePlugin's `__init__` function, allow plugins to safely use this function if they want. Avoids Pylint's "attribute-defined-outside-init" warning.
- Add new notifications:
  - `loaded_notification`: plugin has finished loading (settings are loaded)
  - `unloaded_notification`: plugin has finished unloading
- Allow unified commands (for both public and private chats) to be defined in the `__commands__` attribute. `__publiccommands__` and `__privatecommands__` can still be used if unified commands aren't desired.

All changes in this PR are backwards compatible.

### DIFF
--- a/pynicotine/plugins/anti_shout/__init__.py
+++ b/pynicotine/plugins/anti_shout/__init__.py
@@ -22,19 +22,24 @@ from pynicotine.pluginsystem import BasePlugin
 
 class Plugin(BasePlugin):
 
-    __name__ = "Anti SHOUT"
-    settings = {
-        'maxscore': 0.6,
-        'minlength': 10,
-    }
-    metasettings = {
-        'maxscore': {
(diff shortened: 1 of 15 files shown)


Inso-m-niaC:
Nice, like it. Do you think 'public_command' might be better suited as the last argument? You know, similar to private and public argument placements, but with the addition argument appended.

mathiascode:
> Nice, like it. Do you think 'public_command' might be better suited as the last argument?

You're right, this would be a better idea.

Inso-m-niaC:
Excellent!

PR #1582, opened 2021-09-24
Fix log [Conn] Connection closed by peer (user)

slook:
- closed_connection [Conn] log message was garbled log output spanning multiple lines making the [Conn] log unreadable.
_With this commit it now prints:_
`17:04:46 [Conn] Connection of type P to user Username closed by peer ('12.34.56.78', 12345)`

- connect_to_peer_direct [Conn] log message changed wording because it's too soon to know if successful init will be possible or not. Also added IP address because it's inexpensive and useful here. _With this commit it now prints:_
`17:06:16 [Conn] Attempting direct connection of type P to user Username ('12.34.56.78', 12345)`

### DIFF
--- a/pynicotine/pynicotine.py
+++ b/pynicotine/pynicotine.py
@@ -589,8 +589,7 @@ class NicotineCore:
         self.out_indirect_conn_request_times[conn] = time.time()
 
         log.add_conn(
-            """Direct connection of type %(type)s to user %(user)s failed, attempting indirect connection.
-Error: %(error)s""", {
+            "Direct connection of type %(type)s to user %(user)s failed: %(error)s. Attempting indirect connection.", {
                 "type": conn.conn_type,
                 "user": conn.username,
                 "error": error
@@ -798,7 +797,7 @@ Error: %(error)s""", {
 
                 i.conn = conn
 
-                log.add_conn("Connection established with user %(user)s. List of outgoing messages: %(messages)s", {
+                log.add_conn("Established connection with user %(user)s. List of outgoing messages: %(messages)s", {
                     'user': i.username,
                     'messages': i.msgs
                 })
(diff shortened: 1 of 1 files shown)


slook:
I am unsure if error=None is needed on Line 986 for catching the correct events, or if it can be safely removed?
`    def closed_connection(self, conn, addr, error=None):`

The error is not useful to enter into the log, because it is often contextually incorrect (None) by this point.

mathiascode:
> I am unsure if error=None is needed on Line 986 for catching the correct events, or if it can be safely removed?
> ` def closed_connection(self, conn, addr, error=None):`
> 
> The error is not useful to enter into the log, because "None" is often contextually incorrect by this point.

You can remove it, but you also have to do it at line 1064. Perhaps add a new log string around this line, something like "connection attempt to peer x failed. Error:".

pynicotine/pynicotine.py
```diff
@@ -992,8 +993,8 @@ def closed_connection(self, conn, addr, error=None):
 
             for i in self.peerconns:
                 if i.conn == conn:
-                    log.add_conn("Connection closed by peer: %(peer)s. Error: %(error)s",
```
mathiascode, line 995: This reminds me, we should probably remove "by peer". The connection isn't necessary closed by the other user/peer, sometimes we have to do it too.

slook:
Not sure if we will have a username if we cannot connect. I am unable to trigger this error on line 1064, do you know what scenario would lead up to this?

mathiascode:
I forgot how this code worked. https://github.com/nicotine-plus/nicotine-plus/commit/36f484d2066c0a4c6886b9436dd6469d8705165f removes the confusing call at the end of connect_error, so you don't have to add a new log message.

slook:
Oh, I already did and committed. Now there's conflicts I don't know how to resolve.

mathiascode:
Could you merge the latest changes from master to your branch, so that previous changes such as line 1062-1064 don't appear in the diff?

slook:
Done. Sorry my branch is a mess, I don't know the cleanest way of going about things.

mathiascode:
No problem.

mathiascode (approved):

PR #1584, opened 2021-09-25
Status bar always shows text

slook:
... even when the Log Pane is visible (otherwise, this is just wasted space that could be better used).

### DIFF
--- a/pynicotine/gtkgui/frame.py
+++ b/pynicotine/gtkgui/frame.py
@@ -730,7 +730,6 @@ class NicotineFrame(UserInterface):
     def set_show_log(self, show):
 
         if show:
-            self.set_status_text("")
             self.DebugLog.show()
             self.log_textview.scroll_bottom()
         else:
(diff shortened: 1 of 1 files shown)


pynicotine/gtkgui/frame.py
```diff
@@ -2212,7 +2211,7 @@ def update_log(self, msg, level):
             message_dialog(parent=self.application.get_active_window(), title=title, message=msg)
             return
 
-        if config.sections["logging"]["logcollapsed"]:
+        if level not in ("connection", "message", "miscellaneous"):
```
mathiascode, line 2214: A comment about why specific levels are disabled would be nice to have.

slook:
In addition to the above edited comment, further discussion about ideas for using the status bar to help with essential client functions (mainly Browsing, Downloads and Chat) in https://github.com/nicotine-plus/nicotine-plus/discussions/1533#discussioncomment-1374648 and https://github.com/nicotine-plus/nicotine-plus/discussions/1526

pynicotine/gtkgui/frame.py
mathiascode: Sorry, should've been more specific. A short code comment above this line telling someone who browses the code why this edge case is here.

slook: Omg you want me to put comments in the code?! I hope users don't mind the installation being > 1GB :-D

mathiascode (approved):

PR #1587, opened 2021-09-28
Browse Shares: Folder & File TreeView Keyboard Accessibility Improvements

slook:
**_Navigation and Accessibility (Folder & File TreeViews):_**
- Fixed: Widget focus forgotten after switching tabs, previous file selection was lost when navigating with up and down keys.
- Changed: [`Enter`] Folder - Row_Activated action to Expand/Collapse tree node _(instead of Download Folder without prompt)_.
- Added: [`Enter`] Folders - Switch focus over to FileView if at the deepest tree level.
- Added: [`Shift+Enter`] Folders - Switch focus over to FileView regardless of tree level expansion.
- Added: [`BackSpace`] Files - Switch focus back to FolderView parent.
- Added: [`Left`] Files - Switch focus back over to FolderView if the FileScrolledWindow is fully to the left horizontal edge.
- Added: [`Right`] Folders - Switch focus over to FileView if already expanded the deepest tree level.
- Added: [`=equal`] Folder - Expand selected sub-node _(for UK/US laptops without numeric keypads which share the + key)_.
- Added: [`\backslash`] Folder - Expand sub-folder / Collapse down to parent tree level to view sub-folders.
- Added: [`Ctrl+\backslash`] Folders - Expand / Collapse All ... or switches focus over to Folders if focused on Files.

**_Key-modified event actions for selected item(s):_**
- Changed: [`Enter`] Folder - Row_Activated action to Expand/Collapse tree node _(instead of Download Folder without prompt)_.
- Added: [`Shift+Ctrl+Enter`] Download File(s) (no prompt) / Upload Folder Recursive... / **Download Folder without prompt**.
- Added: [`Shift+Enter`] File(s) - Download multi-selection (no prompt) / Send multiple Files to Player _(if enqueue supported)_.
- Added: [`Ctrl+Enter`] Download To... (with destination prompt) / Upload To User... (either Folder or File(s))
- Added: [`Ctrl+Alt+Enter`] Open in File Manager (for self-browing only, otherwise File Properties same as [Alt+Enter]).

**_Keyboard shortcuts for current Browse tab:_**
- Added: [`Ctrl+F`] Find text entry focus grab.
- Added: [`Ctrl+G`] and [`F3`] Find Next.
- Added: [`Shift+Ctrl+G`] and [`Shift+F3`] Find Previous _(key binding connected, but feature not implemented yet)_.
- Added: [`Ctrl+R` and [`F5`] Refresh.
- Added: [`Ctrl+S`] Save Shares List to Disk.

**_Dialog strings accessibility improvements:_**
- Changed: Title for "Select a Folder" to "Select Destination for Downloading **File(s)** from User" prompt for 'Download **File(s)**'.
- Changed: Title for "Select a Folder" to "Select Destination for Downloading Folder _(with **Subfolders)_** from User" _v_
- Changed: Title for "Upload Folder's Contents" to "Upload Folder _(with Subfolders)_ To User" _^(depend if Recursive or not)_
- Changed: Title for "Upload File(s)" to "Upload File(s) **To User**" to make it clearer what this function does.

_Note:_ After this merge, it will be possible to connect the above [Enter] functions to mouse Double-Click events, as desired...

_Note:_ The above [Enter] functions are currently bound only to the "Return" keysym, but the word 'Enter' is used to indicate any sort of execution trigger is possible in the future, because each ~_accelerator definition may be called by any procedure.


pynicotine/gtkgui/userbrowse.py
```diff
@@ -687,6 +694,13 @@ def on_select_dir(self, selection):
 
         self.set_directory(iterator.user_data)
 
+    def on_select_file(self, selection):
+
+        if selection is None:
+            return
+
+        self.select_files()
```
mathiascode, line 702: I think I would be more comfortable having something like `self.num_selected_files = selection.count_selected_rows()` here instead of requesting data for every row on each selection.

slook: You're right that was inefficient as this is not needed for every case where this is called. Thank you for your suggestion.

pynicotine/gtkgui/userbrowse.py
```diff
@@ -688,6 +695,13 @@ def on_select_dir(self, selection):
 
         self.set_directory(iterator.user_data)
 
+    def on_select_file(self, selection):
+
+        if selection is None:
+            return
+
+        self.num_selected_files = selection.count_selected_rows()
```
mathiascode, line 703: Add `self.num_selected_files = 0` in __init__() too, below `self.selected_files = {}`

pynicotine/gtkgui/userbrowse.py
```diff
@@ -400,22 +408,19 @@ def on_file_row_activated(self, treeview, path, column):
 
     def on_file_popup_menu(self, menu, widget):
 
-        self.select_files()
```
mathiascode, line 403: This line is still needed, otherwise there are no selected files for the context menu items.

slook: Thank you, this has greatly improved my understanding it's purpose, useful for implementation of new key bindings with multi-select actions that are now possible thanks to these experimentations and your help, which is much appreciated.

pynicotine/gtkgui/userbrowse.py
```diff
@@ -688,34 +712,277 @@ def on_select_dir(self, selection):
...
+                # Shift focus across to files
+                self.FileTreeView.grab_focus()
+                return True
+
+        # Expand folder node with arrow cursor
         keycodes, mods = parse_accelerator("Right")
 
-        if keycode in keycodes:
-            path, _focus_column = self.FolderTreeView.get_cursor()
+        if keycode in keycodes or (int(keyval) == 61):  # Laptop unshifted + "=" key
```
slook, line 741: The "+" key on most laptops (that don't have numeric keypads) shares a key with "=".

This keyval solution is rather ugly, but I am unaware of a more graceful way. I haven't tested this for international compatibility yet, but I have a feeling this might cause trouble.

pynicotine/gtkgui/userbrowse.py
```diff
@@ -688,34 +712,277 @@ def on_select_dir(self, selection):
...
             return True
 
+        # Collapse folder node with arrow cursor
         keycodes, mods = parse_accelerator("Left")
 
         if keycode in keycodes:
-            path, _focus_column = self.FolderTreeView.get_cursor()
-            self.FolderTreeView.collapse_row(path)
+            if path is not None:  # Cannot collapse a root that is already collapsed
+                self.FolderTreeView.collapse_row(path)
```
slook, line 750: The variable assignment is moved to the top of the definition, because 'path' is needed for many of the other functions.

The code that follows is long-winded and scruffy, I intend to find more elegant solutions for testing the conditions, ideally a way of parsing the accelerator on one line, and testing for overlapping scenarios more dynamically.

pynicotine/gtkgui/userbrowse.py
```diff
@@ -688,34 +712,277 @@ def on_select_dir(self, selection):
...
         keycodes, mods = parse_accelerator("<Alt>Return")
 
         if state & mods and keycode in keycodes:
             self.on_file_properties()
             return True
 
+        self.on_browse_key_press_event(*args)
+        return False
+
+    def on_browse_key_press_event(self, *args):
```
slook, line 925: This has been added here with the intention of reducing duplication, perhaps some of the above conditions can be re-worked to something more palatable. As these keys need a wider scope, perhaps there would be a better place for this?

pynicotine/gtkgui/userbrowse.py
```diff
@@ -856,8 +1129,12 @@ def on_download_files_to(self, *args):
         if not os.path.exists(path) or not os.path.isdir(path):
             path = download_folder
 
+        # ToDo: Make translatable
+        str_title="Select Destination for Downloading File(s) from User"
```
slook: There is some KeyError with this which I cannot seem to resolve, it seems choose_dir doesn't like the _() syntax.

pynicotine/gtkgui/userbrowse.py
```diff
@@ -77,7 +77,13 @@ def on_switch_browse_page(self, notebook, page, page_num):
 
         for tab in self.pages.values():
             if tab.Main == page:
-                GLib.idle_add(lambda: tab.FolderTreeView.grab_focus() == -1)
+
+                # Remember folder or file selection
```
slook, line 81: Fixes a bug where the focus was forgotten when switching tabs, and any previously file selection was lost when navigating with the keyboard (up and down cursor keys).

pynicotine/gtkgui/userbrowse.py
mathiascode: Keyvals have caused us trouble on international keyboards previously, so they're indeed a rather ugly solution.

We do something like this in a few other locations: https://github.com/nicotine-plus/nicotine-plus/blob/20bb2450353f63d1cafd0e221388202f62b35905/pynicotine/gtkgui/frame.py#L1578-L1579

pynicotine/gtkgui/userbrowse.py
mathiascode: > I intend to find more elegant solutions for testing the conditions

This is fine, but please don't include any additional large changes in this PR.

pynicotine/gtkgui/userbrowse.py
mathiascode: Would self.Main be appropriate?

pynicotine/gtkgui/userbrowse.py
mathiascode: Variable name conflict on line 1119. Rename `_` to something like `_path_start`.

pynicotine/gtkgui/userbrowse.py
slook: Sure, this is already a larger fix than I had intended. Please feel free to make or suggest changes at any point in the future as you see fit.

pynicotine/gtkgui/userbrowse.py
```diff
@@ -191,10 +197,11 @@ def __init__(self, userbrowses, user):
         # Iters for current FileStore
         self.files = {}
         self.totalsize = 0
+        self.num_selected_files = 0
 
         self.dir_store = Gtk.TreeStore(str)
         self.FolderTreeView.set_model(self.dir_store)
-        self.folder_key_controller = connect_key_press_event(self.FolderTreeView, self.on_folder_key_press_event)
+        self.folder_key_controller = connect_key_press_event(self.FolderTreeView, self.on_key_press_event_folder)
```
slook, line 204: This follows the naming scheme in other modules, if you need me to change it back let me know.

pynicotine/gtkgui/userbrowse.py
```diff
@@ -688,34 +711,338 @@ def on_select_dir(self, selection):
...
+            if path is not None and (prs_none):
+                self.FolderTreeView.collapse_row(path)
+                return True
+
+        # Right: Expand folder
+        keycodes, mods = parse_accelerator("Right")
+
+        if (keycode in keycodes and not prs_ctrl and not prs_alt) or (
+            keycode == 21 and prs_none) or (  # EN-US/GB laptop non-shifted +"=" key
+            keycode == 94 and prs_none        # EN-GB "\" key (KeyVal here instead?)
```
slook, line 759: The "=" has been changed to a keycode.
The "\" might be better off as a KeyVal, because it appears in different places on international keyboards, I'm not sure.

pynicotine/gtkgui/userbrowse.py
slook: > Would self.Main be appropriate?

I don't think so no, because all the shortcuts are in reference to the particular user tab page.

> Add another key controller for a container widget

There is no container that has an id assignment, I tried giving a GtkBox an id, but this didn't work. Whilst it would be preferable to have shortcuts working whilst using Find, it's not a severe keyboard navigation trap as it's possible to escape out of the SearchEntry box with the down arrow key into the FolderTreeView. So I'm happy to leave this for the time being.

pynicotine/gtkgui/userbrowse.py
```diff
@@ -358,8 +366,25 @@ def on_expand(self, widget):
             self.expand.set_property("icon-name", "go-down-symbolic")
 
     def on_folder_row_activated(self, treeview, path, column):
-        if self.user != config.sections["server"]["login"]:
-            self.on_download_directory()
```
slook, line 362: It is way too easy to download a folder, as there is no confirmation, this leads to accidental downloads (evidenced by transfers that get immediately cancelled) so I have changed the folder row activation to Expand / Collapse instead.

In the future I intend to restore the ability to download a folder with mouse-click, by using a <Control> modifier keymask, which would use the same conditions as the "Return" key binding, so that code may be moved into here at some point, when I can figure out how to pass on_key_press_event_folder(*args) into here.

pynicotine/gtkgui/userbrowse.py
```diff
@@ -688,34 +711,336 @@ def on_select_dir(self, selection):
...
+                else:
+                    self.SearchEntry.grab_focus()
+
+                return True
+
+        # Ctrl+R or F5: Refresh
+        keycodes_r, _mods_r = parse_accelerator("<Primary>r")
+        keycodes_f5, _mods_f5 = parse_accelerator("F5")
+
+        if (keycode in keycodes_r and prs_ctrl and not prs_shift) or (keycode in keycodes_f5 and prs_none):
```
slook, line 1001: The 'mods' returned by Gtk.accelerator_parse is woefully inadequate for our purposes, because it validates even when additional modifier keys are also pressed (eg: Ctrl+R is not the same as Shift+Ctrl+R). This is where key conflicts arise.

For now, I have done this hack that is currently needed to get the shortcuts working properly, but I feel there must be a cleaner way of doing this. Perhaps our parse_accelerator() function can be expanded to better handle multiple keys, or maybe it would be safer to create a new function such as parse_shortcut() that is more suited to our needs.

pynicotine/gtkgui/userbrowse.py
```diff
@@ -688,34 +711,336 @@ def on_select_dir(self, selection):
...
+            prs_ctrl = True
+
+        if state & mods_shift:  # <GDK_SHIFT_MASK>
+            prs_shift = True
+
+        if state & mods_alt:  # <GDK_MOD1_MASK>
+            prs_alt = True
+
+        if prs_ctrl == prs_shift == prs_alt is False:
+            prs_none = True  # <GDK_MOD2_MASK> = plain
```
slook, line 742: These flags are needed to exclude un-pressed modifier keys from the bindings, since parse_accelerator doesn't seem to be capable of avoiding conflicts by itself. Since this needs to happen for every on_key_press_event function, I would like to do this externally like in utils.py, but I don't yet know how to best accomplish this.

pynicotine/gtkgui/userbrowse.py
```diff
@@ -358,8 +366,25 @@ def on_expand(self, widget):
...
+        demanded_state = original_state
+
+        if original_state:
+            self.FolderTreeView.expand_row(path, False)
+        else:
+            demanded_state = self.FolderTreeView.collapse_row(path)
+
+        if (original_state == demanded_state is False) and int(len(self.file_store)) >= 1:
+            # This is the deepest level, so move focus over to Files if there are any
+            self.FileTreeView.grab_focus()
```
mathiascode, line 384:
```suggestion
        if self.FolderTreeView.row_expanded(path):
            expandable = self.FolderTreeView.collapse_row(path)
        else:
            expandable = self.FolderTreeView.expand_row(path, False)

        if not expandable and len(self.file_store) > 0:
            # This is the deepest level, so move focus over to Files if there are any
            self.FileTreeView.grab_focus()
```

pynicotine/gtkgui/userbrowse.py
```diff
@@ -688,34 +711,336 @@ def on_select_dir(self, selection):
...
+                elif int(len(self.file_store)) <= 0 or (prs_shift):
+                    self.on_upload_directory_recursive_to()
+
+                return True
+
+            else:  # [user is not self]
+                return False
+
+        # Next key binding chain is self.Main
         return False
```
mathiascode, line 1044: This should accomplish what you want, if I understood your code correctly:
```
    def on_key_press_event_browse(self, *args):
        """ Key Bindings (this UserBrowses Page) """

        keyval, keycode, state, widget = get_key_press_event_args(*args)

        # Ctrl+F: Focus Find TextEntry
        keycodes, mods = parse_accelerator("<Primary>f")

        if state == mods and keycode in keycodes:
            self.SearchEntry.grab_focus()
            return True

        # Ctrl+G or F3: Find Next Match
        keycodes_g, mods_g = parse_accelerator("<Primary>g")
        keycodes_f3, mods_f3 = parse_accelerator("F3")

        if (state == mods_g and keycode in keycodes_g) or (state == mods_f3 and keycode in keycodes_f3):
            # Shift+Ctrl+G or Shift+F3: Find Previous Match
            # ToDo: {self.search_position = self.search_position - 2}
            self.on_search()
            return True

        # Shift+Ctrl+G or Shift+F3: Find Previous Match
        keycodes_g, mods_g = parse_accelerator("<Primary><Shift>g")
        keycodes_f3, mods_f3 = parse_accelerator("<Shift>F3")

        if (state == mods_g and keycode in keycodes_g) or (state == mods_f3 and keycode in keycodes_f3):
            # ToDo: {self.search_position = self.search_position - 2}
            self.on_search()
            return True

        # Ctrl+R or F5: Refresh
        keycodes_r, mods_r = parse_accelerator("<Primary>r")
        keycodes_f5, mods_f5 = parse_accelerator("F5")

        if (state == mods_r and keycode in keycodes_r) or (state == mods_f5 and keycode in keycodes_f5):
            self.on_refresh()
            return True

        # Ctrl+S: Save List
        keycodes, mods = parse_accelerator("<Primary>s")

        if state == mods and keycode in keycodes:
            self.on_save()
            return True

        # Shift+Ctrl+S: Idea - Save List of all open tabs
        keycodes, mods = parse_accelerator("<Primary><Shift>s")

        if state == mods and keycode in keycodes:
            return True

        # Ctrl+U: Upload To User...
        keycodes, mods = parse_accelerator("<Primary>u")

        if state == mods and keycode in keycodes:
            if self.user != config.sections["server"]["login"]:
                return False

            self.select_files()

            if self.num_selected_files >= 1:
                self.on_upload_files()

            elif len(self.file_store) >= 1 and self.num_selected_files <= 0:
                self.on_upload_directory_to()

            elif len(self.file_store) <= 0:
                self.on_upload_directory_recursive_to()

            return True

        # Shift+Ctrl+U: Upload Folder To User...
        keycodes, mods = parse_accelerator("<Primary><Shift>u")

        if state == mods and keycode in keycodes:
            if self.user != config.sections["server"]["login"]:
                return False

            self.select_files()

            if self.num_selected_files >= 1:
                self.on_upload_directory_to()
            else:
                self.on_upload_directory_recursive_to()

            return True

        # Next key binding chain is self.Main
        return False
```

pynicotine/gtkgui/userbrowse.py
mathiascode: See https://github.com/nicotine-plus/nicotine-plus/pull/1587/files#r720785164

pynicotine/gtkgui/userbrowse.py
mathiascode: `if state == mods and keycode in keycodes` could probably be abstracted into a utility function? I could do this.

pynicotine/gtkgui/userbrowse.py
mathiascode: Each tab has a separate self.Main container widget (see userbrowse.ui).

pynicotine/gtkgui/userbrowse.py
mathiascode: It would look like this:
```
        # Ctrl+F: Focus Find TextEntry
        if accelerator_activated("<Primary>f", keycode, state):
            self.SearchEntry.grab_focus()
            return True
```

slook: > This should accomplish what you want

Your suggested code doesn't work, I think the bitwise & operator must be needed because the MOD2 (_no modifier keys pressed_) is always present, and this makes the raw variables different even when **_CONTROL_** is pressed and nothing else...

_str(state):_
`<flags GDK_CONTROL_MASK | GDK_MOD2_MASK of type Gdk.ModifierType>`

_str(mods):_
`<flags GDK_CONTROL_MASK of type Gdk.ModifierType>`

So unfortunately, `state` != `mods` (not equal: at least not on the two ThinkPad laptop systems I've tested it on).

> a utility function? I could do this.

This would be highly beneficial in the long run, as faffing about with these state and mod variables for every shortcut is not efficient, ideally any key combination could be tested on one line, and placed within any procedure definition, even if the key_event *args are not available (such as for Ctrl+MouseDblClick usage in on_folder_row_activated for example).

It is important that any extra pressed keys would cause your proposed utility function to return `False`, and only return `True` if the exact key combination is present with nothing more. In a perfect world, I imagine something like:

```
if get_key_press("<Primary>f"):
    self.SearchEntry.grab_focus()
```

pynicotine/gtkgui/userbrowse.py
slook: I have tried adding a key_controller for selfMain into init, unfortunately it causes the Upload To User dialog escape function to be broken (other than that it works well).

`        self.key_controller = connect_key_press_event(self.Main, self.on_key_press_event_browse)`

pynicotine/gtkgui/userbrowse.py
slook: Your commit https://github.com/nicotine-plus/nicotine-plus/commit/a16bd4dc25d809241ac9ce3b63ab2abd518b6e1e will help greatly with these endeavors. The key bindings are going to be much more modular now, which is far better than giant on_key_press_event functions.

pynicotine/gtkgui/userbrowse.py
slook: The solution that https://github.com/nicotine-plus/nicotine-plus/commit/a16bd4dc25d809241ac9ce3b63ab2abd518b6e1e offers makes this irrelevant now.

pynicotine/gtkgui/userbrowse.py
slook: I found the answer to this (21="equal"; 94="backslash"; 22="BackSpace"), link:
https://anzeljg.github.io/rin2/book2/2405/docs/tkinter/key-names.html
https://www.tcl.tk/man/tcl8.4/TkCmd/keysyms.html

pynicotine/gtkgui/userbrowse.py
```diff
@@ -336,6 +374,44 @@ def __init__(self, userbrowses, user):
...
+        """ Shortcut Key Bindings (Tool Bar) """
+
+        setup_accelerator("<Primary>f", self.Main, self.on_search_accelerator)  # Find focus
+        setup_accelerator("<Primary>f", self.FolderTreeView, self.on_search_accelerator)
+        setup_accelerator("<Primary>f", self.FileTreeView, self.on_search_accelerator)
+
+        setup_accelerator("<Primary>g", self.Main, self.on_search_next_accelerator)  # Find on_search Enter (repeat)
+        setup_accelerator("<Primary>g", self.FolderTreeView, self.on_search_next_accelerator)
+        setup_accelerator("<Primary>g", self.FileTreeView, self.on_search_next_accelerator)
+        setup_accelerator("F3", self.Main, self.on_search_next_accelerator)
```
slook, line 386: These function declarations each require entry into 3 objects, which would make a For loop more appropriate (to iterate through a list of: Main; FolderTreeView; FileTreeView) to save on code lines, but I don't know how to implement this.

pynicotine/gtkgui/userbrowse.py
```diff
@@ -357,8 +439,22 @@ def on_expand(self, widget):
...
+            expandable = self.FolderTreeView.collapse_row(path)
+        else:
+            expandable = self.FolderTreeView.expand_row(path, False)
+
+        if not expandable and len(self.file_store) > 0:
+            # This is the deepest level, so move focus over to Files if there are any
+            self.FileTreeView.grab_focus()
+
+        # Note: Other Folder actions are handled by setup_accelerator functions [Shift/Ctrl/Alt+Return]
+        # ToDo: Mouse double-click actions will need *args for keycode state & mods [Ctrl/Alt+DblClick]         # ToDo #
```
slook, line 458: Now that the accelerators each have their own function definition, this will allow selective Mouse-click events to be assigned to them, this is why I have specified the word "Enter" (rather than "Return") in the code comments below.

_See link: https://anzeljg.github.io/rin2/book2/2405/docs/tkinter/key-names.html_
`Return | 36 | 65293 | The enter key... The name Enter refers to a mouse-related event, not a keypress;`

pynicotine/gtkgui/userbrowse.py
```diff
@@ -687,24 +789,233 @@ def on_select_dir(self, selection):
...
+            self.on_download_directory_to()
+
+        return True
+
+    """ Key Bindings (FileTreeView) """
+
+    def on_file_focus_folder_left_accelerator(self, *args):
+        """ Left: focus back parent folder (left arrow) """
+
+        # ToDo: Only if... FileTreeView.selected_column == 0:  # ToDo: (scrolled fully left)                    # ToDo #
```
slook, line 911: This functionality is scuppered by an issue with horizontal scrolling. A mouse-less user would be able to scroll right, but not left, meaning that the FileTreeView can get stuck in an unreadable position.

I need to detect which column is selected (or the value of the h-scrollbar), then only allow the focus shift if it's column[0] (left-most), but I'm unaware of how to implement the GtkTreeView methods. I want to do something like:

```
        _path, focus_column = self.FileTreeView.get_cursor()
        first_column = self.FileTreeView.get_column(0)

        if focus_column == first_column:
            self.FolderTreeView.grab_focus()
```
Do you know of a solution for this? Alternatively, now that the userbrowse.ui is reponsive, I see little benefit to horizontal scrolling. Perhaps it would be better to auto-expand the Filename column, but I don't know how to accomplish this either.

pynicotine/gtkgui/userbrowse.py
```diff
@@ -687,24 +789,233 @@ def on_select_dir(self, selection):
...
+
+    def on_folder_focus_filetree_accelerator(self, *args):
+        """ Shift+Enter: focus selection over FileTree  """
+
+        if len(self.file_store) >= 1:
+            self.FileTreeView.grab_focus()
+
+        else:
+            self.on_folder_expand_sub_accelerator()
+            # ToDo: Expand all branch sub-nodes (*)                                                             # ToDo #
```
slook, line 851: This would be nice but isn't essential. If it can't be easily done then I will remove this comment line.

pynicotine/gtkgui/userbrowse.py
mathiascode: Isn't it enough to add accelerators for Main? They should work in any child widgets, unless there's a bug.

slook: Correct, perhaps I was brushing with broad strokes here. Your new implementation seems much more robust so these forceful overrides are not needed so much anymore.

This same token like goes for the on_folder~ and on_file~ functions aswell, in future my plan is to condense these down by structuring those functions which are shared between the two Treeview widgets, but my ability for this level of logic doesn't come easily.

pynicotine/gtkgui/ui/userbrowse.ui
```diff
@@ -268,7 +267,7 @@
                       <object class="GtkBox">
                         <property name="visible">1</property>
                         <child>
-                          <object class="GtkScrolledWindow">
+                          <object class="GtkScrolledWindow" id="FileScrolledWindow">
```
slook, line 270: This id tag is needed for the `FileScrolledWindow.get_hadjustment()` function in userbrowse.py which enables the use of the "Left" arrow key for both widget focusing and horizontal scrolling.

pynicotine/gtkgui/userbrowse.py
slook: I didn't think I was going to be able to find a solution for this, but I found one in the end:

```
    def on_file_focus_folder_left_accelerator(self, *args):
        """ Left: focus back parent folder (left arrow) """

        h = self.FileScrolledWindow.get_hadjustment()

        if h.get_value() > 0.0:
            # Allow horizontal scrolling
            return False

        self.FolderTreeView.grab_focus()

        return True
```

pynicotine/gtkgui/userbrowse.py
slook: This conversation comment has become misplaced. This was in reference to on_select_file which has been implemented with `selection.count_selected_rows()` on line 775 as per your suggestion.

pynicotine/gtkgui/userbrowse.py
```diff
@@ -361,8 +421,22 @@ def on_expand(self, widget):
...
+            expandable = self.FolderTreeView.collapse_row(path)
+        else:
+            expandable = self.FolderTreeView.expand_row(path, False)
+
+        if not expandable and len(self.file_store) > 0:
+            # This is the deepest level, so move focus over to Files if there are any
+            self.FileTreeView.grab_focus()
+
+        # Note: Other Folder actions are handled by setup_accelerator functions [Shift/Ctrl/Alt+Return]
+        # ToDo: Mouse double-click actions will need *args for keycode state & mods [Ctrl/Alt+DblClick]
```
mathiascode, line 439: What does this mean?

mathiascode:
There are certain "combo" shortcuts I don't quite like. In my opinion, a shortcut should do one thing only, otherwise things quickly become messy and difficult to remember. E.g. Alt+Return currently has three actions depending on what's focused.

pynicotine/gtkgui/userbrowse.py
slook: My intention is to trigger some useful accelerator functions from modified mouse click events, but I can't implement this here because the state of the `Ctrl`, `Alt` and `Shift` keys are not easily available as arguments.

For example `Ctrl+DblClick` = (the same as Ctrl+Return) -> `on_folder_transfer_to_accelerator()`

Perhaps a new or existing function in utils.py could serve this purpose?

mathiascode: Right, a whole different mess to deal with. Might have to look into this and "row-activated" when rewriting treeviews in a future Nicotine+ version.

slook:
I appreciate your feedback as this was something I had pondered upon myself.

> Alt+Return currently has three completely different actions depending on what's focused.

I agree with you that the `Alt` bindings do feel overly complex. This is caused by the fact that we don't have a 'Folder Properties' dialog (Open in File Manager is the closest action we have for that, but this is not possible for a remote folder, which is why I have chosen to fall back to the same action as if Alt was not pressed).

To simplify this would mean that nothing happens when a folder is focused, admittedly this would be better for the purposes of the keyboard shortcut, but would cause a UI block if and when the function is triggered with a mouse event, (because the non-Alt'ed function would be expected).

> things quickly become messy and difficult to remember

The Return key (Enter and Double-Click) bindings are not intended to be remembered, my goal is for the UI operation to feel natural and work in a standard fashion.

A possible solution to make the `Alt` functions more consistent would be to implement @redactedscribe 's suggestion https://github.com/nicotine-plus/nicotine-plus/issues/1628 perhaps the File Properties window can be called after selecting all files in a folder, or ideally the dialog can be enhanced to support being called on a folder.

pynicotine/gtkgui/userbrowse.py
slook: Indeed. I tried looking into it and it got hacky very quickly, a more robust solution would be desirable. Maybe it is not only TreeViews but also widgets like TextEntrys and buttons could benefit with some Ctrl actions or Shift overrides, maybe even some menu-items might benefit from the flexibility offered by an application wide solution.

mathiascode: It also needs to work in GTK4, so most of the existing solutions online are probably invalid at this point. A fun afternoon/weekend project in any case.

PR #1611, opened 2021-10-12
Update pl.po

mariachini:


mathiascode:
Thanks! We started using Weblate earlier today, in case you're interested in using it: https://hosted.weblate.org/projects/nicotine-plus/nicotine-plus/

PR #1621, opened 2021-10-14
Declunkify

hboetes:
Some sentences can use more usual English constructions.


hboetes:
I'm very happy with the results. As a big bonus: The auto-translated sentences make so much more sense now.

PR #1661, opened 2021-11-03
Browse Shares: Simplify accelerator functions

slook:
- Fixed: Alt+Enter currently has three completely different actions depending on what's focused. _Now there is only two._
- Removed: Ctrl+Enter to download folder recursively when folder contains zero files.
- Removed: Ctrl+Alt+Enter to download folder, because this is only for self browse local folder with Open in File Manager.
- Removed: Ctrl+Alt+Enter to download file, because this is only for self browse local file with Open in File Manager.
- Removed: Ctrl+Alt+Enter to open File Properties, because this function is not expected when Ctrl is pressed.

### DIFF
--- a/pynicotine/gtkgui/userbrowse.py
+++ b/pynicotine/gtkgui/userbrowse.py
@@ -867,16 +867,11 @@ class UserBrowse(UserInterface):
         return True
 
     def on_folder_open_manager_accelerator(self, *args):
-        """ Ctrl+Alt+Enter: Open folder in File Manager...
-                            Download Folder Into...    """
+        """ Ctrl+Alt+Enter: Open folder in File Manager... """
 
         if self.user == config.sections["server"]["login"]:
             self.on_file_manager()
-
-        elif len(self.file_store) >= 1:  # [user is not self]
-            self.on_download_directory_to()  # same as Ctrl+Enter
-
-        return True
+            return True
 
     """ Key Bindings (FileTreeView) """
 
--- a/pynicotine/gtkgui/userbrowse.py
+++ b/pynicotine/gtkgui/userbrowse.py
@@ -968,9 +968,6 @@ class UserBrowse(UserInterface):
         elif self.num_selected_files >= 1:  # [user is not self]
             self.on_file_properties()  # same as Alt+Enter
 
-        elif self.num_selected_files <= 0:
-            self.on_folder_transfer_to_accelerator()
-
         return True
 
     def on_file_properties_accelerator(self, *args):
--- a/pynicotine/gtkgui/userbrowse.py
+++ b/pynicotine/gtkgui/userbrowse.py
@@ -862,7 +862,7 @@ class UserBrowse(UserInterface):
             return True
 
         if len(self.file_store) <= 0:
-            # risk of accidental recursive download
+            # don't risk accidental recursive download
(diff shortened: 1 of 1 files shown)


pynicotine/gtkgui/userbrowse.py
```diff
@@ -881,7 +880,7 @@ def on_folder_open_manager_accelerator(self, *args):
             self.on_file_manager()
 
         elif len(self.file_store) >= 1:  # [user is not self]
-            self.on_download_directory_to()
+            self.on_download_directory_to()  # same as Ctrl+Enter
```
mathiascode, line 883: Ctrl+Enter has a recursive case (which I'm not sure is necessary, tbh), this does not. Also, it feels wrong to have this code in a function named "on_folder_open_manager_accelerator". Shouldn't shortcuts be clearly separated, and not trigger shortcuts further down the chain?

slook: > Ctrl+Enter has a recursive case (which I'm not sure is necessary, tbh), this does not

I see your point. The only reason why Ctrl+Enter has a recursive case is for when there are no files then transferring a folder would lead to an unhandled error where nothing happens after selecting a destination folder in the dialog. I think it is better that something always occurs, and this seems like a logical course of action, although admittedly it feels a bit confusing so it is not fully intuitive.

> it feels wrong

I know it seems wrong here, but falling back after returning False doesn't work. An alternative would be to trigger "on_folder_transfer_to_accelerator" which duplicates the condition.

> Shouldn't shortcuts be clearly separated?

Yes, but these aren't what I would call "shortcuts"... these are what I consider to be key "bindings", and my intention is overlap them and provide many alternatives so that it doesn't matter if the user doesn't know how the application's interface is supposed to be operated, they can just mash the keyboard and end up with something useful. In our case, the most useful thing would ideally be a file, or failing that then information about the file.

The downside of this approach is code spaghetti.

pynicotine/gtkgui/userbrowse.py
```diff
@@ -808,36 +804,27 @@ def on_folder_transfer_accelerator(self, *args):
...
+        """ Ctrl+Alt+Enter: Open folder in File Manager... """
 
         if self.user == config.sections["server"]["login"]:
             self.on_file_manager()
-
-        elif len(self.file_store) >= 1:  # [user is not self]
-            self.on_download_directory_to()
-
-        return True
+            return True
```
mathiascode, line 819: Move this outside the if clause, or return False by default, depending on what the intention is.

slook: Done. I see the sense in it, but in our case return True or False never has a useful effect because the key press event won't cascade once it has already parsed. If the conditions were refactored to allow falling back to happen then maybe the code would be more efficient, but it would make it even more difficult to follow the logic.

mathiascode:
Is this ready?

slook:
Yep, tried and tested.

slook:
There is a crash that occurs when pressing an accellator key on a empty model, when user is offline etc.

Will be fixed by https://github.com/nicotine-plus/nicotine-plus/pull/1672/commits/cdc4232288275cf41a20f93d6c3db1283dfedacb in https://github.com/nicotine-plus/nicotine-plus/pull/1672 or otherwise I'll do a separate patch.

PR #1668, opened 2021-11-05
utils.py: Days and Years in human_length

slook:
- Added: Long string format for length more than 24 Hours... 'D:HH:MM:SS' -> '% day %h %m'
- Added: Long string format for length more than 48 Hours... 'D:HH:MM:SS' -> '% days % hours'
- Added: Long string format number of Years if length is more than 365 Days - > '% year % days'
- Added: Long string format without Days if length is more than 2 Years -> '% years'

Long display format option with `long=True` argument if length is 1 Day or more, otherwise the output of the function is the same as it was before.

_Edit: Thank you for the below suggestions which have now been implemented._

### DIFF
--- a/pynicotine/utils.py
+++ b/pynicotine/utils.py
@@ -251,23 +251,25 @@ def make_version(version):
     return (major << 24) + (minor << 16) + (patch << 8) + stable
 
 
-def human_length(seconds):
+def human_length(seconds, long=False):
 
     minutes, seconds = divmod(seconds, 60)
     hours, minutes = divmod(minutes, 60)
     days, hours = divmod(hours, 24)
     years, days = divmod(days, 365)
 
-    if years > 1:
+    if long and years > 1:
         ret = '%i years' % (years)
-    elif years > 0:
+    elif long and years > 0:
         ret = '%i year %i days' % (years, days)
-    elif days > 1:
+    elif long and days > 1:
         ret = '%i days %i hours' % (days, hours)
-    elif days > 0:
+    elif long and days > 0:
         ret = '%i day %ih %im' % (days, hours, minutes)
+    elif days > 0:
+        ret = '%i:%02i:%02i:%02i' % (days, hours, minutes, seconds)
     elif hours > 0:
-        ret = '%02i:%02i:%02i' % (hours, minutes, seconds)
+        ret = '%i:%02i:%02i' % (hours, minutes, seconds)
     else:
         ret = '%i:%02i' % (minutes, seconds)
 
--- a/pynicotine/utils.py
+++ b/pynicotine/utils.py
@@ -244,7 +244,7 @@ def make_version(version):
     stable = 1
 
     if "dev" in version or "rc" in version:
-        # Example: 2.0.1.dev1
+        # Example: 3.2.0.dev1
         # A dev version will be one less than a stable version
         stable = 0
 
@@ -256,11 +256,18 @@ def human_length(seconds):
     minutes, seconds = divmod(seconds, 60)
     hours, minutes = divmod(minutes, 60)
     days, hours = divmod(hours, 24)
+    years, days = divmod(days, 365)
 
-    if days > 0:
-        ret = '%i:%02i:%02i:%02i' % (days, hours, minutes, seconds)
+    if years > 1:
+        ret = '%i years' % (years)
+    elif years > 0:
+        ret = '%i year %i days' % (years, days)
+    elif days > 1:
+        ret = '%i days %i hours' % (days, hours)
+    elif days > 0:
+        ret = '%i day %ih %im' % (days, hours, minutes)
     elif hours > 0:
-        ret = '%i:%02i:%02i' % (hours, minutes, seconds)
+        ret = '%02i:%02i:%02i' % (hours, minutes, seconds)
     else:
         ret = '%i:%02i' % (minutes, seconds)
 


mathiascode:
Keep in mind that human_length is also used for audio length, where these changes are probably not appropriate?

slook:
Yes it is appropriate, although I don't have any audio files that are more than 365 days long to test this!

For audio files more than a day long, the existing format 1:23:45:66 doesn't look correct.

Inso-m-niaC:
This function is also used in the Youtube plugin and the current format is identical to Youtube's player format. The proposed changes aren't suitable.

Inso-m-niaC:
You could add a new function, something like this

```python
def human_length_pretty(seconds):
    intervals = {
        'Year': 31536000,
        'Week': 604800,
        'Day': 86400,
        'Hour': 3600,
        'Minute': 60,
        'Second': 1
    }
    for unit, value in intervals.items():
        value = divmod(seconds, value)[0]
        if not value:
            continue
        if value > 1:
            unit += 's'
        return '%s %s' % (value, unit)
```

it's format clean and could maybe be used in the last seen feature in a new column as well.

Inso-m-niaC:
i wrote a couple more functions while twiddling my thumbs, extending to just days in this instance and mimicking timedelta in the second function. it's food for thought, i guess.

```python
def human_length(secs, timedelta=False):

    minutes, seconds = divmod(secs, 60)
    hours, minutes = divmod(minutes, 60)
    days, hours = divmod(hours, 24)

    if timedelta:
        import datetime
        return datetime.timedelta(seconds=secs)

    if days > 0:
        ret = '%i:%02i:%02i:%02i' % (days, hours, minutes, seconds)
    elif hours > 0:
        ret = '%i:%02i:%02i' % (hours, minutes, seconds)
    else:
        ret = '%i:%02i' % (minutes, seconds)

    return ret


def human_length2(secs, timedelta=False):

    minutes, seconds = divmod(secs, 60)
    hours, minutes = divmod(minutes, 60)
    days, hours = divmod(hours, 24)

    if timedelta:
        hms = '%i:%02i:%02i' % (hours, minutes, seconds)
        if days > 0:
            day_str = 'days' if days > 1 else 'day'
            return '%s %s, %s' % (days, day_str, hms)
        return hms

    if days > 0:
        ret = '%i:%02i:%02i:%02i' % (days, hours, minutes, seconds)
    elif hours > 0:
        ret = '%i:%02i:%02i' % (hours, minutes, seconds)
    else:
        ret = '%i:%02i' % (minutes, seconds)

    return ret


secs = 897654
print(human_length(secs, True))
print(human_length2(secs, True))

secs = 54
print(human_length(secs, True))
print(human_length2(secs, True))

```
output:
10 days, 9:20:54
10 days, 9:20:54
0:00:54
0:00:54

slook:
>` ret = '%i:%02i' % (minutes, seconds)`

For some reason, this condition isn't working in your output, but I can't figure out why...

>  0:00:54

I think I prefer the option of your seperate `human_length_pretty()` function

Inso-m-niaC:
`hms = '%i:%02i:%02i' % (hours, minutes, seconds)` is returned if the second parameter is True, not `ret = '%i:%02i' % (minutes, seconds)`.

slook:
Of course, yes it does work I didn't read your code properly. I do like this function because it allows the opportunity for the strings to be translated. "timedelta" has the wrong meaning, because it doesn't calculate the difference between two times like the Python timedelta function does, see link:
https://pynative.com/python-timedelta/#h-convert-string-to-timedelta

I'm wondering if theres something already standard that we could use instead of all this complexity, see link:
https://docs.python.org/3/library/datetime.html#strftime-and-strptime-behavior

However relying on the systems locale brings about unpredicatability.

Inso-m-niaC:
But datetime.timedelta isn't even being used, the second function merely returns the same format that datetime.timedelta would. Ignore the first function, it was just there to compare the format returned by the second function. While writing it i did consider using the variable `timedelta_format` but it is just an example function and wording isn't that important.

Can you elaborate on what this PR is for?

slook:
This is interesting, thank you for your suggestions I have a better understanding now.

> Can you elaborate on what this PR is for?

Various things, my primary intention is for use by /gtkgui/userbrowse.py that I am developing with the ability to display the total size of a folder on the frame's/an additional Statusbar widget. This will happen in userbrowse.set_directory() and select_files():

With human_size(long=True):
image
image


Without:
image
image

It's okay for ordinary album type folders:
image

We already store human_size in Tree for single files, so there is no need to call it again:
image

Showing Length seems appropriate for this application since many folders are music albums, and displaying the total play time is useful and interesting to see, but some folders can be incredibly large too!

Currently there is no way to determine the total size of a folder or multi-selection in Nicotine+ without starting a download and then cancelling it, length is just an extra bonus since it is inexpensive to calculate it from data already stored in TreeView.

long human_length is also intended for displaying the total size/length information for multiple files (or a whole folder) in the File Properties dialog, but the implementation of a Statusbar would reduce the need for that dialog box anyway.

The frame.Statusbar is currently unsuitable for into hints, so mat has rightfully suggested that Browse Shares is to have it's own Statusbar for each child page. Once the implementation is refined, then a similar thing can occur in Search Files.

Also, Transfer Statistics dialog will display the time since records began.

Other ideas might include; client session uptime; days since application install and buddy last seen as you suggested.

Inso-m-niaC:
Thanks for the explanation.

i quickly whipped up another

```python

def human_length(seconds, long=False):

    days, seconds = divmod(seconds, 86400)

    if long and days > 0:
        day_str = 'days' if days > 1 else 'day'
        return '%s %s, %s' % (days, day_str, human_length(seconds))

    minutes, seconds = divmod(seconds, 60)
    hours, minutes = divmod(minutes, 60)

    if days > 0:
        ret = '%i:%02i:%02i:%02i' % (days, hours, minutes, seconds)
    elif hours > 0:
        ret = '%i:%02i:%02i' % (hours, minutes, seconds)
    else:
        ret = '%i:%02i' % (minutes, seconds)

    return ret


secs = 86400 * 500 + 1
print(human_length(secs, True))

secs = 54
print(human_length(secs, True))

```

output:
500 days, 0:01
0:54

slook:
Marking as draft, since there is no immediate need for this at the moment.

I am not sure if the divmod function has a performance penalty, this needs further testing.

PR #1671, opened 2021-11-05
Log History Pane: Interface terminology of Debug Logging

slook:
- Changed: The term 'Debug Logging' is not appropriate, since the 'Log History' Pane is an integral part of the application's interface.
- Changed: Log category label 'Miscellaneous' -> 'Debug', because this is where debug log messages appear.
- Changed: Log category label 'Messages' -> 'Protocol', because it was ambiguous with Chats.
- Changed: The visual order of the categories, to group together the 3 categories that are excluded from the Statusbar.
- Changed: 'Show Log Pane' -> 'Show Log History Pane' in accordance with the correct terminology used elsewhere.
- Changed: 'Show Debug Log Controls' -> 'Show _Log Pane Controls'  because not every category is for debugging.
- Changed: 'View' -> 'Open' log File because this was ambiguous as to how to view the debug log category in the Pane.
- Added: 'Show_Log Pane Controls' toggle open in the popover context menu, to make the log categories easier to find.
- Fixed: Missing mnemonics in Log History Pane popover context menu.

### DIFF
--- a/pynicotine/gtkgui/frame.py
+++ b/pynicotine/gtkgui/frame.py
@@ -112,15 +112,17 @@ class NicotineFrame(UserInterface):
 
         # Popup menu on the log windows
         PopupMenu(self, self.LogWindow, self.on_popup_menu_log).setup(
-            ("#" + _("Find…"), self.on_find_log_window),
             ("", None),
-            ("#" + _("Copy"), self.log_textview.on_copy_text),
-            ("#" + _("Copy All"), self.log_textview.on_copy_all_text),
+            ("#" + _("_Find…"), self.on_find_log_window),
             ("", None),
-            ("#" + _("View Debug Logs"), self.on_view_debug_logs),
-            ("#" + _("View Transfer Log"), self.on_view_transfer_log),
+            ("#" + _("_Copy"), self.log_textview.on_copy_text),
+            ("#" + _("Copy _All"), self.log_textview.on_copy_all_text),
             ("", None),
-            ("#" + _("Clear Log View"), self.log_textview.on_clear_all_text)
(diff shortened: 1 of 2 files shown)


mathiascode:
> only one log category is for debugging purposes.

This is not true. Messages from all categories (except for downloads/uploads) are not translated, not guaranteed to be user-friendly, and may change at any moment. An ordinary user is not supposed to enable e.g. protocol and connection log messages and end up freezing their client in the process.

slook:
Yes indeed you are right. Perhaps some ToolTips on the CheckBoxes might be a good idea?

I was also considering a separator bar for the Connections, Protocol and Debug categories, or some other method of making these categories less prominent.

Perhaps an alternative way of toggling the Downloads/Uploads categories would negate the need for enabling the Show Debug Controls option... Imagine if the Downloads, Uploads options were always visible, perhaps docked the left or right hand side of the Pane?

mathiascode:
> would negate the need for enabling the Show Debug Controls option

Perhaps we could move the controls to a submenu in the context menu? Something like this:

Find
Copy
Copy All
Categories ->
- Downloads
- Uploads
- Search
- Chat
- Connections (debug)
- Transfers (debug)

etc.

In any case, we should probably change the "debug" log folder to a generic log folder, and replace the debug references in the logging preferences as well.

slook:
Gosh those menu actions are quite involved, I don't know where to start.

mathiascode:
> My only concern is if this would be cumbersome if the menu disappears on click.

You can press the space bar to bypass this.

slook:
Closing as superseded by https://github.com/nicotine-plus/nicotine-plus/commit/963c350f04c27085567bec023de1cf2ffe0ab07d

PR #1674, opened 2021-11-06
Access the Log Categories from View menu

slook:
Added: Item in View menu, since 4 options are relevant when the Log History Pane is hidden.

This would only be appropriate if the selected Log items did not affect file logging like they currently do.

### DIFF
--- a/pynicotine/gtkgui/frame.py
+++ b/pynicotine/gtkgui/frame.py
@@ -990,7 +990,8 @@ class NicotineFrame(UserInterface):
         menu.setup(
             ("$" + _("Prefer Dark _Mode"), "win.preferdarkmode"),
             ("$" + _("Use _Header Bar"), "win.showheaderbar"),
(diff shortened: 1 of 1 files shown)


pynicotine/gtkgui/frame.py
```diff
@@ -990,7 +990,8 @@ def create_view_menu(self):
         menu.setup(
             ("$" + _("Prefer Dark _Mode"), "win.preferdarkmode"),
             ("$" + _("Use _Header Bar"), "win.showheaderbar"),
-            ("$" + _("Show _Log History Pane"), "win.showlog"),
+            ("$" + _("Show Log _History Pane"), "win.showlog"),
+            (">" + _("_Log Categories"), self.popup_menu_log_categories),
```
slook, line 994: There is a mnemonic conflict here.

slook:
Closing since this is not appropriate with the current log implementation, nor will it be suitable after future refactoring.

PR #1678, opened 2021-11-07
Use strings that can easily be translated

hboetes:
The existing string, although well intended, was hard to understand. I tried explaining it differently, and then translated it to German, French and Dutch, to test if the translation still made sense. Since it didn't, I tried again with another string, and that one worked in all four languages.

### DIFF
--- a/pynicotine/gtkgui/userbrowse.py
+++ b/pynicotine/gtkgui/userbrowse.py
@@ -930,7 +930,7 @@ class UserBrowse(UserInterface):
     def on_download_directory_to(self, *args, recurse=False):
 
         if recurse:
-            str_title = _("Select Destination Folder for this Recursive Download")
+            str_title = _("Select Destination Folder for this Download (with subfolders)")
         else:
             str_title = _("Select Destination Folder for this Download")
 
--- a/pynicotine/gtkgui/userbrowse.py
+++ b/pynicotine/gtkgui/userbrowse.py
@@ -930,9 +930,9 @@ class UserBrowse(UserInterface):
     def on_download_directory_to(self, *args, recurse=False):
 
         if recurse:
-            str_title = _("Select destination folder for this recursive download")
+            str_title = _("Select Destination Folder for this Recursive Download")
         else:
-            str_title = _("Select destination folder for this download")
+            str_title = _("Select Destination Folder for this Download")
 
         choose_dir(
             parent=self.frame.MainWindow,
--- a/pynicotine/gtkgui/userbrowse.py
+++ b/pynicotine/gtkgui/userbrowse.py
@@ -930,7 +930,7 @@ class UserBrowse(UserInterface):
     def on_download_directory_to(self, *args, recurse=False):
 
         if recurse:
-            str_title = _("Select Destination folder for Downloading Multiple Folders")
+            str_title = _("Select Destination Folder for Downloading Multiple Folders")
         else:
             str_title = _("Select Destination Folder")
 
--- a/pynicotine/gtkgui/userbrowse.py
+++ b/pynicotine/gtkgui/userbrowse.py
@@ -930,9 +930,9 @@ class UserBrowse(UserInterface):
     def on_download_directory_to(self, *args, recurse=False):
 
         if recurse:
-            str_title = _("Select Destination for Downloading Folder with Subfolders from User")
+            str_title = _("Select destination folder for this recursive download")
         else:
-            str_title = _("Select Destination for Downloading a Folder from User")
+            str_title = _("Select destination folder for this download")
 
         choose_dir(
             parent=self.frame.MainWindow,
--- a/pynicotine/gtkgui/userbrowse.py
+++ b/pynicotine/gtkgui/userbrowse.py
@@ -930,9 +930,9 @@ class UserBrowse(UserInterface):
     def on_download_directory_to(self, *args, recurse=False):
 
         if recurse:
-            str_title = _("Select Destination Folder for this Download (with subfolders)")
+            str_title = _("Select Destination folder for Downloading Multiple Folders")
         else:
-            str_title = _("Select Destination Folder for this Download")
+            str_title = _("Select Destination Folder")
 
         choose_dir(
             parent=self.frame.MainWindow,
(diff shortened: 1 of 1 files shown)


pynicotine/gtkgui/userbrowse.py
```diff
@@ -930,9 +930,9 @@ def on_download_directory_to_selected(self, selected, recurse):
     def on_download_directory_to(self, *args, recurse=False):
 
         if recurse:
-            str_title = _("Select Destination for Downloading Folder with Subfolders from User")
+            str_title = _("Select destination folder for this recursive download")
         else:
-            str_title = _("Select Destination for Downloading a Folder from User")
+            str_title = _("Select destination folder for this download")
```
slook, line 935: Yes! I like this, although it is Missing Title Case.

pynicotine/gtkgui/userbrowse.py
```diff
@@ -930,9 +930,9 @@ def on_download_directory_to_selected(self, selected, recurse):
     def on_download_directory_to(self, *args, recurse=False):
 
         if recurse:
-            str_title = _("Select Destination for Downloading Folder with Subfolders from User")
+            str_title = _("Select destination folder for this recursive download")
```
slook, line 933: This is a very tricky one isn't it. I tried several things when I was writing these strings and nothing seems to make sense.

I am concerned that folks won't know what "recursive" means, this is not a very common English word. My intention was to make it clear that Sub Folders will be included in the download, it is hard to explain clearly.

hboetes: It is indeed. Even more so if you want to take translatability into account.

The word "recursive" can be found 37 times in ./pynicotine/ and it's not that uncommon in English. Even the simple english wikipedia explains it.

pynicotine/gtkgui/userbrowse.py
hboetes: That's not too hard:
`Select destination folder for this download of %(num_files)s, (%(size)s)`

hboetes: Title Case: https://titlecaseconverter.com/

> Title case is a style that is traditionally used for the titles of books, movies, songs, plays, and other works. In title case, all major words are capitalized, while minor words are lowercased. A simple example would be Lord of the Flies.

slook: Suggestion for future implementation:_
` Select Destination Folder for Downloading %(num_files)s Files (%(size)s)`

pynicotine/gtkgui/userbrowse.py
slook: Us and many users are all geeks so this technical language is understood by most, but not everybody understands the purpose of this function. It does cause confusion so I feel mats choice is correct in https://github.com/nicotine-plus/nicotine-plus/commit/9ad9389b9a4a0090577155330c41d267bfdf4b9e and the rest of the interface needs to follow suit.

I don't think it will ever be likely that the number/size of selected Subfolders and Files will be counted by this function, so this string is probably going to be fairly permanent.

hboetes: In the meanwhile, @mathiascode has replaced `recursive` with a construction containing something like `including subfolders` and I adjusted the PR to reflect that change.

pynicotine/gtkgui/userbrowse.py
```diff
@@ -930,9 +930,9 @@ def on_download_directory_to_selected(self, selected, recurse):
     def on_download_directory_to(self, *args, recurse=False):
 
         if recurse:
-            str_title = _("Select Destination for Downloading Folder with Subfolders from User")
+            str_title = _("Select Destination Folder for this Download (with subfolders)")
```
mathiascode, line 933: Perhaps "this" is not necessary here?

hboetes: That works somewhat in the English language, but makes translating the sentence a lot harder.

So, this is what happens to the version with this:
> Zielordner für diesen Download auswählen (mit Unterordnern) 

And this is the one without.
> Wählen Sie den Zielordner für den Download (mit Unterordnern)

So, the first sentence, because of the "this" the translator knows what belongs to what and produces a very clear translation.

With the second, the translator has to guess what belongs to what. It's still a correct sentence, but the sentence is already harder to read, since it does not start with the subject.

See also: https://www.dorisandbertie.com/goodcopybadcopy/2017/02/14/sentence-structure-how-to-make-your-sentences-much-easier-to-read

slook: In this case, it matters not that the sentance makes sense, rather what is more important here is the clarity of which direction the Folder is going. It is important that we must disambiguate between selecting a local Folder to Download _into_, as opposed to selecting a folder to be downloaded.

This was why I originally included "from".

hboetes: Your proposal looks good, but as is translates to:

> Wählen Sie das Ziel zum Herunterladen mehrerer Ordner vom Benutzer aus 

Assuming you don't speak German, that sentence misses the point completely.

But let's make two small changes:

> Select Destination folder for Downloading Multiple Folders from %(user)s

Now we specify the destination type, and don't add the word User, but do add the name of the user and presto:

> Wählen Sie den Zielordner zum Herunterladen mehrerer Ordner von %(user)s
> Selecteer de doelmap voor het downloaden van meerdere mappen van %(user)s 

That's what we wanted to say.

slook: I am concerned that putting obnoxious user names with special characters and whatnot in the dialog box title might cause unforeseen problems and be visually undesirable in many cases.

hboetes: OK, that statement contradicts what you previously proposed, but not to worry. The following sentence is very translatable.

> Select Destination folder for Downloading Multiple Folders

And I'm sure people will understand they are downloading from a user.

slook: I think we should avoid duplicating the word "folder" twice, because this creates ambiguity.

Why can we not have `Select Destination for Downloading Multiple Folders` ?

hboetes: Now that you mention it…

slook: Thank you for your input on this @hboetes your expertise with the language is very much appreciated :)

hboetes: Thanks. I was actually surprised how we could have this long discussion, but now that I think about it, it really makes sense, saying what you want to say in such a way that it can't be misunderstood, in a clear and concise way is really not that easy.

PR #1679, opened 2021-11-07
Don't use scroll-event to check auto-away

slook:
The recently connected click pressed gesture is plenty frequent enough to reset the away timer, in https://github.com/nicotine-plus/nicotine-plus/commit/c5ac3b79301a8e506ff5cd78b39dfa1baafdda31 is a very good idea.

The away timer works, however the notification tray icon doesn't, but that is a separate issue.


mathiascode:
I guess we'll see how this goes.

PR #1681, opened 2021-11-08
settings/search.ui: Layout tweaks and hints

slook:
- Changed: Layout of Search History options, to group widgets together horizontally to save vertical space.
- Changed: Layout of Search Result Filter options, because the Enable default filters toggle was in a misleading place. 
- Changed: Position and label of "Free slot" -> Longer string similar to tooltip and Result Filter Help.
- Changed: Position of Include and Exclude, because there's minimal purpose to set as defaults here. Added: "text" in label.
- Changed: Position and label of "regular expressions" -> "advanced RegEx", because regular is ambiguous with ordinary to a layman. Added: tooltip to explain.
- Added: Greyed out Filter labels when Default result filters off, because on some themes Text Entries don't look disabled. 

image
image


pynicotine/gtkgui/ui/settings/search.ui
```diff
@@ -227,9 +273,26 @@
...
                 <child>
-                  <object class="GtkEntry" id="FilterOut">
+                  <object class="GtkBox">
                     <property name="visible">1</property>
-                    <property name="tooltip-text" translatable="yes">Filter out results whose file paths contain the specified text. Multiple phrases and words can be specified, e.g. exact phrase|music|term|exact phrase two</property>
+                    <property name="focusable">0</property>
+                    <child>
+                      <object class="GtkLabel">
+                        <property name="visible">1</property>
+                        <property name="label">            </property>
```
slook, line 282: I would be glad of any suggestion how to do an indent, because I don't like these blank labels?

pynicotine/gtkgui/ui/settings/search.ui
```diff
@@ -365,62 +479,73 @@
...
+                    <property name="visible">1</property>
+                  </object>
+                </child>
               </object>
             </child>
+            <child>
+              <object class="GtkCheckButton" id="RegexpFilters">
+                <property name="label" translatable="yes">Allow advanced RegEx for Include and Exclude Text</property>
+                <property name="visible">1</property>
+                <property name="tooltip_text" translatable="yes">A RegEx (Regular Expression) is a sequence of meta-characters and characters, e.g. [ \d \D \s \S \w \W ] and operators { + * ? | ^ } ( )</property>
```
slook, line 539: A user thought that "regular" was the normal filter mode, so I have moved this option down to a lower position and given it a different label to disambiguate. Please review as I'm not sure how RegEx filtering works in N+

pynicotine/gtkgui/ui/settings/search.ui
```diff
@@ -185,24 +227,26 @@
...
-                <child>
-                  <object class="GtkEntry" id="FilterIn">
+                  <object class="GtkBox">
                     <property name="visible">1</property>
-                    <property name="tooltip-text" translatable="yes">Filter in results whose file paths contain the specified text. Multiple phrases and words can be specified, e.g. exact phrase|music|term|exact phrase two</property>
+                    <property name="focusable">0</property>
+                    <child>
+                      <object class="GtkLabel">
+                        <property name="visible">1</property>
+                        <property name="label">            </property>
```
mathiascode, line 236: We shouldn't do this. It's not reliable with different themes and fonts.

mathiascode:
> Clear Default Result Filters when Clear Filter History button is pressed, because it makes sense to do this here.

I'm not sure about this one. These actions are not the same, and I wouldn't expect my default filters to be wiped when clearing the history.

pynicotine/gtkgui/ui/settings/search.ui
slook: I can't get the spacing property to work. Indeed I tried on several themes and it is slightly different on each.

slook:
The placement the button appears as though this is what the function should be in this arrangement. You are correct it is unwise to combine these functions, ideally there should be a separate button for each.

Can these text entries have the built-in little clear widgets?

pynicotine/gtkgui/ui/settings/search.ui
mathiascode: The spacing property is for spacing between elements, margin-start/end adds spacing to the left and right of an element.

In any case, we should probably give this some thought before adding it. It would look out of place to only include such spacing in the search category.

mathiascode:
> Can these text entries have the built-in little clear widgets?

I want to add those for more text entries (https://github.com/nicotine-plus/nicotine-plus/discussions/1535#discussioncomment-1474126), but it needs a common implementation in Nicotine+. As far as I know, only GtkSearchEntry has them by default.

pynicotine/gtkgui/ui/settings/search.ui
slook: Thank you, perhaps in future I can look into doing the widget indentation of the settings categories overall, I agree it is best to avoid these awkward individual style hacks, so I have removed the label spacers.

hboetes:
- Punctuation characters → punctuation marks
- Certain clients don't send search results, if special characters (added comma)
- search results, if special (added comma)
- Clears → Clear

And there were a few more missing commas. Please have a look at that.

slook:
Thanks, I have updated the tooltip string this PR adds. Really it is beyond the scope of this PR to alter the strings, the layout was the main concern here, given this is likely to be the first Preferences page that is opened from the main cog button on the default main window functionality modus operandi of the client.

slook:
Closing due to conflicts. I will re-visit the label indent spacing at a later date, it is not important right now.

PR #1683, opened 2021-11-09
settings/userinterface.ui: Move TraySettings from Notifications

slook:
- Added: "Display tray icon" and "Minimize to tray on startup" moved into 'User Interface' (from 'Notifications')
- Changed: "View previous primary tab on startup" -> "Remember previous primary tab on startup"
- Removed: "Primary Tabs" catagory header label, because it was unnecessary, the tabs are a part of the "User Interface".

### DIFF
--- a/pynicotine/gtkgui/ui/settings/userinterface.ui
+++ b/pynicotine/gtkgui/ui/settings/userinterface.ui
@@ -20,29 +20,23 @@
           </object>
         </child>
         <child>
-          <object class="GtkCheckButton" id="DarkMode">
-            <property name="label" translatable="yes">Prefer dark mode</property>
-            <property name="visible">1</property>
-            <property name="tooltip_text" translatable="yes">Note that the operating system&apos;s theme may take precedence.</property>
-          </object>
-        </child>
-        <child>
           <object class="GtkFlowBox">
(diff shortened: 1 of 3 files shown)


mathiascode:
I like the grid layout for the tab position options, but many of the other changes to the tab option layout feel noisy and unorganized. We don't have checkboxes next to section headers anywhere else.

slook:
I know, my thoughts as well.

Take whatever you want from it., if you like I will strip it back to just the top three check boxes for the TraySettings?

pynicotine/gtkgui/ui/settings/userinterface.ui
```diff
@@ -20,16 +20,68 @@
...
+                        <property name="visible">1</property>
+                        <property name="use-underline">1</property>
+                        <signal name="toggled" handler="on_toggle_tray" swapped="no"/>
+                      </object>
+                    </child>
+                    <child>
+                      <object class="GtkLabel">
+                        <property name="visible">1</property>
+                        <property name="halign">start</property>
+                        <property name="label">    </property>
```
mathiascode, line 64: In the future, specify "spacing" for the container box instead of using this hack.

PR #1688, opened 2021-11-11
transferlist.py: Update with lazy_load and Refresh

slook:
- Fixed: Significant and unpleasant delay on switching to either of the transfers tabs when there is a very large data store.
- Fixed: Slight application freeze during startup while loading large transfers history.
- Added: Function for delayed loading of interface to fix hang on switching to a transfer tab with a large data store.
- Added: F5 shortcut key for manual Refresh.

This solves the problem of the slugglish tab switching, but it doesn't address another related performance issue caused by transfers.py calling update on the entire tree model for each individual file operation. Many times these clash.


slook:
This PR inspired performance improvements such as https://github.com/nicotine-plus/nicotine-plus/commit/f071b2b5b01557eb2315208a1b2af5e505325299 which properly address the underlying issues with row updating, so this workaround is no longer required.

PR #1689, opened 2021-11-12
Store and display speeds as integers

slook:
It is both uninteresting and expensive to know the speed to the nearest 0.1 B/s or 0.1 KiB/s, it's better to round the value up to a whole number for both internal storage and visual display purposes.

- Changed: Don't store bytes with a floating point in Transfers list items, it caused parent rows to be slightly off.
- Changed: Don't display any speed with a decimal place, since B/s; KiB/s; MiB/s etc already provide the accuracy needed.

- ToDo: Ideally, it is only needed to update if the speed in Transfers if it's different by more than 2 KiB/s or so (2048 bytes).

The aim is reduce the update frequency of values in Transfer rows, but the bytes are usually different anyway so it doesn't seem to make much difference. Unsure as to the best place to check for this.

_In future, it could be useful to reduce speed accuracy for less frequent and/or faster calcuations in the core (then this would stop Transfers from being provided long values by process_file_input() in slskproto), but that might have unforeseen consequences if the server requires a specific data type or some such._

Additional improvements for speed and limit calculations require further experimentation and is beyond the scope of this PR.

### DIFF
--- a/pynicotine/gtkgui/transferlist.py
+++ b/pynicotine/gtkgui/transferlist.py
@@ -377,14 +377,7 @@ class TransferList(UserInterface):
         return "%s / %s" % (human_size(currentbytes), human_size(size))
 
     def get_hspeed(self, speed):
-
-        hspeed = ""
-
-        if speed > 0:
-            speed = float(speed)
-            hspeed = human_speed(speed)
-
-        return hspeed
+        return human_speed(int(speed)) if speed > 0 else ""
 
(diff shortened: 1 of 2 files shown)


mathiascode:
Thanks, I made additional changes in https://github.com/nicotine-plus/nicotine-plus/commit/10f79cda2575b389bc10f4911e279ab23d711c14.

PR #1691, opened 2021-11-12
Don't display maximum sockets on SocketStatus bar

slook:
- Removed: Max sockets don't need to be displayed on SocketStatus bar
- Removed: slskproto doesn't need to be imported into frame, since the whole module is required just to get a static value.
- Added: Standard log entry to report maximum data sockets allowed by OS platform, upon establishing local port binding.

Before: image - - - After: image

Since the MAXSOCKETS value always seems to be the same (it is usually "1024" on most systems, but can be less in some rare cases), it is a waste of both time and space (and possibly resources).

**_DONE:_** Perhaps if knowing this value is important to people running Win32 or with some file limit set, maybe it would be sufficient to mention it in the Standard log, because it isn't something that is going to change during the session.

### DIFF
--- a/pynicotine/slskproto.py
+++ b/pynicotine/slskproto.py
@@ -558,7 +558,9 @@ class SlskProtoThread(threading.Thread):
                 self.listen_socket.bind((ip_address, listenport))
                 self.listen_socket.listen()
                 self.listenport = listenport
-                log.add(_("Listening on port %i"), listenport)
+                log.add(_("Listening on port: %i"), listenport)
+                log.add(_("Data socket limit: %(socks)s maximum set by %(platform)s system"),
+                        {'socks': MAXSOCKETS, 'platform': sys.platform})
                 break
 
             except socket.error:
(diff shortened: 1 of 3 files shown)


pynicotine/slskproto.py
```diff
@@ -558,7 +558,8 @@ def bind_listen_port(self):
                 self.listen_socket.bind((ip_address, listenport))
                 self.listen_socket.listen()
                 self.listenport = listenport
-                log.add(_("Listening on port %i"), listenport)
+                log.add(_("Listening on port: %i"), listenport)
+                log.add(_("Maximum data sockets: %i"), MAXSOCKETS)
```
mathiascode, line 562: In my opinion, this is still a difficult string for translators.

@hboetes what do you think?

mathiascode: > It's a log entry, didn't you plan on not translating log entries?

Log entries outside the [Debug] categories are still translated.

On second thought, I'll just move this to the "Miscellaneous" debug category.

PR #1693, opened 2021-11-13
Search Files: Add '+' indicator for hidden results

slook:
- Added: Append plus symbol "+" to Results count label if Filters is hiding results, or if reached 'Maximum per search' value.
- Added: Tooltip on Results CounterButton to display "Total: _n_" if visible results are filtered, or "Total: **>** _n_**+**" if over max.
- Added: Clicking Results CounterButton clears the filters if there are any filtered results, or opens Searches preferences.

_Edit: Tooltip text for total results when filtered:_
If max results reached: - - - If not max results reached:
image image


mathiascode:
> or if reached 'Maximum per search' value.

In my opinion, this is misleading. If the limit is reached, we don't store any additional results to display later.

The value of numtotalresults is incorrect when the limit is reached. It currently includes the max number of results + any additional discarded results from the last user that was added.

pynicotine/gtkgui/search.py
```diff
@@ -1136,7 +1138,23 @@ def select_results(self):
...
+
+        max_results = config.sections["searches"]["max_displayed_results"]
+
+        if self.numtotalresults > self.numvisibleresults:
+            # Append plus symbol "+" if Result Filter is active and/or reached 'Maximum per search'
+            str_plus = "+"
+
+            # Display the tooltip for total, but only if we know the exact number of total results
+            str_total = (_("Total: %s") %
+                (self.numtotalresults if not self.numtotalresults > max_results else \
```
mathiascode, line 1150: You can simplify this by turning the counter variable in add_result_list and add_user_results into self.numtotalresults instead.

slook: Sorry I don't follow exactly what you mean by that.

mathiascode: Sorry, I now noticed that you added some new syntax for cases where max results is reached.

I wouldn't rely on a slightly incorrect value in numtotalresults for this. What happens if the last user happens to have a number of results that leaves numtotalresults on exactly the max limit? Also, if numtotalresults were to be used elsewhere in the future, it would also give the wrong value in most cases without the developer being aware of this.

slook: > What happens if the last user happens to have a number of results that leaves numtotalresults on exactly the max limit

The possibility of this problem had occurred to me also, and as yet I have been unable to test this.

> numtotalresults ... would also give the wrong value in most cases

Then I suggest changing the name of this variable to num_displayed_results (and changing the other variables to snake_case also).

mathiascode: > > What happens if the last user happens to have a number of results that leaves numtotalresults on exactly the max limit
> 
> The possibility of this problem had occurred to me also, and I haven't as yet been able to induce this scenario, but in theroey the "+" symbol would not be displayed in this case, and the Total: value would be displayed in accordance to the results gathered, but may well not be aware that the limit had been exceeded, or was about to be.

It's unlikely that you'll encounter this without a bunch of bad luck, but it's always good to keep these kind of scenarios in mind for stability. :)

> Agreed, it makes me wonder if setting a flag during the counting would be better, rather than testing the amounts for this condition.

I would set a max_limit_reached flag or similar.

slook: > it's always good to keep these kind of scenarios in mind for stability. :)

Somehow it was possible for for result_list to be None when using the useless 'Clear All Results' function, so my previous len(result_list) crashed without a specific check for that, so I have decided to count the num_results_found at the end of the loop instead. 'counter' is best left alone to do it's own thing imo.

This is accurate and I haven't seen it fail, so this is ready to merge.

mathiascode: > 'counter' is best left alone to do it's own thing imo

To simplify things, I ended up removing it anyway.

PR #1699, opened 2021-11-14
Remove the term "popup" from all texts, it is redundant

hboetes:
Since the nicotine+ notifications use the system's notification mechanism and no pop-ups, like in a browser, the term 'popup[sic]' is redundant and misleading. Let's remove them from the texts.

I've also noticed the term is used in the code. I can imagine these can/should be removed as well.


hboetes:
oops, forgot to merge

PR #1700, opened 2021-11-14
no more popups

hboetes:
Since the nicotine+ notifications use the system's notification mechanism and no pop-ups, like in a browser, the term 'popup[sic]' is redundant and misleading. Let's remove them from the texts.


slook:
_I can confirm on Xfce these UI elements are indeed called "notifications":_
image

I think this goes for several other operating systems also.

PR #1701, opened 2021-11-14
use more accurate description for restoring the previously active tab

hboetes:


slook:
It is necessary to mention "primary tab" otherwise it might seem that the feature is not working properly, because this function is not able restore selection of any secondary tab. See https://github.com/nicotine-plus/nicotine-plus/commit/2bc4439da844ebc0fc5117cf855ab4fe68ab09e5 and https://github.com/nicotine-plus/nicotine-plus/discussions/1535#discussioncomment-1464711

To be honest the whole terminology of "primary tabs" and "secondary tabs" is undesirable in my opinion, but at least it is clear.

hboetes:
I had to think a bit about that and what description would be most common and this seems the most commonly used method,

So: Main tab, secondary tab.

https://edtech.gnomio.com/pluginfile.php/3825/mod_book/chapter/528/Grades-View_Tabs2.png

Primary/secondary tab might be confused with the first and second tab.

And remembering is the act of storing the information, restoring is the act of reproducing the stored info.

So, let's change the PR a bit.

hboetes:
Of course the changes caused a conflict. recreating with a new PR.

Issue #885, opened 2020-12-02
Finish running transfers and quit


hboetes:
Maybe it would be nice if there was an option to allow the currently running uploads to finish and then quit?

mathiascode:
This shouldn't be difficult to implement, it's mainly the design I've been thinking about. I wonder if a toggleable item in the menu would do?

![Screenshot from 2022-09-14 16-24-18](https://user-images.githubusercontent.com/8754153/190166134-2369da61-daaa-47dd-ba8a-510fc8690a8c.png)

I thought about a separate quit action in preferences -> user interface, but the main issue is that quitting would appear broken if uploads are active and the X button does nothing. There would also be no way to revert the action, and no indication that it's active.

slook:
> the main issue is that quitting would appear broken if uploads are active and the X button does nothing

And the other problem is that after any amount of time the application will suddenly quit, by which time it may be unexpected, so there does need to be an obvious visual indication that the application is waiting to quit.

The dilemma is understandable, but I do think a separate 'Finish Uploads and Quit' action could be preferred over the proposed menu item, if ideally the main window will be frozen and a dialog box would be displayed stating something like "Nicotine+ will quit after _x_ uploads are finished" and giving the options "Return to Nicotine+", "Hide Nicotine+" and "Quit Nicotine+ Now".

hboetes:
Perhaps add a notification bar at the download/upload window under the button bar, that says:

> n+ will close down as soon as these uploads/downloads have finished. Further uploads will be queued and started at the next start of n+
> To cancel this press: button

slook:
If a notifcation is used then it would need to be visible in every part of the application, not only the downloads/uploads tabs.

hboetes:
I'm not against that proposal.

slook:
The menu item idea would work if combined with an InfoBar warning of the impending quit, or some other equivalently ugly and obtrusive interface component.

What happens if new uploads are queued since the Quit request? I suppose the active upload count would need to reach zero?

h8uthemost:
Would love to have this. I actually came here to request this, glad I searched first.

mathiascode:
Having thought some more about this, here's an idea I had:

Menu -> Quit always shows a confirmation dialog, but never contains a checkbox. We could place the toggle for this option here:

![](https://github.com/nicotine-plus/nicotine-plus/assets/8754153/3d50ce39-9379-4968-81b8-73caf22985d2)

[...]

PR #1702, opened 2021-11-15
Fix the Quit and Hide mechanisms and key bindings

slook:
_Essential urgent fixes:_
- Fixed: `Ctrl+Q` shortcut key assignment respects the 'When closing Nicotine' option to avoid unexpected termination.
- Fixed: Close button respects the 'When closing Nicotine+' option.

_Important fixes:_
- Fixed: Quit from main menu respects the relevant 'When closing Nicotine+' option if confirmation is set or not.
- Removed: 'Remember choice' option when Quitting from main menu or tray, since this option applies only to 'Close'.

_Optional fixes:_
- Added: `Ctrl+Alt+Q` shortcut key for force Quit (without prompt), regardless of preference setting.
- Added: `Shift+Ctrl+Q` shortcut key for Close / Run in Background options in dialog (to override the remembered setting once).

_Additional fixes:_
- Changed: Always show Exit confirmation dialog (but don't show Remember choice checkbox) on select Quit in the Tray menu.
- Reverted for now: "Disable Tray Icon" dialog button when the Exit dialog is activated from the Tray menu Quit item.
- Fixed: Main menu separator item above Quit wasn't visible.
- Added: Ellipsis on the main menu Quit... item to indicate if the exit confirmation dialog is enabled or not.


mathiascode:
The "Run in Background" mode was decoupled from the tray icon on purpose (and replaced all references of "Minimize to Tray"), since tray icons are a dead standard, certain desktops have removed them entirely (GNOME, elementary OS), and they're not guaranteed to be available on a user's system if they customize it. Starting Nicotine+ from anywhere in the OS should bring a running instance to the foreground again.

mathiascode:
https://github.com/nicotine-plus/nicotine-plus/commit/df12b66398cb1fe06721f4f81b8b7e20ce00aae6

slook:
Thank you for looking at this @mathiascode 

> The "Run in Background" mode was decoupled from the tray icon on purpose... ...tray icons are a dead standard

I agree the idea of tray icons is old hat and I also prefer the term "Run in Background". Done.

The application is far too easy to quit or hide accidentally. No matter what the feature is called I still think it is a bad idea to give an option in the exit dialog which allows the window to go "zombie" on the vast majority of environments.

It would be expected that the tray icon would appear if it wasn't already active, if possible, unless the user has set a preference to specifically disallow this. Currently, the application is far too easy to quit or hide accidentally. A 'Minimize to system tray' exit option (exitdialog=3) is needed in Preferences to align with the 'Minimize to tray on startup' option. This makes sense on testing.

Hiding the window shouldn't be the default behaviour, I feel the application should make an effort to keep the window accessible so that the user doesn't have to terminate it externally or start it up again, because this could lead to some users thinking that the application has crashed, especially if this was inadvertantly set to the remembered exit choice.

> Quit should always quit the program

Agreed, in my latest draft I have removed Quit menu item from the Remember choice mechanism, and this always respects the users remembered preference, for confirm or not confirm, I had thought this was an oversight because of the key assignment.

I have added a confirmation to the exit tray Quit... menu item, with a feature to 'Disable Tray Icon' because this would be the first thing people look for if they don't want the icon. Selecting this disables the remembered choice (if that choice was Run in Background), such that the next time Ctrl+Q is pressed they will be prompted to remember choice again.

> Closing the window using the X button should respect the option

Done. This now works as expected, including with the `Ctrl+Q` shortcut.

pynicotine/gtkgui/widgets/dialogs.py
```diff
@@ -230,3 +230,89 @@ def option_dialog(parent, title, message, callback, callback_data=None,
         self.add_button(third, Gtk.ResponseType.REJECT)
 
     self.present_with_time(Gdk.CURRENT_TIME)
+
+
+def custom_dialog(parent, title, message, callback, callback_data=None, cancel=True, selectable=True, third="",
```
mathiascode, line 235: This is a bit too massive for my liking. I think the addition of tooltips in a message dialog is a sign that things are becoming too complicated.

I made the default buttons easier customize in https://github.com/nicotine-plus/nicotine-plus/commit/5981796a55a587e9f35115ce3b4a421a39e548d1. What else is missing from the existing dialogs?

slook: I agree, I don't like the scrappyness of the strings in the code. Really it is a proof of concept, I don't think the tool-tips look like they belong in dialogs, it is a bit unconventional.

> What else is missing from the existing dialogs?

The ability to be able to hide the checkbox, because NoneType isn't possible in the callback response.

pynicotine/gtkgui/widgets/dialogs.py
```diff
@@ -230,3 +230,89 @@ def option_dialog(parent, title, message, callback, callback_data=None,
...
+    if checkbox_label:
+        self.checkbox = Gtk.CheckButton()
+        self.checkbox.set_label(checkbox_label)
+        self.checkbox.set_active(checkbox_value)
+        self.checkbox.set_tooltip_text(checkbox_tip)
+
+        self.get_message_area().add(self.checkbox)
+        self.checkbox.show()
+    else:
+        self.checkbox = None  # option can be hidden
```
slook, line 284: > What else is missing from the existing dialogs?

The ability to be able to hide the checkbox, because NoneType isn't possible in the callback response.

pynicotine/gtkgui/widgets/dialogs.py
```diff
@@ -230,3 +230,89 @@ def option_dialog(parent, title, message, callback, callback_data=None,
...
+        self.button_5 = self.add_button(id_5, 5)
+        self.button_5.set_tooltip_text(tip_5)
+    if id_6:
+        self.button_6 = self.add_button(id_6, 6)
+        self.button_6.set_tooltip_text(tip_6)
+    if id_9:
+        self.button_9 = self.add_button(id_9, 9)
+        self.button_9.set_tooltip_text(tip_9)
+
+    self.set_default_response(default_button)
```
slook, line 316: - The ability to set which default_button is focused on Enter key-press

pynicotine/gtkgui/widgets/dialogs.py
mathiascode: Thanks, this sounds like 3.2.1+ territory.

slook: Yes I am not suggesting we replace option_dialog() now, because this would be too risky for the other dialogs. That is why I have called it custom_dialog() for the time being, it works for the purposes of the exit dialog.

However, I do feel that the exit dialogs 'Remember choice' functionality and the `Ctrl+Q` shortcut does urgently need fixing for the release of 3.2.0 , because the current quitting behaviour is unacceptable with the program exiting so easily like it does at the moment.

mathiascode: > urgently need fixing for the release of 3.2.0

What's left on your list of urgent fixes? I have a rather hectic month in front of me, and I'd like to release 3.2.0 soon.

slook: > What's left on your list of urgent fixes?

Well anything that causes the application to terminate unexpectedly would be considered "urgent" (even if it's not technically 'critical'), and this PR resolves the only issue I am currently aware of that causes this @mathiascode 

I have reverted the Disable Tray Icon, gone back to the standard option_dialog and minimized string changes to only those that are necessary to perform the required functions ("Quit" and "Close").

> I have a rather hectic month in front of me, and I'd like to release 3.2.0 soon.

Me too, I feel these last development pushes has been very productive indeed. After your recent optimizations the application is operating and running better than it ever has done.

I don't want to push any more changes unless something 'urgent' or 'critical' comes up during final testing.

slook:
I don't know why the MacOS in-tree unit testing failed? it seems unrelated to the changes in frame.py

pynicotine/gtkgui/frame.py
```diff
@@ -1980,47 +1999,75 @@ def on_critical_error_threading(self, args):
...
 
             if self.MainWindow.get_property("visible"):
-                self.MainWindow.hide()
+                self.hide()
+
+    def exit_dialog(self, remember=None, tray_quit=None):
+
+        option_dialog(
+            parent=self.MainWindow,
+            title=_('Close Nicotine+') if remember else _('Quit Nicotine+'),
```
mathiascode, line 2026: Let's just use "Quit Nicotine+".

pynicotine/gtkgui/frame.py
```diff
@@ -1980,47 +1999,75 @@ def on_critical_error_threading(self, args):
...
+            message=_('Do you really want to exit?'),
+            second_button=_("_Quit"),
+            third_button=_("Run in _Background") if self.MainWindow.get_property("visible") else "",
+            checkbox_label=_("Remember choice") if not tray_quit and remember is True else "",
+            callback=self.on_exit_dialog_response
+        )
+        return True
+
+    def on_tray_quit(self, *args):
```
mathiascode, line 2035: Move this to trayicon.py.

pynicotine/gtkgui/frame.py
```diff
@@ -939,16 +955,19 @@ def add_preferences_item(self, menu):
         menu.setup(("#" + _("_Preferences"), "app.settings"))
 
     def add_quit_item(self, menu):
-        menu.setup(("#" + _("_Quit"), "app.quit"))
+
+        ellipsis = "…" if config.sections["ui"]["exitdialog"] else ""
+
+        menu.setup(
+            ("", None),
+            ("#" + _("_Quit") + ellipsis, "app.quit")
```
mathiascode, line 963: You already defined a translated string with the ellipsis included in trayicon.py, use it instead.

mathiascode: Do tray menus support mnemonics? I can't test right now.

pynicotine/gtkgui/frame.py
```diff
@@ -754,10 +754,26 @@ def set_up_actions(self):
...
-        self.application.set_accels_for_action("app.quit", ["<Primary>q"])
+
+        # Window (system menu and events)
+
+        action = Gio.SimpleAction.new("close", None)  # 'When closing Nicotine+'
+        action.connect("activate", self.on_close_request)
+        self.application.add_action(action)
+        self.application.set_accels_for_action("app.close", ["<Primary>q"])
+
+        action = Gio.SimpleAction.new("exit_dialog", None)  # Show all exit options
```
mathiascode, line 768: Wouldn't it be better to have a shortcut that simply makes Nicotine+ run in the background instead?

pynicotine/gtkgui/frame.py
slook: No, they do not.

pynicotine/gtkgui/frame.py
slook: I tried this, but if the tray icon is not visible for some reason then it can lead to yet another situation where the application window disappears unexpectedly.

Until we have a separate Minimize to Tray function (that is able to show/hide the tray icon) then I feel a prompt is required incase the user pressed the wrong key.

mathiascode: Isn't <Shift><Primary>q a rather obscure shortcut already? I don't foresee a situation where a user accidentally enables it.

If there's a concern that the user accidentally hides the window instead of quitting the application, perhaps the shortcut could use another key that is rarely used.

slook: > Isn't <Shift><Primary>q a rather obscure shortcut?... I don't foresee a situation where a user accidentally enables it. 

It is right next to `<Shift><Primary>w` (show Wishlist), and it is near to the numeric keys on the row above so it is possible to press Q accidentally whilst Shift and Ctrl are also pressed, especially if sticky keys are enabled.

> perhaps the shortcut could use another key

The only logical key I can think of would be some key modified `F4` combo but that is likely to get into key conflict territory. I have also implemented `<Primary><Alt>Q` for force-quit, so if the remembered choice is set to 'Run in Background' then it is possible do any of the three actions available in the 'When closing Nicotine+' options at any time.

> perhaps we don't need the shortcut at all?... it's probably unlikely that the user wants anything else than the remembered choice.

Right now this is the case, but in the future if the exit dialog gets more buttons or a dropdown for other exit options, such as "[Finish running transfers then quit](https://github.com/nicotine-plus/nicotine-plus/issues/885)", or other suchlike, then this shortcut would serve a useful purpose to be able to override the remembered choice with some other action.

mathiascode: > Right now this is the case

Let's remove it for now. Options such as "Finish running transfers then quit" will require further design discussions.

pynicotine/gtkgui/frame.py
mathiascode: You've already added a new string anyway, so might as well finish the work properly here. Otherwise we have to change it later if RTL languages are added.

kiplingw:
@slook, it would be appreciated if there was an option to remove the Exit confirmation dialogue when selecting to exit. I noticed it recently when I clicked exit from the status bar menu. It's annoying when I am trying to close a bunch of applications quickly before a shutdown or restart.

mathiascode:
I'm leaning towards removing the exit dialog from the tray menu's "Quit" item entirely, since it should be pretty clear what it does without requiring confirmation. Since other applications also quit immediately when pressing this item in their tray menus, it should be a familiar concept for the user.

PR #1706, opened 2021-11-16
plugin.ui: "Preferences" button / "Properties" window -> "Settings"

slook:
Fixed: The window is called Plugin "Properties", the "Preferences" button label doesn't match the internal "Settings" terminology.

_Changed:_
image


mathiascode:
Perhaps it should be the other way around? Not that it matters that much, but I associate "Properties" with read-only.

slook:
True, but then there will be two things both called "Preferences" (the main settings, and the plugin options).

What about "Options"?

Issue #1698, opened 2021-11-14
warning if no files are shared


hboetes:
Sometimes, for whatever reason, I don't share files at startup. A share may be inaccessible, or the database may have become corrupted. And I don't notice until I download something from someone who has a leech plugin. I'd rather get a warning from nicotine+

Therefore I'd like to request showing a warning at start-up, if no files are shared.

slook:
If this gets implemented as a "popup notification" then it must be off by default, because messages from Leech Detectors is already bad enough for nagging new users with fresh installs.

```
15:05:41 0 folders found after rescan
15:05:41 Finished rescanning shares 
```
It would be useful if these entries were the other way around, or combined into one line, so that it would stick on the Statusbar.

hboetes:
Right, not a system-notification, nor a "pop-up", since both interrupt the work-flow and are annoying.

On the other hand, not sharing files is something important. For example, the N+ icon dot could go red.
![image](https://user-images.githubusercontent.com/2285225/141687999-929c97ae-f7d7-4c9c-bdf0-bff9dc590122.png)

slook:
> the N+ icon dot could go red

The user is likely to wonder why it looks like they are disconnected when they are not.

I agree that Red is a suitable color for an error condition, but this already has the meaning of "Disconnected", so if we start using for displaying application status in addition to network status, then users will complain that it doesn't change colour during other sorts of error conditions that they might feel are equally as important to them, a failed transfer for example.

The tray icon mechanism is already problematic enough as it is, for example it doesn't work on MacOS, it incorrectly displays the red dot unless it was already enabled before startup, and the Quit mechanisms are non-nonsensical in several scenarios. This is something I have been working on recently, so I know that implementing something like this will be overly complex.

I think we should stick to improving the log system and what text is displayed on the Statusbar, these are areas where improvements are less prone to introducing strange application behaviour.

hboetes:
Having a clear message in the status bar sounds like a good plan indeed.

slook:
_Changed the log output of the scanner to this:_
![image](https://user-images.githubusercontent.com/88614182/142294684-d05da7e4-0607-4c50-8fc4-4e1bbc2d827c.png)

Perhaps we will append the number of errors encountered, but I don't think that information is counted at the moment.

slook:
[...]

PR #1707, opened 2021-11-17
shares.py: Display finished count after rescan

slook:
- Removed: "Finished rescanning shares" message, because the "folders found" count is more useful to end with on the Status bar after a manual scan, helps with #1698 , since the user will be able to see if the number of folders counted is correct or not after a manual rescan at least.

This doesn't fully address the issue for scan at startup, because if there is a folder not found error then it is very likely that the rescan will be completed before other startup messages appear that wipe out any Scanner messages from the Statusbar.

### DIFF
--- a/pynicotine/shares.py
+++ b/pynicotine/shares.py
@@ -104,8 +104,7 @@ class Scanner:
                 _new_mtimes, new_files, _new_streams = self.rescan_dirs("buddy", new_mtimes, new_files,
                                                                         new_streams, self.rebuild)
 
-                self.queue.put((_("%(num)s folders found after rescan"), {"num": len(new_files)}, None))
-                self.queue.put((_("Finished rescanning shares"), None, None))
+                self.queue.put((_("%(num)s folders found after rescan, finished."), {"num": len(new_files)}, None))
 
                 self.create_compressed_shares()
 
--- a/pynicotine/shares.py
+++ b/pynicotine/shares.py
@@ -104,7 +104,7 @@ class Scanner:
                 _new_mtimes, new_files, _new_streams = self.rescan_dirs("buddy", new_mtimes, new_files,
                                                                         new_streams, self.rebuild)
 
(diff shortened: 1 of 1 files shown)


pynicotine/shares.py
```diff
@@ -104,8 +104,7 @@ def run(self):
                 _new_mtimes, new_files, _new_streams = self.rescan_dirs("buddy", new_mtimes, new_files,
                                                                         new_streams, self.rebuild)
 
-                self.queue.put((_("%(num)s folders found after rescan"), {"num": len(new_files)}, None))
-                self.queue.put((_("Finished rescanning shares"), None, None))
+                self.queue.put((_("%(num)s folders found after rescan, finished."), {"num": len(new_files)}, None))
```
mathiascode, line 107: We don't include full stops in log messages, unless there are multiple sentences.

slook: This isn't really a very good sentence anyway. Maybe the word ", finished" isn't needed at all.

mathiascode: It can probably be removed.

slook:
```
[2021-11-17 22:39:53] Connecting to server.slsknet.org:2242
[2021-11-17 22:39:53] Rescanning shares…
[2021-11-17 22:39:53] 25 folders found before rescan, rebuilding…
[2021-11-17 22:39:53] Error while scanning folder /media/user/USER/Music PLAY: [Errno 2] No such file or directory: '/media/user/USER/Music PLAY'
[2021-11-17 22:39:53] Rescan progress: 4 %
```
...
```
[2021-11-17 22:39:53] Rescan progress: 96 %
[2021-11-17 22:39:53] Rescan progress: 100 %
[2021-11-17 22:39:53] Rescan complete: 25 shared folders
[2021-11-17 22:39:53] Connected to server server.slsknet.org:2242, logging in…

```
image

PR #1711, opened 2021-11-19
Remove feature to auto-share download folder

mathiascode:
This is an old feature from the PySoulSeek days that causes more harm than good.

- It's somewhat painful to maintain and keep in sync with the rest of the scanning code
- With the addition of the username subfolder feature for downloads, it causes privacy concerns
- The general consensus is that downloads should be moved to another folder and reviewed before sharing, since sharing the download folder directly can contribute to issues with incomplete albums
- The original Soulseek client had this feature, but removed it

### DIFF
--- a/pynicotine/config.py
+++ b/pynicotine/config.py
@@ -180,7 +180,6 @@ class Config:
                 "downloaddir": os.path.join(self.data_dir, 'downloads'),
                 "uploaddir": os.path.join(self.data_dir, 'received'),
                 "usernamesubfolders": False,
-                "sharedownloaddir": False,
                 "shared": [],
                 "buddyshared": [],
                 "uploadbandwidth": 10,
@@ -479,10 +478,19 @@ class Config:
         load_file(self.filename, self.parse_config)
 
         # Clean up old config options
-        self.remove_old_options()
+        migrate_download_share = self.remove_old_options()
 
         # Update config values from file
         self.set_config()
+
+        # Convert special download folder share to regular share
+        if migrate_download_share:
+            shares = self.sections["transfers"]["shared"]
+            virtual_name = "Downloaded"
+            shared_folder = (virtual_name, self.sections["transfers"]["downloaddir"])
+
+            if shared_folder not in shares and virtual_name not in (x[0] for x in shares):
+                shares.append(shared_folder)
 
         # Load command aliases from legacy file
         try:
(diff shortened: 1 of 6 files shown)


francisuk1989:
Soulseek hasn't yet removed the "username subfolder feature for downloads," however i find this feature useful rather then a privacy concern however i think another way doing it, is

1) Disable it by default and leave it Enabled for them people who do use it.

2) incomplete downloads have a random filename and hes way;

If i searched for "HiFi - you got the sponge"

if two searches matches the same results from different users then downloading them will conflicted with the same file from two users?

mathiascode:
The username subfolder feature is already disabled by default, and will not be removed. This PR removes auto-sharing of your download folder (the "Automatically share completed downloads" checkbox in Preferences -> Transfers -> Downloads (old Nic+ versions) / Preferences -> Shares (new Nic+ version)).

The "privacy concern" I refer to in my original message is when these username folders are auto-shared, revealing the origin of the download (I've seen complaints about this before).

slook:
I agree with the sentiment of this PR and the comments about it so far.

However it is the easiest way for new users to start sharing, so I would be concerned that removing this feature would increase the number of complaints about users who don't share anything. Also, it would cause existing users inconvenience if their shares stop working, this would be difficult for them to trouble-shoot if they were unaware of the changes.

If this feature is removed, then some new feature needs to be implemented that enables the easy sharing of Downloads from some default sharing folder that is set upon install (that can be changed). For example a "Share File(s)" item in the Downloads context menu or a "Share" button in the Downloads bar, that moves the file into the shared folder.

mathiascode:
> However it is the easiest way for new users to start sharing, so I would be concerned that removing this feature would increase the number of complaints about users who don't share anything.

I hardly see anyone with a "Downloaded" folder when performing a search in the "nicotine" room, and ones that have it usually share other folders too.

slook:
> I still see the folder without issues on master.

For some reason on my dev install it was removed from the Shares list, but is still shared because the "Automatically share completed downloads" checkbox is enabled. It will be interesting to see if the folder stops being shared once this option is removed, perhaps I will test this branch to find out...

mathiascode:
> if the folder stops being shared once this option is removed

Yes, but I could probably add some migration code to add the folder to the regular shares list.

slook:
What is the `git` command to `clone` the `remove-download-share` branch?

mathiascode:
> I have tested the remove-download-share branch and confirm that indeed the virtual 'Downloaded' folder is no longer shared, it was previously pointed to a non-default folder called "Complete" in the filesystem, as per the 'Download folder:' setting in the Downloads preferences.

I've added some migration code.

slook:
Yes, the migration resulted in the folder being successfully added into Shares as "Downloaded".

francisuk1989:
> The "privacy concern" I refer to in my original message is when these username folders are auto-shared, revealing the origin of the download (I've seen complaints about this before).

Makes sense. Thanks

PR #1716, opened 2021-11-20
use more accurate description…

hboetes:
…For restoring the previously active main tab

This is a second attempt for this PR, since 1701 created a huge conflict after applying proposed changes.

I'll create it as a draft and after everyone agrees, I will update the pot-file.

### DIFF
--- a/pynicotine/gtkgui/ui/settings/userinterface.ui
+++ b/pynicotine/gtkgui/ui/settings/userinterface.ui
@@ -59,8 +59,8 @@
                 </child>
                 <child>
                   <object class="GtkCheckButton" id="TabSelectPrevious">
-                    <property name="label" translatable="yes">Remember previous primary tab on startup</property>
-                    <property name="tooltip_text" translatable="yes">Start with Search Files by default.</property>
+                    <property name="label" translatable="yes">Restore the previously active Main Tab on start-up</property>
+                    <property name="tooltip_text" translatable="yes">By default the Search Files Tab is activated at start-up.</property>
                     <property name="visible">1</property>
                   </object>
                 </child>
--- a/pynicotine/gtkgui/ui/settings/userinterface.ui
+++ b/pynicotine/gtkgui/ui/settings/userinterface.ui
@@ -109,7 +109,7 @@
                 <child>
                   <object class="GtkLabel">
                     <property name="visible">1</property>
-                    <property name="label" translatable="yes">Visible primary tabs:</property>
+                    <property name="label" translatable="yes">Visible main tabs:</property>
                     <property name="xalign">0</property>
                     <property name="mnemonic_widget">EnableSearchTab</property>
                   </object>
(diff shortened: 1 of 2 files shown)


slook:
> "Restore the previously active Main Tab on start-up"

Whilst I don't have any particular with this statement (apart from "restore" vs "remember"), and I am in support of "Main", changing this string would mean changing many other strings across the entire application to match references to "Primary tabs" and "startup". Including in the Keyboard Shortcuts window and for other preferences.

Technically the main notebook are tabs, but some might consider these widgets to be more like buttons, the different windows to be the old "mode" terminology, idk, it is difficult to be clear about this, whilst not confusing the "secondary tabs" which form the main navigation of each main window mode.

I don't think this has ever really been particularly clear until the "primary" and "secondary" terminology, it's just a shame that those designations seem somewhat awkward in my opinion.

Given that the state of the secondary tabs are not restored in some of the application modes, this feature is only able to "remember" which mode was previously selected.

hboetes:
That the code does not exactly match the text is not a problem, this has been done before. For example, `popup` is still everywhere in the code, but not in the texts. 

Here are the only four instances of `Primary`. If Mathias agrees, I can create a PR to replace them all, after 3.2.0 that is.

```
msgid "Change Primary Tab"
msgid "Primary Tabs"
msgid "View previous primary tab on startup"
msgid "Visible primary tabs:"
```

slook:
> ... only four instances of Primary ... to replace them all, after 3.2.0 that is.

That's not so bad then. Replace them with what?

hboetes:
Please check PR 1701 for the description. I had to close it because of a large conflict.

slook:
> So: Main tab, secondary tab.

I prefer "Main" better than "Primary".

hboetes:
Excellent, I'll change the other instances of primary as well.

slook:
Obviously I don't have the final say, that is just my opinion.

pynicotine/gtkgui/ui/settings/userinterface.ui
```diff
@@ -59,8 +59,8 @@
                 </child>
                 <child>
                   <object class="GtkCheckButton" id="TabSelectPrevious">
-                    <property name="label" translatable="yes">Remember previous primary tab on startup</property>
-                    <property name="tooltip_text" translatable="yes">Start with Search Files by default.</property>
+                    <property name="label" translatable="yes">Restore the previously active Main Tab on start-up</property>
```
mathiascode, line 62: Any reason for capitalizing main tab?
Isn't "startup" more common in the context of software?

hboetes: > Any reason for capitalizing main tab?
 I did that because it's the subject of the sentence. Maybe I'm a bit confused by the title case. :-}

> Isn't "startup" more common in the context of software?
Yes, you are right, in the context of computers it is.

hboetes: Then again, 'Search Files' in the next sentence uses Title Case. What is your reasoning behind the use of Title Case or not?

slook: To identify the label name of the main tab that will be started with, as opposed to some other action related to file searching in general.

Title Case is generally not used in the interface apart from the labels of the main tabs and other UI components such as the Log History pane for example.

PR #1743, opened 2021-12-02
Hyperlinks to TRANSLATORS.md and ABOUT.ui credits list

slook:
It makes more sense for these code blocks to be links instead, to make it easier for the translators to add themselves.

If you would prefer hyperlinks instead of local links then I can change it if that would be better.


mathiascode:
If your intention is to provide the GitHub view for these files, the current links won't on the website.

slook:
You're right, having hyperlinks here is probably better in most cases.

PR #1752, opened 2021-12-05
shortcuts.ui: Move item into correct group

slook:
- Changed: Move "Remove Selected Row" to 'Transfers', because this shortcut doesn't apply to other lists
- ToDo: Rename to "Clear / Remove Transfer", but translations are frozen at the moment


mathiascode:
I believe it applies to the wishlist as well, but we should implement it for more lists. I think we still have two lists that use inline editing in the preferences dialog; these should be converted to an entry dialog before a delete shortcut is added.

slook:
You are correct. I will close this then.

PR #1756, opened 2021-12-05
Explain underscores in translation strings.

hboetes:
After a discussion on IRC, we decided it's a good idea to explain the underscores a bit.

### DIFF
--- a/doc/TRANSLATIONS.md
+++ b/doc/TRANSLATIONS.md
@@ -36,8 +36,14 @@ Using consistent terms for the same things makes the translation easier to under
 
 You can use the "Search and Replace" tool to help you with that.
 
-### Underscores(_) in the strings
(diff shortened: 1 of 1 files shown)


slook:
Please consider removing "so it doesn't really matter what they are" from the sentence, because this gives the impression that no thought is required when testing for conflicts and ergonomics, whereas a lot of thought goes into making sure the mnemonics make sense. If anything, more emphasis is needed to highlight that it is important that they are properly tested.

hboetes:
OK. Truth be told, I have no idea how to use the mnemonics. I'm looking forward to your proposal, please explain it as well as you can.

slook:
The portion of a label that enables keyboard selection of a menu item or button is defined by an underscore character "_" before the mnemonic letter to be accessed with a keyboard accelerator, for example Alt+F for the native "_File" menu. It is important that selectable items within the same menu or interface area each have unique mnemonics to avoid any key conflicts that may cause navigation problems for users who are unable to use a mouse or other pointing device.

If possible, the first letter in the string will usually be the most ideal choice for a mnemonic key, but if that is already in use by a conflicting item then put the underscore in front of another letter that makes logical sense, and is easy for the user to remember whilst avoiding conflict or confusion with other mnemonic keys or shortcut keys that are used in the user interface. In some cases, the same letter may appear in your language string that is used in the English string, which may or not be appropriate.

hboetes:
I just discovered that you can see the mnemonics if you press just the alt-key. =)

PR #1760, opened 2021-12-08
ScannerProgress warning if no files are shared (#1698)

slook:
Proof of concept for keeping the ScannerProgress bar visible if there is a scanning exception error, or if there are 0 shared folders.

image

Just ideas for #1698 the implementation is a rough draft, but the progress bar gets stuck and needs styling as a button with a red bar (that could allow easy access on click to the Shares settings page of Preferences).

Also with aim to improve debug logging for the shares scanner, and caching number of shared folders and files into variables for other procedures to reference.

### DIFF
--- a/pynicotine/shares.py
+++ b/pynicotine/shares.py
@@ -687,15 +687,12 @@ class Shares:
     def send_num_shared_folders_files(self):
         """ Send number publicly shared files to the server. """
 
-        if not (self.core and self.core.logged_in):
-            return
-
         try:
             shared = self.share_dbs.get("files")
             index = self.share_dbs.get("fileindex")
 
             if shared is None or index is None:
-                return
+                return None, None
 
             try:
                 sharedfolders = len(shared)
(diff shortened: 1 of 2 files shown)


pynicotine/shares.py
```diff
@@ -705,10 +702,15 @@ def send_num_shared_folders_files(self):
                 sharedfolders = len(list(shared))
                 sharedfiles = len(list(index))
 
-            self.queue.append(slskmessages.SharedFoldersFiles(sharedfolders, sharedfiles))
+            if self.core and self.core.logged_in:
+                self.queue.append(slskmessages.SharedFoldersFiles(sharedfolders, sharedfiles))
```
slook, line 706: Perhaps it would be better to split this function into count_num_shared_folders_files() and send_num_shared_folders_files()

slook:
Closing, this is not the right solution to resolve #1698 

We need to keep track of Scanner errors in a better way for the purpose of raising an appropriate notification.

Issue #1782, opened 2021-12-29
Allow uploading multiple files to a user at once until limits are reached


toofar:
### Describe the solution you'd like
Currently only one file is uploaded to a user at a time. Consider the following case:

1. I have four open slots configured
2. I have one user who has requested several files for upload
3. We only upload one file at a time so most of my open slots are unused

Maybe maxing out the configured slots with uploads to one user isn't a desirable behaviour because of the assumption that bandwidth between the two of us would already be fully consumed by one upload. But I don't know if that is true. So here I am proposing the other behaviour to see what other people's opinions are (and because I already wrote most of the code before thinking of that downside).

I suppose I was thinking of slots being filled by "files being uploaded" instead of "users being uploaded to".

### Additional context
If it turns out the current behaviour is on purpose I have a branch that changes it to allow uploading multiple files to a user: https://github.com/nicotine-plus/nicotine-plus/compare/master...toofar:feat/upload_candidates_tests_and_tweaks?expand=1

And the first and third commits on that branch may be of interest in any case. The first adds some tests for `get_upload_candidate()` (although they could be changed to be less biased toward my proposed behaviour) and the third has a micro-optimisation for the new round robin code.

mathiascode:
See https://github.com/nicotine-plus/nicotine-plus/discussions/1772

As long as concurrent uploads to a user aren't officially supported in the protocol and official clients, I won't support them in Nicotine+.

slook:
Clearly a lot of thought has been put into @toofar 's suggested enhancements to the upload queue system, and I get the general sense that the current mechanism seems to be inadequate, so improvements would be desirable here.

My main issue with it is that the options are non-sensical. For example, why is it that ["transfers"]["uploadbandwidth"] and ["transfers"]["uploadslots"] cannot be set at the same time? Is it because they conflict with each other or is it because one option simply overrides the other due to the way it's coded? For a start, I propose depreciating the ["transfers"]["useupslots"] option, and allowing a value of "0" to set 'unlimited' or "100000" rather than "1" (unless this isn't allowed on the server?).

Idk if @toofar 's code and test unit can hopefully be re-purposed to provide the improved functionality in compliance with the slsk protocol specifications?

toofar:
> I'm testing concurrent transfers to a single user at the moment and am finding that Qt and Nicotine+ handle them just fine while NS does not

Ah, that's a pity. I didn't think about other clients making assumptions :slightly_smiling_face: 

[...]

PR #1783, opened 2021-12-29
transfers.py: Remove 'useupslots' toggle config option

slook:
- Removed: "Limit number of upload slots" toggle option, because these options wouldn't seem to conflict?*
- Changed: Some local variable names to match the names of the config option names, to make the code easier to follow.

*ToDo: Needs testing, the consequences of allowing both ["transfers"]["uploadbandwidth"] and ["transfers"]["uploadslots"] together with each other are unknown (perhaps there is a good reason for only allowing one or the other?). See #1782 

*ToDo: If testing is successful, then `useupslots` needs properly deprecating and migration code for remove old option in config.py , and the checkbox needs changing to a label in settings/uploads.ui and switching sanity logic in preferences.py

### DIFF
--- a/pynicotine/transfers.py
+++ b/pynicotine/transfers.py
@@ -449,20 +449,20 @@ class Transfers:
 
     def slot_limit_reached(self):
 
-        maxupslots = self.config.sections["transfers"]["uploadslots"]
+        uploadslots = self.config.sections["transfers"]["uploadslots"]
 
-        if maxupslots <= 0:
-            maxupslots = 1
+        if not uploadslots:
+            return False
 
         in_progress_count = 0
-        now = time.time()
+        current_time = time.time()
 
         for i in self.uploads:
             if i.conn is not None and i.speed is not None:
                 # Currently transferring
                 in_progress_count += 1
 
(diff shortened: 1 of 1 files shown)


mathiascode:
> (perhaps there is a good reason for only allowing one or the other?)

They are mutually exclusive. If useupslots is set, you always have a fixed number of upload slots. If not, you'll be allocated new slots as long as the bandwidth limit isn't reached.

The preferences are unclear, but I haven't invested time in improving the UI yet.

slook:
Thanks, it was tricky to work out which option sets the Upload Slots reported by the server (for view User Info etc).

> They are mutually exclusive

Then it should be a pair of radio button options.

> I haven't invested time in improving the UI yet.

Frankly, I don't know if there is really much value in the ["transfers"]["uploadbandwidth"] ("Queue uploads if transfer speeds reaches") option either, since this seems to be confusing to the point where it is difficult to see how all these various queuing and limiting features are supposed to work alongside each other. The more I look at this, the more I struggle to see how the UI could be arranged to circumvent this confusing plethora of overlapping options.

PR #1784, opened 2021-12-29
Add basic tests for get_upload_candidate

toofar:
Add basic tests for get_upload_candidate, and some tidy up.

"basic" in particular because round robin mode has some dependency on `user_update_times` which these tests aren't doing anything with. And that is updated when connections are closed and file transfers are denied or time out which are a bit out of scope of this test.
The test setup and helpers are a bit verbose, I think that is because transfers.py is not written to be particularly testable. Might just be because I ran out of brain power. Hopefully the tests themselves are clear enough though so you only have to read through that stuff once.

### DIFF
--- a/pynicotine/transfers.py
+++ b/pynicotine/transfers.py
@@ -2088,7 +2088,7 @@ class Transfers:
         round_robin_queue = not self.config.sections["transfers"]["fifoqueue"]
         privileged_queue = False
 
-        queued_transfers = OrderedDict()
+        first_queued_transfers = OrderedDict()
         queued_users = {}
         uploading_users = set()
 
@@ -2096,11 +2096,8 @@ class Transfers:
             if i.status == "Queued":
                 user = i.user
 
-                if user not in uploading_users:
-                    if user not in queued_transfers:
-                        queued_transfers[user] = []
-
-                    queued_transfers[user].append(i)
+                if user not in first_queued_transfers and user not in uploading_users:
+                    first_queued_transfers[user] = i
 
                 if user in queued_users:
                     continue
@@ -2117,19 +2114,23 @@ class Transfers:
 
                 uploading_users.add(user)
 
-                if user in queued_transfers:
-                    del queued_transfers[user]
+                if user in first_queued_transfers:
+                    del first_queued_transfers[user]
 
         oldest_time = None
         target_user = None
 
         if not round_robin_queue:
-            for user in queued_transfers:
+            # skip the looping below (except the cleanup) and get the first
+            # user of the highest priority we saw above
+            for user in first_queued_transfers:
                 if not privileged_queue or (privileged_queue and queued_users[user]):
                     target_user = user
                     break
 
         for user, update_time in list(self.user_update_times.items()):
+            # some cleanup, should probably log a warning as this case is
+            # probably a logic bug
             if user not in queued_users:
                 del self.user_update_times[user]
                 continue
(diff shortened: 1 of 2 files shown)


pynicotine/transfers.py
```diff
@@ -2117,43 +2113,46 @@ def get_upload_candidate(self):
...
+                    continue
+
+                target_user = user
+                break
 
         for user, update_time in list(self.user_update_times.items()):
+            # some cleanup, should probably log a warning as this case is
+            # probably a logic bug
             if user not in queued_users:
                 del self.user_update_times[user]
```
mathiascode, line 2136: We should also update the user time in the abort_transfer function (right before the call to check_upload_queue), otherwise you can end up with stuck queued uploads if you abort active uploads in a certain way. Aborting an upload should pass the torch to the next user in line, not start another upload for the same user.

mathiascode: With your changes, I believe a user won't be added to queued_users if they have queued uploads and the user's first transfer in the list is an active one.

toofar: > We should also update the user time in the abort_transfer function

If you say so, pushed.

> a user won't be added to queued_users if they have queued uploads and the user's first transfer in the list is an active one

So they will get cleared from user_update_times even though they still have uploads queued? I see what you mean, it should still work out if they get added back in when their current transfer ends but we maybe aren't updating it everywhere consistently.
I've changed it back. It changed the tests to show that if there are privileged users queued then non-privileged users won't get any slots even if the privileged users couldn't actually use them because they already have one active. I guess that is expected, if less than ideal.

mathiascode:
The FIFO queue doesn't seem to work as intended, or am I misunderstanding the tests?

toofar:
I don't know, what do you see the intention as being? It picks the first upload from a user that doesn't already have an active upload. That's about all it can do. Or would you want it to only pick uploads for the first user and just wait until their active one had finished?

Not sure why the windows tests are failing. Possibly it is flaky? Which wouldn't be good, I would assume it was the timing element of finding the oldest item. Not sure how to reproduce it though, can you tell the CI to re-run?

slook:
> would you want [FIFO] to only pick uploads for the first user and just wait until their active one had finished?

Yes, that sounds partially right. The tooltip states "First In, First Out: Files will be uploaded in the order they were queued." so ideally it would iterate through an ordered list, otherwise we would be assuming that the user queued all their files in one go, which may not be case. Idk what the existing functionality does, tbh I can't tell much difference between the two methods, it seems like it's basically a lottery either way!

toofar:
> assuming that the users queued all their files in one go, which may not be case

You mean if user one queues up one album then comes back and queues a second one, but meantime two other users have queued stuff, the second album should be lumped in together with the first one in terms of priority?

I see what you mean. That is indeed how it works. I was thinking that the method I put tests around worked in terms of users, and it kind does but it gets those users by going through the list of queued _files_, in order, and getting users from there. So my description was wrong. It goes through the queued files in order until it finds the first one from a user who doesn't have an active upload.

Maybe the confusion is due to the tests only marking the currently active upload as finished _after_ picking the next candidate. So while there are multiple users queued you should never see two candidate uploads for the same user in a row. Meaning even in fifo mode it'll jump between users a bit. While this might make the "ideal" queuing behaviour less visible in the tests I think this is the more common case. The alternative would only happen if we someone finished every transfer before going an looking for the next one. I guess if you only had one slot available it would look like that.

I've added two more test cases to hopefully illustrate those cases. 1) "returning user" showing that the priority of a users's latter queued uploads isn't confused with that of their earlier queued uploads 2) the "no contention" case which make the ideal queuing behaviour more clear but would only happen in practice if you only had one slot open.

slook:
Nice find!

This makes me wonder if there are other similar timing issues affected by this, that may explain laggy performance issues on some systems and whatnot. It is useful to know about this.

mathiascode:
My bad, I was thinking about a situation where you have a single upload slot.

Thank you for creating the tests!

PR #1788, opened 2021-12-30
search.py: Improve logging for search requests

slook:
+ Changed: String "returning" -> "found" because the connection attempt could fail to actually return the result to the user
+ Changed: Don't do len(fileinfos) if we already know it is the same length as numresults
+ Added: Debug log for ValueError in create_search_result_list()


pynicotine/search.py
```diff
@@ -532,17 +532,10 @@ def process_search_request(self, searchterm, user, searchid, direct=False):
...
-                })
-        else:
-            log.add_search(
-                _("User %(user)s is searching for \"%(query)s\", returning %(num)i results"), {
-                    'user': user,
-                    'query': searchterm_old,
-                    'num': numresults
-                })
+        log.add_search(
+            _("User %(user)s is %(mode)s for \"%(query)s\", found %(num)i results"), {
```
mathiascode, line 536: I would prefer to keep two separate strings. I'm not sure if the change will work in every language.

slook:
A translated string is changed in this PR, so I suggest that it is not merged until after the next Stable release.

pynicotine/search.py
```diff
@@ -532,16 +529,14 @@ def process_search_request(self, searchterm, user, searchid, direct=False):
         self.core.send_message_to_peer(user, message)
 
         if direct:
-            log.add_search(
-                _("User %(user)s is directly searching for \"%(query)s\", returning %(num)i results"), {
-                    'user': user,
-                    'query': searchterm_old,
-                    'num': numresults
-                })
+            log.add_search(_("User %(user)s is directly searching for \"%(query)s\", found %(num)i results"), {
```
mathiascode, line 532: I'd like the log message to mention what's actually happening, e.g. "attempting to send x results".

slook: What would be really ideal, is to enter the log in slskproto such that we know whether the results were successfully sent or not (depending upon if the associated connection was established etc), but the logic for tracking the message throughout the course of the connection attempt is probably highly complex.

The pluginsystem has a similar limitation, whereby the incoming search notification is generated too soon to know if there was a hit or a miss, nor how many results were found/sent.

PR #1789, opened 2021-12-31
transfers.py: slot_limit_reached() Performance

slook:
+ Added: As soon as `slot_limit_reached` is `True`, then break out of the loop, instead of iterating the entire `uploads` list
+ Changed: Iterate the uploads list in reverse, because active transfers are likely to be the most recent items
+ Changed: Local variable name '`now`' -> '`current_time`' in accordance with convention
+ Changed: Local variable name '`maxupslots`' -> '`uploadslots`' to match the name in config, it might as well be the same

- ToDo: Checking for `last_status_change` is slow, so it would be better to avoid doing this if possible

### DIFF
--- a/pynicotine/transfers.py
+++ b/pynicotine/transfers.py
@@ -449,10 +449,10 @@ class Transfers:
 
     def slot_limit_reached(self):
 
-        maxupslots = self.config.sections["transfers"]["uploadslots"]
+        uploadslots = self.config.sections["transfers"]["uploadslots"]
 
-        if maxupslots <= 0:
-            maxupslots = 1
+        if uploadslots <= 0:
+            uploadslots = 1
 
         in_progress_count = 0
         current_time = time.time()
@@ -470,7 +470,7 @@ class Transfers:
                         or i.status == "Getting status"):
                     in_progress_count += 1
 
-            if in_progress_count >= maxupslots:
+            if in_progress_count >= uploadslots:
                 # slot_limit_reached, so stop iterating
                 return True
 
--- a/pynicotine/transfers.py
+++ b/pynicotine/transfers.py
@@ -455,14 +455,14 @@ class Transfers:
             maxupslots = 1
 
         in_progress_count = 0
-        now = time.time()
+        current_time = time.time()
 
-        for i in self.uploads:
+        for i in reversed(self.uploads):
             if i.conn is not None and i.speed is not None:
                 # Currently transferring
                 in_progress_count += 1
 
-            elif (now - i.last_status_change) < 30:
+            elif (current_time - i.last_status_change) < 30:  # ToDo: performance bottleneck
                 # Transfer initiating, changed within last 30 seconds
 
                 if (i.req is not None
@@ -470,7 +470,11 @@ class Transfers:
                         or i.status == "Getting status"):
                     in_progress_count += 1
 
-        return in_progress_count >= maxupslots
+            if in_progress_count >= maxupslots:
+                # slot_limit_reached, so stop iterating
+                return True
+
+        return False
 
     def bandwidth_limit_reached(self):
 


pynicotine/transfers.py
```diff
@@ -455,22 +455,26 @@ def slot_limit_reached(self):
...
 
-        for i in self.uploads:
+        for i in reversed(self.uploads):
             if i.conn is not None and i.speed is not None:
                 # Currently transferring
                 in_progress_count += 1
 
-            elif (now - i.last_status_change) < 30:
+            elif (current_time - i.last_status_change) < 30:  # ToDo: performance bottleneck
                 # Transfer initiating, changed within last 30 seconds
```
slook, line 466: When there are over 2000 historic uploads to iterate, testing this condition can become very resource intensive when there are many queued transfers and a low number of slots is set. Breaking out of the loop seem to work around this, but maybe it is redundant to check this arbitrary state like this, maybe the next conditions are faster so they can be processed first?

pynicotine/transfers.py
```diff
@@ -449,28 +449,32 @@ def file_limit_reached(self, user):
...
 
                 if (i.req is not None
                         or i.conn is not None and i.speed is None
                         or i.status == "Getting status"):
```
slook, line 470: When there are a large number of historic uploads to iterate, testing this condition can become very resource intensive when there are many queued transfers and a low number of slots is set. Breaking out of the loop seems to work around this, but maybe it is redundant to check this arbitrary state like this, maybe the second set of conditions are faster so they can be processed first?

mathiascode: Most of this seems redundant now, I think we can simply check the status messages.
`if i.status in ("Getting status", "Establishing connection"):`

And the top one:
`if i.status == "Transferring":`

mathiascode:
"Changed: Iterate the uploads list in reverse, because active transfers are likely to be the most recent items"

Recent items are currently appended to the left for performance reasons, so the code was already doing this.

pynicotine/transfers.py
```diff
@@ -449,28 +449,32 @@ def file_limit_reached(self, user):
...
+        current_time = time.time()
 
-        for i in self.uploads:
+        for i in reversed(self.uploads):
             if i.conn is not None and i.speed is not None:
                 # Currently transferring
                 in_progress_count += 1
 
-            elif (now - i.last_status_change) < 30:
+            elif (current_time - i.last_status_change) < 30:  # ToDo: performance bottleneck
```
mathiascode, line 465: I believe this time check is redundant today, because transfers time out after 30 seconds, and I don't think we have issues with stuck transfer statuses anymore.

mathiascode:
Implemented some changes in https://github.com/nicotine-plus/nicotine-plus/commit/0748c34748f759c592ad78f1c80ce3a4efe78f53. Let me know if anything broke.

slook:
Thanks, you have done a thorough job @mathiascode seems to work much more smoothly, I will keep an eye on it.

Since the limiter performance bottlenecks seem to be gone now, this further highlights the performance issues happening in transferlist.py:

- `update_parent_rows (pynicotine/gtkgui/transferlist.py:373)` <-- when tab visible CPU 100% !!!
`if isinstance(useriter, Gtk.TreeIter):`

- `update_bandwidth (pynicotine/gtkgui/transferlist.py:306)` <-- when tab not visible
`for i in self.transfer_list:`

It good that py-spy now shows the problem more clearly, but idk what sort of optimizations are needed there.

mathiascode:
Can you show to full py-spy output?

slook:
image

slook:
Surely shouldn't it be `time.time()` ?

https://github.com/nicotine-plus/nicotine-plus/blob/2535643e45343f93a99a51c6675b16ddede443aa/pynicotine/gtkgui/transferlist.py#L324
https://github.com/nicotine-plus/nicotine-plus/blob/2535643e45343f93a99a51c6675b16ddede443aa/pynicotine/gtkgui/transferlist.py#L390

mathiascode:
Should be improved in https://github.com/nicotine-plus/nicotine-plus/commit/d387bc046cf0ff0707fc575e8ea6639b7483a909

slook:
Nope. update_parent_rows has disappeared from py-spy but there is still 100% CPU usage when the uploads tab is visible:

image

slook:
Since https://github.com/nicotine-plus/nicotine-plus/commit/7423875b9e0947a0954864f16e39ba9315b59d74 still unresolved, (text completion turned off):

image

mathiascode:
> update_parent_rows has disappeared

Well, in that case that issue was solved. As we discussed earlier, most of the CPU usage seems to be somewhere in GTK, and won't necessarily appear in the py-spy output.

slook:
GTK 3.22.11 ... Indeed that does seem to be the case. I can't test this scenario on a different version of GTK to verify, because I don't have a profile with a large upload list, unless perhaps I can copy it over.

PR #1794, opened 2022-01-02
transfers.py: Improve logging for loading, queueing and aborting

slook:
+ Added: Log successful loading of stored transfers json, useful for analyzing startup sequence
+ Added: Log attempt to json store list, useful for analyzing runtime update
+ Added: Log number of paused/aborted transfers, useful for analyzing duplication of events during exit sequence
+ Added: Log reason for aborted transfers, useful for analyzing stuck and zombie file transfers
+ Fixed: Don't try to abort "User logged off" transfers during disconnect/exit, otherwise downloads are aborted twice on exit

- Changed: Syntax of log.add_transfer(...), simplify bracketing and reduce line continuations
- Changed: Log strings (mostly non-translatable ones), convert some jargon to human readable
- Changed: Log string format mapped variable names such as 'request' -> 'token'
- Added: Log 'reason' for aborted, queued/denied transfers, or "Allowed" instead of "1"
- Added: User visible [Downloads] log translatable warning for 'File already downloaded' message
- Removed: Unindent nested condition so to not log queued upload if already queued, and avoid unnecessary virtual2real call
- Removed: Don't show started download INCOMPLETE pathname in [Downloads] log/status bar, use base_name instead
- ToDo: Add INCOMPLETE pathname message somewhere in [Transfers] log
- ToDo: Investigate if "Queued" is triggered by "Queued (prioritized)" and "Queued (privileged)" or if need startswith("Queued")

This improved logging has highlighted issues with the exit and aborting sequence, and resuming active transfers across sessions are sometimes Denied due to "File not shared" and "User logged off" statuses, but those bugs are beyond the scope of this PR so I will remove the # comments after review, if the logging is deemed to be useful and worth keeping even after debugging.


mathiascode:
Are there any important fixes/observations that should be addressed before 3.2.1?

slook:
No, this logging, and additional logging/profiling done locally, has so far only highlighted non-urgent issues/optimizations, nothing that can't wait. For example...

https://github.com/nicotine-plus/nicotine-plus/blob/043768feea7c93cfd28c7e6e48ad3a8fb6adc66d/pynicotine/gtkgui/transferlist.py#L322-L326

... results in each json being written twice (4 writes in total) because update() is called twice:
```
12:49:32 [Transfer] Stored list of 1675 uploads to /home/user/.nicotine/uploads.json
12:49:32 [Transfer] Stored list of 201 downloads to /home/user/.nicotine/downloads.json
12:49:32 [Transfer] Stored list of 1675 uploads to /home/user/.nicotine/uploads.json
12:49:32 [Transfer] Stored list of 201 downloads to /home/user/.nicotine/downloads.json
```
Might as well use `self.page_id` , this halves the number of json dumps...
```
                self.frame.np.transfers.save_transfers(self.page_id)
```

... but this doesn't have a large impact on poor performance of the transfers UI, which I have verified is mainly caused by the sheer quantity of value iterations in update_specific(): Ideally, we could perhaps store and then directly address the relevant Gtk.Tree row object '<0x0000...>' references in order to fix that, but this will require extreme changes to the model implementation.

_Observations about other minor issues highlighted by the improved logging:_

- "Queued (prioritized)" ; "Queued (privileged)" statuses don't seem to have any affect on the order (or it may even cause lower priority), but this needs more testing. Somewhere we need something like `status == startswith("Queued")` or expand the scope of the queue status `modifier`, or some other solution.

- "User logged off" status might not be handled correctly, the peer might want this status passed as a 'reason' at the other end if there is a manual disconnect, or maybe not... it is not very clear if this is supposed to refer to local status or remote status, and confusion exists at several places within the logic regarding this. Other clients are likely to have related bugs also.

- Transfers is unaware of the shares Scanner status (since there is no 'self.initialized' flag for that module). This ignorance affects previously Queued uploads during startup, since check_user() is called too early (or it should return some other status). The upshot is transfers which become stuck on "File not shared", and are therefore never automatically retried in some cases.

- Statuses exist which don't seem to be implemented, such as "Pending shutdown", but it is not clear how this is intended to be used. Further investigation of what other clients do (and expect from Nicotine+) during their exit/disconnection sequence is needed to correct some of the erroneous behaviours that arise when inter-operating with other clients.

While the above points are noteworthy, it is not considered important nor urgent.

mathiascode:
> ... results in each json being written twice (4 writes in total) because update() is called twice:
> 
> ```
> 12:49:32 [Transfer] Stored list of 1675 uploads to /home/user/.nicotine/uploads.json
> 12:49:32 [Transfer] Stored list of 201 downloads to /home/user/.nicotine/downloads.json
> 12:49:32 [Transfer] Stored list of 1675 uploads to /home/user/.nicotine/uploads.json
> 12:49:32 [Transfer] Stored list of 201 downloads to /home/user/.nicotine/downloads.json
> ```
> 
> Might as well use `self.page_id` , this halves the number of json dumps...
> 
> ```
>                 self.frame.np.transfers.save_transfers(self.page_id)
> ```

Good observation, this could definitely be improved, since dumping to json can be somewhat slow with large transfer lists.

>     * "Queued (prioritized)" ; "Queued (privileged)" statuses don't seem to have any affect on the order (or it may even cause lower priority), but this needs more testing. Somewhere we need something like `status == startswith("Queued")` or expand the scope of the queue status `modifier`, or some other solution.

What does "order" mean in this case? The modifier is never part of the status (simply appended on the GUI side), and prioritized/privileged transfers should work as intended, at least according to our unit tests.

>     * "User logged off" status might not be handled correctly, the peer might want this status passed as a 'reason' at the other end if there is a manual disconnect, or maybe not... it is not very clear if this is supposed to refer to local status or remote status, and confusion exists at several places within the logic regarding this. Other clients are likely to have related bugs also.

`User logged off` is strictly a local status, it should never be sent to another Soulseek client. `Cancelled` is the main status SoulseekQt sends in almost all cases, so we do it as well.

>     * Transfers is unaware of the shares Scanner status (since there is no 'self.initialized' flag for that module). This ignorance affects previously Queued uploads during startup, since check_user() is called too early (or it should return some other status). The upshot is transfers which become stuck on "File not shared", and are therefore never automatically retried in some cases.

I've known about this issue for a while, but I'm still not sure what to do about it. I believe SoulseekQt blocks the GUI/connecting to the server until the rescan has finished, but this is the exact opposite of what we want to do (fast startups).

>     * Statuses exist which don't seem to be implemented, such as "Pending shutdown", but it is not clear how this is intended to be used. Further investigation of what other clients do (and expect from Nicotine+) during their exit/disconnection sequence is needed to correct some of the erroneous behaviours that arise when inter-operating with other clients.

This status is used by SoulseekQt for https://github.com/nicotine-plus/nicotine-plus/issues/885, and is sent if you try to queue new uploads while the feature is enabled.

slook:
> Good observation, this could definitely be improved, since dumping to json can be somewhat slow with large transfer lists.

Actually on application exit there can be up to 8 json dumps, because both uploads and downloads are called in both the quit() and disconnect() functions.

> What does "order" mean in this case? The modifier is never part of the status...

I have noticed uploads marked as "Queued (prioritized)" to buddies not being chosen over ordinary users, particularly at startup so this might be due to the Scanner bug...

> I'm still not sure what to do about it... ...what we want to do (fast startups).

Blocking the transfers from starting until the Scanner has fully initialized (ie, either completed scanning shares or loaded the existing database depending upon startup preference), would both solve the failed upload attempts and improve startup performance. Perhaps a `self.initialized` flag should be set to `False` on init of Scanner or Shares, and then set to `True` after the threaded Scanner has finished, then this could provide a signal to the Transfers module that it is safe to update statuses, or so that it can begin loading stored transfers only after avoiding the race condition.

> Cancelled is the main status SoulseekQt sends in almost all cases, so we do it as well.

We don't send any message on shutdown/exit/disconnect, because `abort_transfer(i)` is never called, only removes the TreeView entry from our own uploadsview:

https://github.com/nicotine-plus/nicotine-plus/blob/110b775994a11f7da99f3a26a2646f5266f81c6d/pynicotine/transfers.py#L2380-L2385

This is evidenced by the fact that the improved logging shows that multiple attempts to abort uploads are made upon exit. It is probable that any remote client gets "Cannot connect" or whatever other reason is determined locally at the other end.

mathiascode:
> > I'm still not sure what to do about it... ...what we want to do (fast startups).
> 
> Blocking the transfers from starting until the Scanner has fully initialized (ie, either completed scanning shares or loaded the existing database depending upon startup preference), would both solve the failed upload attempts and improve startup performance. Perhaps a `self.initialized` flag should be set to `False` on init of Scanner or Shares, and then set to `True` after the threaded Scanner has finished, then this could provide a signal to the Transfers module that it is safe to update statuses, or so that it can begin loading stored transfers only after avoiding the race condition.

The problem is that other users will resume their downloads as soon as we appear online on the server, and they expect an answer right away (allowed/denied).

> > Cancelled is the main status SoulseekQt sends in almost all cases, so we do it as well.
> 
> We don't send any message on shutdown/exit/disconnect, because `abort_transfer(i)` is never called, only removes the TreeView entry from our own uploadsview:
> 
> https://github.com/nicotine-plus/nicotine-plus/blob/110b775994a11f7da99f3a26a2646f5266f81c6d/pynicotine/transfers.py#L2380-L2385
> 
> This is evidenced by the fact that the improved logging shows that multiple attempts to abort uploads are made upon exit. It is probable that any remote client gets "Cannot connect" or whatever other reason is determined locally at the other end.

I think I made this decision because we're not supposed to abort the transfer when logging off. Otherwise, the download may appear "Cancelled" on their end, and not resume anymore when we reconnect. On shutdown, there wouldn't be any time to send a message anyway.

The other client is supposed to watch our online status, notice when we've gone offline, and make changes to the transfers based on that.

slook:
> The problem is that other users will resume their downloads as soon as we appear online on the server, and they expect an answer right away (allowed/denied). Perhaps we could send a "Queued" reason or something similar here, I'm not sure.

At the moment, we send "Denied" (0). Perhaps in the case where the proper response is unknown (such as during Scanner initialization), it would be better to send nothing at all. Then at least the remote client will know to retry after a while.

Queued would also work... if it later turns out the file is unavailable then so be it... Perhaps we can set our max_uploads_slots_reached and max_bandwidths_reached flags to True by default during initialization to achieve this.

> The other client is supposed to watch our online status, notice when we've gone offline, and make changes to the transfers based on that.

That makes sense. A code comment would be useful here to explain why uploads are not aborted the same as downloads.

slook:
This was useful for a time but is obsolete now, since many of the required fixes have been implemented now. Closing.

PR #1812, opened 2022-01-09
remove all trailing whitespace

hboetes:


doc/DEVELOPING.md
```diff
@@ -88,9 +88,9 @@ Tests are defined in the *test/* folder, and should be expanded to cover larger
 Nicotine+ tries to follow [Semantic Versioning](https://semver.org/) when possible. As cited in the specification:
 
 > Given a version number MAJOR.MINOR.PATCH, increment the:
-> 
->   MAJOR version when you make incompatible API changes,  
->   MINOR version when you add functionality in a backwards compatible manner, and  
+>
+>   MAJOR version when you make incompatible API changes,
+>   MINOR version when you add functionality in a backwards compatible manner, and
 >   PATCH version when you make backwards compatible bug fixes.
```
slook, line 94: It looks like this was intended to be a bulleted list, but is has come out as an indented quote.

hboetes: I can live with that, but it also modified some images. That's not good!

hboetes:
it also modified some images. That's not good!

slook:
Oh oops! also don't modify `config` because those whitespaces are intentional

hboetes:
OK, thanks for the heads-up.

PR #1813, opened 2022-01-09
remove almost all trailing whitespace

hboetes:


hboetes:
> my| heads up, markdown uses two trailing spaces for line breaks
> Han| no way! That is sooooooooooo ugly!

PR #1838, opened 2022-01-17
Use significant decimals, rather than one decimal behind the dot.

hboetes:
Current -> Proposed:
1.2 Gb  -> 1.23 Gb

This provides sufficient precision without taking up too much space.


hboetes:
Another advantage, columns look properly sorted. 
image

mathiascode:
I like this change, but we need to handle values between 1000 and 1024 somehow:
Screenshot from 2022-01-18 21-20-09

PR #1841, opened 2022-01-20
Update control

SwampRabbit:
add python3-pytest-xvfb to build depends, needed to build in a Debian pbuilder schroot, may also be needed for Debian sbuild but I haven't tested that

once xvfb-run dh_auto_test is hit, it will fail with xvfb-run: error: xauth command not found

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -15,7 +15,8 @@ Build-Depends:
     python3-gdbm,
     python3-gi (>= 3.18),
     python3-setuptools,
-    xvfb
+    xvfb,
+    python3-pytest-xvfb
 Vcs-Git: https://github.com/nicotine-plus/nicotine-plus.git
 Vcs-browser: https://github.com/nicotine-plus/nicotine-plus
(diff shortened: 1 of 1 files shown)


mathiascode:
Thanks! I suppose we could use pytest and the pytest-xvfb plugin for the tests in the Debian package, and eliminate xvfb-run completely?

SwampRabbit:
> Thanks! I suppose we could use pytest and the pytest-xvfb plugin for the tests in the Debian package, and eliminate xvfb-run completely?

I believe that python3-pytest-xvfb (the Debian package name) uses xvfb-run anyway to fake the GUI tests in chroots.
I'm not super familiar with any of the python testing modules enough to suggest anything.  :(

I see you have builds working on Launchpad, so it must be doing something different from Debian standard pbuilder/sbuild chroot.

SwampRabbit:
@mathiascode I believe this isn't needed anymore now with #1842 so I'm closing it

PR #1856, opened 2022-01-30
SLSKPROTOCOL.md: Improve info about Peer Code 9 (FileSearchResult)

slook:
+ Added: FileSearchResult Send attribute data: bitrate; duration; vbr
+ Added: FileSearchResult Receive attribute data: *Order varies between clients*
+ Changed: Wording around "Iterate for number of [items]", because "size" was ambiguous with filesize

### DIFF
--- a/doc/SLSKPROTOCOL.md
+++ b/doc/SLSKPROTOCOL.md
@@ -2819,7 +2819,6 @@ A peer sends this message when it has a file search match. The token is taken fr
         5.  Iterate for <ins>number of attributes</ins>
             1.  **uint** <ins>place in attributes</ins> *(Order varies between clients and file type)*
             2.  **uint** <ins>attribute</ins>
-                - *Either bitrate or sample rate*
     6.  **bool** <ins>slotfree</ins>
     7.  **uint** <ins>avgspeed</ins>
     8.  **uint64** <ins>queue length</ins>
@@ -2832,7 +2831,6 @@ A peer sends this message when it has a file search match. The token is taken fr
          5.  Iterate for <ins>number of attributes</ins>
              1.  **uint** <ins>place in attributes</ins> *(Order varies between clients and file type)*
              2.  **uint** <ins>attribute</ins>
-                 - *Either bitrate or sample rate*
 
 ### Peer Code 15
 
--- a/doc/SLSKPROTOCOL.md
+++ b/doc/SLSKPROTOCOL.md
@@ -2782,7 +2782,7 @@ A peer sends this message when it has a file search match. The token is taken fr
         1.  **uchar** 1
         2.  **string** <ins>filename</ins>
         3.  **uint64** <ins>size</ins>
-        4.  **string** <ins>ext</ins>
+        4.  **string** <ins>ext</ins> *(SoulseekNS requires "mp3" to show the attributes)*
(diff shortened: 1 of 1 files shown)


doc/SLSKPROTOCOL.md
```diff
@@ -2777,57 +2777,59 @@ A peer sends this message when it has a file search match. The token is taken fr
...
+    9.  **uint** <ins>number of privately shared results</ins>
     10.  Iterate for <ins>number of privately shared results</ins>
          1.  **string** <ins>filename</ins>
          2.  **uint64** <ins>size</ins>
          3.  **string** <ins>ext</ins>
-         4.  **uint** <ins>number of attributes</ins>
-         5.  Iterate <ins>number of attributes</ins>
-             1.  **uint** <ins>place in
-                 attributes</ins>
+         4.  **uint** <ins>number of attributes</ins> 2 *or* 3 *(depending on <ins>ext</ins>)*;  0 *(no metadata)*
```
mathiascode, line 2830: This can actually be 1 in rare cases (sent from SoulseekQt), and I believe the attribute could either be a bitrate or sample rate, but I haven't analyzed how SoulseekQt differentiates these two yet.

doc/SLSKPROTOCOL.md
```diff
@@ -2777,57 +2777,59 @@ A peer sends this message when it has a file search match. The token is taken fr
...
+         5.  Iterate for <ins>number of attributes</ins>
+             1.  **uint** <ins>place in attributes</ins> *(Order varies between clients and <ins>ext</ins>)*
```
mathiascode, line 2832: SoulseekQt doesn't use `ext` anymore, it just sends a blank string (but SoulseekNS requires the value to be "mp3" to show the attributes).

doc/SLSKPROTOCOL.md
slook: How curious, I suppose it is unlikely that a sample rate would be less than 4096, and bitrates are usually less than that.

doc/SLSKPROTOCOL.md
```diff
@@ -2777,58 +2777,62 @@ A peer sends this message when it has a file search match. The token is taken fr
...
-         4.  **uint** <ins>number of attributes</ins>
-         5.  Iterate <ins>number of attributes</ins>
-             1.  **uint** <ins>place in
-                 attributes</ins>
+         3.  **string** <ins>ext</ins> *(Always blank from SoulseekQt clients)*
+         4.  **uint** <ins>number of attributes</ins> *(Usually 2 or 3; can be 1 in rare cases; 0 is no metadata)*
+         5.  Iterate for <ins>number of attributes</ins>
+             1.  **uint** <ins>place in attributes</ins> *(Order varies between clients and file type)*
              2.  **uint** <ins>attribute</ins>
+                 - *Either bitrate or sample rate*
```
mathiascode, line 2835: This doesn't belong here. We should document these attributes at a later point, preferably in a separate section.

slook: Yes I see several sections deal with attributes, and outlining them isn't straight-forward.

My main objective is to indicate what N+ sends, and hope that others eventually feel inspired to follow suit.

PR #1857, opened 2022-01-30
README.md: Clarify Legal and Privacy section

slook:
- Changed: "since" -> "which is not", because it was ambiguous with the Nicotine+ Team used to operate the Soulseek server, since this was never the case.


hboetes:
Yes, better.

PR #1885, opened 2022-02-13
Result Filters: Accept non-integer for size and bitrate, also add not operator !

slook:
+ Added: Not ("!") operator for size and bitrate result filters, because this helps to exclude unwanted files of a particular length
+ Changed: Accept non-integer value, because if the size is factorized then an exact match was practically impossible
+ Added: Approximate to within +/- unit tolerance, a quarter of unit each way, because the user cannot know the exact bytes
+ Added: Run multiple conditions if filter is split with or-pipes "|"
- Changed: Result Filter Help text, @hboetes please can you review these changes, if this makes sense to you?
- Added: Size in bytes (B), in File Properties so it's possible to discover the exact size for use in the Size result filter.

### DIFF
--- a/pynicotine/gtkgui/search.py
+++ b/pynicotine/gtkgui/search.py
@@ -823,7 +823,10 @@ class Search(UserInterface):
                 adjust = factor / 8 if factor > 1024 and sdigit < 104857600 else factor  # TODO: GiB
 
                 if (sdigit - adjust) <= value <= (sdigit + adjust):
-                    return True if used_operator == "==" else False
+                    if used_operator == "!=":
+                        return False
+
+                    return True
 
                 if used_operator == "!=":
                     allowed = True


pynicotine/gtkgui/search.py
```diff
@@ -797,10 +797,19 @@ def check_digit(self, sfilter, value, factorize=True):
...
         try:
-            sfilter = int(sfilter) * factor
+            sfilter = float(sfilter) * factor
         except ValueError:
             return True
 
+        # Exact size match is unlikely, so approximate within +/- unit tolerance
+        if used_operator == "==":
+            return (value >= sfilter - (factor / 2)  # pylint:disable=chained-comparison
+                    and value <= sfilter + (factor / 2))
```
slook, line 807: If there is a more graceful way to do this, then any suggestions are welcome.

PR #1886, opened 2022-02-13
Result Filters: Allow spaces and commas in Country Code filter entry string

slook:
+ Added: In addition to the "|" (or separator), also allow " " (space), "," (comma), "&", "+", "or", "and" in Country filter string
+ Added: In addition to the Country Name tooltip, also display the Country Code in brackets for ease of reference


slook:
If we use RegEx for one filter, we'd have to do the same for other filters. Unwanted complexity. Closing.

PR #1890, opened 2022-02-15
iconnotebook.py: Just set color without parsing

slook:
- Removed: Don't parse color with Gdk in Notebook, because at this point the config value should be correct
- Changed: Import html.escape module wide, because it is called too frequently to be imported on each chat message
+ Added: Parse for valid color when setting colors in Preferences (this might not be exactly the correct place?)

Aims to avoid recursion crash in #1889

### DIFF
--- a/pynicotine/gtkgui/widgets/iconnotebook.py
+++ b/pynicotine/gtkgui/widgets/iconnotebook.py
@@ -21,6 +21,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import sys
+from html import escape
 
 from gi.repository import Gdk
 from gi.repository import Gtk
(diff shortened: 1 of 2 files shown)


pynicotine/gtkgui/widgets/iconnotebook.py
```diff
@@ -141,12 +142,9 @@ def _pack_children(self):
         if sys.platform != "darwin":
             self._add_close_button()
 
-    def _set_text_color(self, color):
+    def _set_text_color(self, color=None):
 
-        color_rgba = Gdk.RGBA()
-
-        if color_rgba.parse(color):
```
mathiascode, line 148: Someone can manually edit the config to include any value here, so some kind of validation is still desirable.

slook: Yes, I'm glad you raised this, we would need our own way of parsing this since we cannot import Gdk into config.

I suppose we need a hexadecimal value in the format #RRGGBB , we could somehow at least check for that even if the color isn't a valid one.

slook:
Superseeded by https://github.com/nicotine-plus/nicotine-plus/commit/4e03848ed7530b6350a0f08617a6a374c1cfe0f3 closes this (probably).

Issue #1887, opened 2022-02-13
Wishlist: take into account search filters


garylaski:
As far as I am aware, the wishlist only takes into account the search text, and executes searches with your default filters. It would be a great feature for the wishlist to be able to save an entire search configuration (filters).

slook:
Yes you are correct, and without the extra filters or other means of discarding unwanted results there is limited value in the wishlist at the moment. There is a partial workaround to this, see https://github.com/nicotine-plus/nicotine-plus/discussions/1604#discussioncomment-1467766 

Before this feature is added, we should add the missing Length filter.

mathiascode:
> missing Users filter

A "Discard repeated results" wishlist feature would essentially do this: https://www.slsknet.org/news/comment/2926#comment-2926

Requested in https://github.com/nicotine-plus/nicotine-plus/discussions/1604

slook:
Interesting, that is quite rudimentary, but probably effective enough at discarding unwanted results. There have been several variations of this over the years, pyNicotine had "Close and Ignore" perhaps that worked in a similar fashion.

While we're at it we might as well store the entire `filters` set, alongside the Wishlist search term. The most accurate method of discarding search results might vary depending on the nature of the desired file.

Probably raw Bytes values would be the best method of differentiating between two files with the same filename.

mathiascode:
> pyNicotine had "Close and Ignore" perhaps that worked in a similar fashion.

To clarify, old Nicotine+ versions had an option to reopen closed search tabs when new results arrived, and the ignore feature was mainly used with this option. Nowadays this doesn't make much sense, since search requests are distributed faster and results fill up instantly in most cases.

slook:
Update: This issue is one step closer since #1727 has been closed with #1892

mathiascode:
Closing in favor of #1604

Bunny-Head:
Sorry for re-open this but, was the "repeated results" applied? If so, I can't find it out, can you point me where?

And btw, thank you so much for still developing and improving Nicotine+!

slook:
No, `discard_repeated` it is not added yet but it is planned see https://github.com/nicotine-plus/nicotine-plus/discussions/1604#discussioncomment-4660210

Issue #1727, opened 2021-11-30
Filter for song length


brutalbombs:
I really miss a feature to filter on song length in both nic+ and slsk as of now. I usually encounter this problem when I'm looking for a special song, but this song is either included in mixes or collections. Usually one could just filter out the false positives by adding "-[collectionname]" to the search string, but at times the song is just renamed and looks like the original file I'm looking for.

This is especially tedious with wish-list searches, as they keep finding the same short 4m30s song when i'm looking for the 6m20s version.

What i would really like is to be able to set a SONG LENGTH filter with operators like >, <, = something akin to this:

1. sl>2m30s
2. sl<5m0s
3. sl=2m33s

This would greatly help when doing some pretty deep digs with the wishlist function.

slook:
I agree @brutalbombs a Length filter is missing, and is very relevant for the file types concerned.

Since the Result Filters bar is rather crowded, this needs figuring out so that the additional Length filter can be added. I note that Speed is also missing (but is less important than Length). This makes me wonder if it could be worth toggling on/off the desired filter types to help with https://github.com/nicotine-plus/nicotine-plus/discussions/1535#discussioncomment-1444130

As to integration of Filters with the Wishlist, this is something that is planned for a future release, and as you pointed out, the old results quickly become annoying and this prevents the feature being used to it's full potential, see https://github.com/nicotine-plus/nicotine-plus/discussions/1604

Recent and ongoing improvements to the Result Filters allow scope for your suggested improvements in the future (after the next Stable release or two), see https://github.com/nicotine-plus/nicotine-plus/issues/1400

> something akin to this:
>    `>2m30s`

Hopefully the parsing of filters will allow for either the `h:mm:ss` or `Hh Mm Ss` syntax for song length.

slook:
@brutalbombs The feature you requested has been added into Search Files (but is not yet integrated for Wishlist searches see open issue #1887 still in progress).

You can use the [latest unstable build](https://github.com/nicotine-plus/nicotine-plus/blob/master/doc/TESTING.md) Nicotine+ 3.3.0.dev1 to test it out. Any feedback would be most welcomed.

"Duration" filter query format can be `HH:MM:SS` or `MM:SS` or `S`

_Multiple conditions can be in any order, for example..._
`>7:00|<11:00|!7:30|!7:33|!6:46|=12:45|=260|=1:58:30`
_... shows:_
- all tracks between 7 and 11 minutes long, except for...
[...]

PR #1892, opened 2022-02-16
Result Filters: Add 'Length' filter

slook:
+ Added: FilterLength ComboEntryBox for Result Filters bar, default filter in Preferences, history config, etc.
+ Added: Not ! operator and split conditions in check_digit() for Length, Size and Bitrate filters, to enable explicit exclusions.
+ Added: Approximate to within +/- filesize tolerance, 0.1 of a unit each way, if the user doesn't enter the exact bytes.
+ Moved: factorize filesize string interpreter code into separate factorize() function in utils.py.
+ Changed: Accept non-integer value, because if the size is factorized then an exact match was practically impossible.
+ Changed: Text strings in Result Filter Usage popover, added 'Length' section.
+ Changed: "Brittany Spears" -> "Aphex Twin", although perhaps a non-copyrighted artist would be better for the example.

image

_Multiple conditions can be in any order, for example..._
`>7:00|<11:00|!7:30|!7:33|!6:46|=12:45|=260|=1:58:30`
_... shows:_
- all tracks between 7 and 11 minutes long, except for...
- hide tracks that are exactly 7:30 or 7:33 or 6:46 long
- and also show songs that are exactly 12:45 long, exactly 260 seconds, and exactly 1h58m30s long.

_For the Length filter, = "exactly" means to within 1 second (decimals are not supported)._
_For the File Size filter, = "exactly" means to within 0.1 MiB (or within 1 MiB if over 100 MiB), unless raw size is specified._

In future, if we add entries programmatically (on_combo_icon_click() or context menu?) then raw Seconds or Filesize are faster to process and dead accurate, therefore more suitable for storing in the Wishlist for the purpose of excluding unwanted results.

This could help us to provide a mechanism to ignore unwanted Wishlist results (see #1887), because any amount of `!` (not) and `=` (exact) conditions are possible, by stringing together !1|!2|!3|... conditions. This would probably be easier and more efficient than creating a database of ignored Wishlist results.

**_Status: Ready for inclusion into master for further testing (feedback welcomed), aims to close feature request https://github.com/nicotine-plus/nicotine-plus/issues/1727_**


mathiascode:
Thanks! I'll review and test this as soon as possible, but finishing 3.2.2 is my current priority with the free time I have.

pynicotine/gtkgui/search.py
```diff
@@ -887,6 +902,9 @@ def check_filter(self, row):
         if filters["filtertype"] and not self.check_file_type(filters["filtertype"], row[11]):
             return False
 
+        if filters["filterlength"] and not self.check_digit(filters["filterlength"], row[16].get_uint64(), False):
```
mathiascode, line 905:
```suggestion
        if filters["filterlength"] and not self.check_digit(filters["filterlength"], row[16].get_uint(), False):
```

pynicotine/gtkgui/ui/search.ui
```diff
@@ -159,14 +159,14 @@
                 <property name="margin-start">12</property>
                 <property name="margin-end">6</property>
                 <child>
-                  <object class="GtkFlowBox">
+                  <object class="GtkFlowBox" id="FiltersFlowBox">
```
mathiascode, line 162: No need for these new ids.

slook: Okay I shall remove them. My future intention is to provide some sort of mechanism for showing/hiding the visibility to allow the filters bar to be customized.

pynicotine/gtkgui/search.py
slook: Agreed it is adequate to allow for files which are up to 68 years long!

mathiascode:
@hboetes Could you check if the strings in this PR sound fine to you?

pynicotine/gtkgui/ui/popovers/searchfilters.ui
```diff
@@ -108,7 +117,7 @@
                         <property name="xalign">0</property>
                         <property name="margin-start">12</property>
                         <property name="label" translatable="yes">For order-insensitive filtering, as well as filtering several exact phrases, vertical bars can be used to separate phrases and words.
-    Example: Spears|Brittany|My beautiful album|hello</property>
+    Example: Twin|Aphex|Analogue Bubblebath|Digeridoo</property>
```
mathiascode, line 120: As you mentioned, this should be replaced with a non-copyrighted example before the PR is merged.

mathiascode:
Merging for further testing. Follow-up PRs can be created if necessary.

slook:
@redactedscribe Please can you test the new functionality detailed above https://github.com/nicotine-plus/nicotine-plus/pull/1892#issue-1139921545 in the [latest unstable build](https://github.com/nicotine-plus/nicotine-plus/blob/master/doc/TESTING.md) of Nicotine 3.3.0.dev1 and provide any feedback you may have about it. Thanks.

slook:
1. Slightly more useful defaults added in #2238 as also suggested in #2141.
2. Adding support for the 1h30m syntax would be desirable, but there are complexity where spaces might be used in the condition (such as 1h 30m), so I shall not be implementing another syntax definition until orphaned parts of the condition strings are properly handled and/or we have a mechanism of providing feedback about syntax errors and/or a method that is able to interpret sub-parameters that contain spaces.
3. Good point, although for now I think excluding results with no time is better done manually if the user has explicitly entered a !0 condition, because a filter may reduce the results down to nothing but it is not obvious whether or not this is accurate or if it is due to some syntax error.
4. The future intention is to implement an `is_filter_valid` flag or similar that would ideally notfiy exactly which condition has failed, or at least indicate that a failure does exist somewhere within the filter entry. The current implmentation of making the in/out TextEntry boxes background red is slightly hacky, so I would like to implement something more suitable for this purpose.

PR #1902, opened 2022-02-18
Remove 'Clear Finished' buttons

slook:
- Removed: 'Clear Finished' button, because it is undesirable since the transfer lists are more permanent than they used to be

The Clear Finished button only duplicates a function available in the adjacent button menu, so it serves little purpose.


pynicotine/gtkgui/ui/uploads.ui
```diff
@@ -176,16 +148,16 @@
                     <property name="visible">True</property>
                     <property name="spacing">6</property>
                     <child>
-                      <object class="GtkLabel">
+                      <object class="GtkImage">
                         <property name="visible">True</property>
-                        <property name="label" translatable="yes">Clear Uploa_ds…</property>
```
mathiascode, line 181: I'm fine with giving the button removal a go, but I'd prefer if we kept the arrow on menu buttons for consistency.

slook: It looks a bit too funky with both the Eject icon and the Pan-Up icon, so I have reverted the button to how it was before, with the Pan-Up icon on the right side as is standard for a menu button.

PR #1904, opened 2022-02-18
Private Chat: Add Text-to-Speech toggle button

slook:
+ Added: Text-to-Speech toggle button in Private Chat, functions the same as Public Chats

image


pynicotine/gtkgui/privatechat.py
```diff
@@ -342,9 +351,11 @@ def message_user(self, msg):
         if text.startswith("/me "):
             line = "* %s %s" % (self.user, text[4:])
             tag = self.tag_action
+            speech = text[4:]
```
mathiascode, line 354: This should probably be line[2:] instead, similar to chat rooms. Sometimes the username is a part of the sentence in /me messages.

PR #1905, opened 2022-02-18
declunkify

hboetes:


mathiascode:
Could you make the same changes to the strings in uploads.ui as well?

hboetes:
> Could you make the same changes to the strings in uploads.ui as well?

Yes! They are surprisingly similar. :-)

slook:
I don't think we need two buttons, the menu one is enough imo. See https://github.com/nicotine-plus/nicotine-plus/pull/1902

mathiascode:
We can continue the discussion in #1902

PR #1909, opened 2022-02-19
Add link to the actual quote.

hboetes:


README.md
```diff
@@ -38,7 +38,7 @@ If you feel like contributing to Nicotine+, there are several ways to help:
 # Where did the name Nicotine come from?
 
 > I was in a geeky mood and was browsing [http://www.bash.org](http://www.bash.org)'s QDB.  
-I stumbled across this quote:  
+I stumbled across [this quote](http://www.bash.org/?10159):
```
slook, line 41: Instead of this additional link, the link in the line above might as well point to the specific page, instead of linking to bash.org's home page. Also the statement can be combined onto one line...

```
> I was in a geeky mood and was browsing [bash.org](http://www.bash.org/?10159)'s QDB,  I stumbled across this quote:
```

> I was in a geeky mood and was browsing [bash.org](http://www.bash.org/?10159)'s QDB,  I stumbled across this quote:

README.md
```diff
@@ -38,7 +38,7 @@ If you feel like contributing to Nicotine+, there are several ways to help:
 # Where did the name Nicotine come from?
 
 > I was in a geeky mood and was browsing [http://www.bash.org](http://www.bash.org)'s QDB.  
-I stumbled across this quote:  
+I stumbled across [this quote](http://www.bash.org/?10159):  
 >> **\<etc>** so tempting to release a product called 'nicotine' and wait for the patches.  
 >> **\<etc>** then i would have a reason to only apply one patch a day. otherwise, i'm going against medical advise.
```
slook, line 43: The original quote has a spelling error `advise` - > `advice` , nobody will notice if it's different so we might as well change it to read correctly.

hboetes: It's an IRC quote. They are unedited by nature.

README.md
slook: Yes and that's good, I'm just saying the existing hyperlink should be changed instead of creating a new one.

PR #1912, opened 2022-02-20
Private Chat: Remember per-user 'Log' toggle setting

slook:
+ Added: `["logging"]["private_users"]` config setting to remember 'Log' toggle, functions the same as ["rooms"]
+ Changed: Add user to users before watching them, a status notification could be missed if the response is very fast

Obviously the 'Log private chat by default' option takes precedence, and makes this list unnecessary, but not storing the users causes a confusion if that option then gets turned off.

In future it could be worth implementing some conditions to handle this better, such that a list of users is only stored in config if required, but for now this shall be adequate.


mathiascode:
I'm converting this to a draft until I have time to look closer into what to do with logging in general. I'd rather not introduce another config option yet that will possibly be migrated to another format anyway.

slook:
Sure, the purpose of this PR was only to explore an idea for one possible method of implementation in the hope that it might be useful, if only for inspiration. 

It is not my expectation that anything be merged as is unless it's obvious that the offered solution is concrete, and in this case there are clearly issues with conflicting default settings that require further consideration.

mathiascode:
My current plan:
- Add two config option dictionaries for rooms and private chats. A room/user is the key, a boolean value is used as the value to indicate if logging is enabled.
- Add buttons in Preferences -> Logging to reset these dictionaries

Thoughts?

PR #1914, opened 2022-02-22
Chat Rooms: Check username mention is not a substring

### LINKED ISSUE #1790: Chat mentions with 1 character username
Nicotine+ version:  
Operating System/Distribution: Windows 10 21H2

### Describe the bug
If the message contains my username, in this case the character `b` I get mentioned in chat.

### Expected behavior
Only get mentioned when someone actually sends `b` separately or `@b`

Send a message with a word that contains `b` in some of the channels I'm in

### Additional context
image

slook:
Fixes #1790

+ Added: Check that a username mention is not a substring, and don't do find() in text unless it contains a possible mention
+ Moved: Get tag type (action, local, hilite or remote) into separate method, so it can be used elsewhere eg. append_log_lines()
+ Changed: Show tab hilite before generating mentioned notification popup, to avoid a slightly laggy UI stutter upon mention
+ Changed: Append to room TextView before writing to log file, this appears to feel more responsive upon receiving chat echo
+ Changed: Speech specifies the public room where the message originated from, instead of "Public"
+ Fixed: Strip "/me " from notification popup text, when we are mentioned in action line

_RegEx could be used for this purpose, but there are many caveats with that approach regards escaping special characters etc._

`not in` is much faster than `find()` so the function of these changes is quicker in most cases, unless there's a mention.

`punctuation_list` is generated once upon `_init_` because this is much faster that extrapolating the list of strings for each test. 

A `while` loop is required because the first mention could be a subword, whereas subsequent mention(s) may not be.

- The new function could be just as easily placed into utils.py or textview.py for general purpose usage to search for whole words in any string.


mathiascode:
I'll test this with a few different usernames soon.

mathiascode:
In the future, could you create separate PRs for changes that don't directly relate to the original PR (mainly keep bug fixes and cosmetic changes/general cleanups separate, if possible)? It takes longer to get things reviewed and merged if commits are added days after the PR was submitted, especially if the diff becomes large. Thanks :)

slook:
Agreed this PR snow-balled sorry about that. The unrelated fixes can be implemented another time.

Closing in favour of #1935 :)

PR #1925, opened 2022-02-27
Rename human

hboetes:
I propose renaming the following functions and variable, to make their function clearer:

humanize→human_number
human_length→human_time
h_length→h_time

Which I executed with the following commands:
```
cd pynicotine
sed -i 's|humanize|human_number|' $(ag -l humanize|ag -v config.py)
sed -i 's|human_length|human_time|' $(ag -l human_length)
sed -i 's|h_length|h_time|' $(ag -l h_length)
```

### DIFF
--- a/pynicotine/gtkgui/chatrooms.py
+++ b/pynicotine/gtkgui/chatrooms.py
@@ -57,7 +57,7 @@ from pynicotine.gtkgui.widgets.ui import UserInterface
 from pynicotine.logfacility import log
 from pynicotine.utils import clean_file
 from pynicotine.utils import delete_log
-from pynicotine.utils import humanize
+from pynicotine.utils import human_number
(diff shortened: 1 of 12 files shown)


slook:
Whilst your suggested names make sense, it is likely that user's custom plugins will break as a result of this change.

The confusion isn't large enough to warrant these breaking changes.

hboetes:
I've never heard of somebody writing a custom plugin. %-)

slook:
We have a plugin system, so it can be assumed that people use it and the functions in utils.py are common in those plugins.

For example, there is a plugin called NoEmbed Info that is similar to the YouTube Info plugin, that would be broken by these changes.

mathiascode:
I don't think these function names are bad enough to warrant a name change right now.

hboetes:
ok, no worries

PR #1932, opened 2022-02-28
transfers.py: Don't watch users of Finished status

slook:
+ Changed: When loading stored downloads, it seems pointless to watch users of "Filtered" or "Finished" transfers.


mathiascode:
Thanks, done in accaa331074015e6482f83e4e95e755b726b9485.

PR #1936, opened 2022-03-04
Uploads Settings: Improve strings in Queue Behavior section

slook:
+ Changed: Strings in Queue Behavior section, because the old labels are no longer suitable since the new radio buttons
+ Added: A tooltip to clarify the function of the Dynamic Upload Slots option widget doohickey (open to suggestions)
+ Changed: "Individual limits do not apply to buddies" to clarify that this setting does not affect speed limits

image


pynicotine/gtkgui/ui/settings/uploads.ui
```diff
@@ -240,7 +240,7 @@
           <object class="GtkLabel">
             <property name="visible">True</property>
             <property name="xalign">0</property>
-            <property name="label" translatable="yes">Queue Limits</property>
```
mathiascode, line 243: Would it be better to change `Upload Speed Limits` to `Speed Limits` instead, to avoid redundancy?

pynicotine/gtkgui/ui/settings/uploads.ui
```diff
@@ -380,7 +380,7 @@
                 <child>
                   <object class="GtkLabel">
                     <property name="visible">True</property>
-                    <property name="label" translatable="yes">Upload queue type:</property>
+                    <property name="label" translatable="yes">Free slot allocation:</property>
```
mathiascode, line 383: The old string was more descriptive, imo.

mathiascode:
@hboetes This one is for you.

pynicotine/gtkgui/ui/settings/uploads.ui
slook: The redundancy is needed to aid navigation, describing for instance "the Speed Limits in the Uploads section" is not as easy as referring to "the Upload Speed Limits".

Each section title should uniquely identify the relevant settings, without ambiguity across pages.

pynicotine/gtkgui/ui/settings/uploads.ui
slook: I disagree, "queue type" could seem to mean anything, and we already have "Queue Behavior" as the sub-section title.

The intention here is to associate this setting with the "Free Slot" mechanism, to help users understand that this section controls the behaviour of their Free Slots, i.e. the method by which their Free Slots are allocated.

hboetes:
Yes, this can use some improvements. It requires some careful thought.

slook:
Indeed this Preferences page contains some of the application's most mysterious settings, which many users seem to have only a limited or incorrect understanding of.

The inter-play between the parameters is far from obvious at the moment.

pynicotine/gtkgui/ui/settings/uploads.ui
slook: On the other hand, the shorter sub-section headings are easier to read, see above screenshot.

mathiascode:
I'll revisit this as part of https://github.com/nicotine-plus/nicotine-plus/discussions/1881#discussioncomment-4709086 soon.

pynicotine/gtkgui/ui/settings/uploads.ui
```diff
@@ -384,8 +385,7 @@ First In, First Out: Files will be uploaded in the order they were queued.</prop
               <object class="GtkRadioButton" id="use_upload_slots_radio">
                 <property name="visible">True</property>
                 <property name="hexpand">True</property>
-                <property name="label" translatable="yes">Limit number of upload slots to:</property>
-                <property name="tooltip-text" translatable="yes">If disabled, slots will automatically be determined by available bandwidth limitations.</property>
+                <property name="label" translatable="yes">Maximum number of active uploads:</property>
```
mathiascode, line 388: I like the original "Fixed number of upload slots" proposal more. It's easier to understand if both options use the same terminology.

mathiascode:
Sorry for taking so long on this one. I'm not sure what I didn't like about the change originally, but revisiting it now, it makes sense to me.

pynicotine/gtkgui/ui/settings/uploads.ui
```diff
@@ -359,7 +359,8 @@ First In, First Out: Files will be uploaded in the order they were queued.</prop
               <object class="GtkRadioButton" id="_use_upload_bandwidth_radio">
                 <property name="visible">True</property>
                 <property name="hexpand">True</property>
-                <property name="label" translatable="yes">Queue uploads if total transfer speed reaches (KiB/s):</property>
+                <property name="label" translatable="yes">Allocate upload slots until total transfer speed reaches (KiB/s):</property>
+                <property name="tooltip-text" translatable="yes">Upload slots automatically determined by available bandwidth</property>
```
mathiascode, line 363: Not sure if this tooltip is necessary.

pynicotine/gtkgui/ui/settings/uploads.ui
```diff
@@ -359,7 +359,8 @@ First In, First Out: Files will be uploaded in the order they were queued.</prop
               <object class="GtkRadioButton" id="_use_upload_bandwidth_radio">
                 <property name="visible">True</property>
                 <property name="hexpand">True</property>
-                <property name="label" translatable="yes">Queue uploads if total transfer speed reaches (KiB/s):</property>
+                <property name="label" translatable="yes">Allocate upload slots until total transfer speed reaches (KiB/s):</property>
```
mathiascode, line 362: I think the length of the string was the original issue I had with it, but `total transfer speed -> total speed` seems like a good compromise. Maybe we could also move the Queue Behavior section after Upload Speed Limits to make the relationship more obvious.

PR #1947, opened 2022-03-07
Preferences: Unify Logging / Folder Locations

slook:
+ Removed: Log Folder Locations section
+ Changed: Labels to CheckButtons, the intention here is unclear
+ Added: ToolTips containing the existing strings

Working on this highlights the fact that Log to File toggles needs to be independant of the Log History TextView toggles, otherwise these settings seem to have a misleading effect. The per-tab Log toggles in Chats also add to the confusion.

It's possible that the chat logging toggles and location selectors belong in the Chats -> Chat History section.

If that were the case, then the Downloads / Uploads want splitting apart and placing in the respective pages, which would leave this page rather obsolete, it would become a debug section which seems undesirable.

image

Wider changes to the logging mechanism are required for this to be useful.

### DIFF
--- a/pynicotine/gtkgui/ui/settings/log.ui
+++ b/pynicotine/gtkgui/ui/settings/log.ui
@@ -4,10 +4,10 @@
   <object class="GtkSizeGroup">
     <property name="mode">horizontal</property>
     <widgets>
-      <widget name="RoomLogDirLabel"/>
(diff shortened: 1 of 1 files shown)


slook:
Closing, more thought is required because this approach isn't going to work.

Issue #1963, opened 2022-03-20
Nicotine+ 3.2.2 does not work on Windows


sbstn87:
Hi,

the latest build does not work on Windows 10. After trying to run it, nothing happens. The same happens with 3.2.1. Going back to 3.2.0 solves the issue. Uninstalling Nicotine first and installing either 3.2.1 or 3.2.2 does not help.

mathiascode:
Two things to try:

1. Does the portable Windows version work?
2. Can you temporarily rename the `%AppData%/Roaming/nicotine` folder, and see if Nicotine+ starts?

sbstn87:
1. The portable version does work and preserves my user settings.
2. After renaming the folder Nicotine+ starts as if it was a fresh install - asks to input username and password.

mathiascode:
Can you back up your `%AppData%/Roaming/nicotine` folder, and temporarily remove files inside it (downloads.json and downloads.json.old at once, uploads.json and uploads.json at once, config and config.old at once), one by one, to see which one causes Nicotine+ to stop working?

mathiascode:
Here's a build that should show output in the console, hopefully it will reveal something useful: https://github.com/mathiascode/nicotine-plus/actions/runs/2012428690

You need to start `Nicotine.exe` in an existing console window.

sbstn87:
I tried removing the files you mentioned above, one by one, but Nicotine+ did start successfully each time. 

I cannot see the installer I can download in the link you have provided.

hboetes:
> I cannot see the installer I can download in the link you have provided.

Try scrolling down a bit.

sbstn87:
Here's the output I get:
---
PS C:\Program Files\Nicotine+> .\nicotine+.exe
Windows fatal exception: code 0xc0000139

Thread 0x00000f30 (most recent call first):
  File "<frozen importlib._bootstrap>", line 228 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 1173 in create_module
  File "<frozen importlib._bootstrap>", line 565 in module_from_spec
  File "<frozen importlib._bootstrap>", line 666 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 986 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1007 in _find_and_load
  File "<frozen importlib._bootstrap>", line 228 in _call_with_frames_removed
[...]

PR #1971, opened 2022-03-23
nicotine: load_module -> nicotine

slook:
`load_module` conflicts with the name of a function in the Python runtime library

Troubleshooting #1963


mathiascode:
Since it seems like the old name didn't cause an issue after all, and it's more descriptive, closing.

PR #1972, opened 2022-03-24
gui/userbrowse.py: allow Refresh if connect failed

slook:
+ Changed: Wait until loading shares list is actually in progress before disabling the RefreshButton

In some cases of connection failure the button always remains disabled (greyed-out) because a time-out never triggers the error bar (which re-enables the button), such that if the remote shares has since become available then only the keyboard shortcut is possible to load the shares list, or else the Browse page must be closed and re-opened to initialize browsing.

The blocked UI leads mouse users into wrongly thinking that an initial error has persisted.

A different alternative fix might be possible by calling `show_connection_error()` in cases where all the associated connection attempts are finally stopped, refused or removed, but tracking this is more complex.

### DIFF
--- a/pynicotine/gtkgui/userbrowse.py
+++ b/pynicotine/gtkgui/userbrowse.py
@@ -478,18 +478,15 @@ class UserBrowse(UserInterface):
         else:
             self.progressbar1.set_fraction(0.5)
 
-        self.RefreshButton.set_sensitive(False)
-
     def message_progress(self, msg):
 
         if msg.total == 0 or msg.position == 0:
-            fraction = 0.0
+            self.progressbar1.set_fraction(0.0)
         elif msg.position >= msg.total:
-            fraction = 1.0
+            self.progressbar1.set_fraction(1.0)
         else:
-            fraction = float(msg.position) / msg.total
-
-        self.progressbar1.set_fraction(fraction)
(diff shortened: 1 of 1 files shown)


mathiascode:
> only the keyboard shortcut is possible to load the shares list

Oh no, that's no good. I forgot that the keyboard shortcut can be used at any time, and holding down F5 will spam a user with requests...

mathiascode:
I think the right approach would be to show an error every time a connection closes, but the progress bar hasn't reached 100%. Are there other situations that can lead to a stuck state?

slook:
> I think the right approach would be to show an error every time a connection closes

Good idea, but I don't know a clean method of doing this.

> direct/indirect connection attempts don't show an error by mistake

This is where the complexity lies, a failed attempt doesn't give rise to an error in most cases.

> Are there other situations that can lead to a stuck state?

As far as I know, the only situation that causes the UI to become stuck is if we have a closed port, the remote user has an open port (indirect connection), and their shares are non-configured or yet to scan or a local scanning error has occured at their end.

Also we might as well show a connection error straight away and not even attempt the browse if we know the user is offline, waiting for the time-out seems unreasonably long in this case, since there is no UI feedback whatsoever.

slook:
Closing as the situation is improved now. Any further refinement would require an alternative solution closer to the core.

PR #1975, opened 2022-03-25
trayicon.py: Suppress error if no custom icon theme path is set

slook:
+ Added: Check that the custom icon theme path setting is not "(None)" before scanning directory for icons

`[2022-03-25 14:20:24] [Misc] Error accessing custom tray icon path : [Errno 2] No such file or directory: ''`

### DIFF
--- a/pynicotine/gtkgui/widgets/trayicon.py
+++ b/pynicotine/gtkgui/widgets/trayicon.py
@@ -155,16 +155,19 @@ class BaseImplementation:
 
     @staticmethod
     def check_icon_path(icon_name, icon_path, icon_type="local"):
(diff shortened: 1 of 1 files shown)


pynicotine/gtkgui/widgets/trayicon.py
```diff
@@ -155,16 +155,16 @@ def set_alternative_speed_limit(self, enable):
 
     @staticmethod
     def check_icon_path(icon_name, icon_path, icon_type="local"):
-        """
```
mathiascode, line 158: The check is fine, but I prefer the old comment.

Issue #1448, opened 2021-07-11
Anyone willing to maintain an official Nicotine+ Debian package?


mathiascode:
We would greatly appreciate if someone had the time to reintroduce and maintain the official Nicotine+ Debian package, which was previously removed when Nicotine+ still used Python 2: https://tracker.debian.org/pkg/nicotine

If you're interested, or know anyone else who would be interested in maintaining the official Debian package, please let us know! Thank you. :)

We have unofficial Debian packaging for our PPAs, which could be reused: https://github.com/nicotine-plus/nicotine-plus/tree/master/debian

Relevant links:  
https://www.debian.org/doc/manuals/developers-reference/ch05.html#reintroducing-pkgs  
https://mentors.debian.net/intro-maintainers

Relevant downstream bug reports:  
https://bugs.debian.org/955514
https://bugs.debian.org/966000  
https://bugs.debian.org/990852
https://bugs.debian.org/1004401

mathiascode:
CC @kiplingw @pabs3

kiplingw:
I don't have time, otherwise I would. Thanks for opening the issue @mathiascode.

Rabioli:
I would be willing to do it throught the open SUSE Build Service, not only the Debian repo but others that also their build service manage, I would need help however and guidance as I was trying to make it run on my end I have problems installing it with pip even with all dependencies met. It says something about cairo and a missing confing file.

kiplingw:
Whoever do it would probably be best positioned to do so if they were already an official Debian Developer.

slook:
Debian devs have updated the open bug about this at link:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=990852

On there, Kip Warner <kip@thevertigo.com> (Senior Software Engineer) seems very interested to package a .deb onto Debian buster testing or stable repo's with modern GTK3 dependencies and etc, so it might be worth following up on that bug report.

I am unsure if Kip would be willing to build from source, but he sure seems keen to pull the package in from the PPA tree.

Well done everybody for getting the project working properly again, and good luck getting it back into the main repo, the slsk community appreciates it!

mathiascode:
> Debian devs have updated the open bug about this at link:
> https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=990852
> 
[...]

PR #1977, opened 2022-03-29
debian/copyright: merge with downstream

slook:
Authored by Francois Mazen <francois@mzf.fr> @mzf-guest (Debian package maintainer)

This level of detail about the authors seems to be required for Debian packaging, see link:
https://salsa.debian.org/python-team/packages/nicotine/-/blob/master/debian/copyright

François has kindly compiled this data and seems to have done a thorough job (although perhaps some of the dates and list-order might need to be tweaked to ensure accuracy).

Suggest syncing these changes here into master to keep both debian packaging as close as possible as discussed in https://github.com/nicotine-plus/nicotine-plus/issues/1448#issuecomment-1037728289

### DIFF
--- a/debian/copyright
+++ b/debian/copyright
@@ -1,21 +1,83 @@
-Format: http://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 Upstream-Name: nicotine
 Upstream-Contact: Nicotine+ Team <nicotine-team@lists.launchpad.net>
 Source: https://github.com/nicotine-plus/nicotine-plus/
 
 Files: *
-Copyright: 2001-2003 PySoulSeek Contributors
-           2003-2004 Nicotine Team
-           2004-2022 Nicotine+ Team
+Copyright: 2001-2003 Nir Arbel <nir@slsk.org>
+                     Brett W. Thompson <brettt@tfn.net>
+                     Josselin Mouette <joss@debian.org>
+                     blueboy <bluegeek@eresmas.com>
+                     Christian Swinehart <cswinehart@users.sourceforge.net>
+                     Hyriand <hyriand@thegraveyard.org>
+                     Geert Kloosterman <geertk@ai.rug.nl>
+                     Joe Halliwell <s9900164@sms.ed.ac.uk>
+                     Alexey Vyskubov <alexey.vyskubov@nokia.com>
+                     Jason Green <smacklefunky@optusnet.com.au>
+           2001-2005 Alexander Kanavin <ak@sensi.org>
+           2003-2004 Hyriand <hyriand@thegraveyard.org>
+                     osiris <osiris.contact@gmail.com>
+                     SmackleFunky
+                     (va)\*10^3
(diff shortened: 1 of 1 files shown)


mzf-guest:
Thanks for the upstream update!

> although perhaps some of the dates and list-order might need to be tweaked to ensure accuracy

Do no hesitate to add the missing email addresses and update the history accordingly. The current lists are based on AUTHORS.md and TRANSLATORS.md. Is there other history sources?

debian/copyright
```diff
@@ -27,11 +89,16 @@ Copyright: 2016-2017 Bowtie AB
...
-Copyright: 2014-2022 Tom Wallroth
-           2020-2022 Nicotine+ Team
+Files: pynicotine/metadata/*
+Copyright: 2014-2021 Tom Wallroth
+           2020-2021 Mat (mathiascode) <mail@mathias.is>
 License: MIT
 
+Files: debian/*
+Copyright: 2022 Francois Mazen <francois@mzf.fr>
+                Bastian Germann <bage@debian.org>
```
mzf-guest, line 99: This is only true for the official Debian package, and need to be changed for the package that is provided by nicotine team.

mathiascode:
> Thanks for the upstream update!
> 
> > although perhaps some of the dates and list-order might need to be tweaked to ensure accuracy
> 
> Do no hesitate to add the missing email addresses and update the history accordingly. The current lists are based on AUTHORS.md and TRANSLATORS.md. Is there other history sources?

The correct years for each author can be pulled from the copyright headers of some of the older source files, e.g. pynicotine.py, but this will probably require an afternoon to check all the files. I could do this unless @slook wants to.

Missing email addresses can possibly be found in the commit history, but I'm not sure if it's worth including emails in this file (at least for older authors), considering most of them are likely non-functional at this point.

mathiascode:
Are there guidelines for who should be included in this list? A few of the people listed in AUTHORS.md have never touched the project files, and I'm not sure it would be appropriate to include them as copyright holders of the files.

debian/copyright
slook: I disagree, because both of you have the option of adding yourself or being added by us into AUTHORS.md and about.py at any point should you wish to do so, and such a request is very likely to be approved. Therefore, it is easier just to leave the entries in.

mathiascode:
I've gone through the archives and updated the list here: https://github.com/nicotine-plus/nicotine-plus/commit/3d146fcdd6bf7171f3002c8ed517d2f92fda79fb

It includes authors that have contributed code, art or documentation to the main project codebase/repository (also one-time contributions).

I'll do the same for translators soon.

debian/copyright
mathiascode: @slook Downstream packaging is different from ours, and mine and Kip's names are covered by the general file copyright already.

mathiascode:
Translators added here: https://github.com/nicotine-plus/nicotine-plus/commit/54eb293ecdcb313230a8536d1624a5e72892d585

I had to dig though a lot of files, but at least we have proper credits now.

mathiascode:
@mzf-guest see above for updated lists

PR #1984, opened 2022-04-14
settings/userinterface.ui: clarify 'Prefer dark mode' tip

### LINKED ISSUE #1983: nicotine+ always on dark mode regardless of my settings
Nicotine+ version:  3.2.2
Fedora Linux 35 KDE Plasma spin

### Describe the bug
Nicotine from the moment I installed it with dnf always stays on dark mode even though I unchecked the 'prefer dark mode' box in the user interface settings.

### Expected behavior
Dark mode should only be turned on when the 'prefer dark mode' box is checked and when the box is unchecked the interface should go back to light mode

Fresh install of nicotine+ from the default fedora repos, use a wayland session under kde, attempt to manually toggle dark mode on and off

### Additional context
image

slook:
May close #1983

+ Changed: Tooltip text for 'Prefer dark mode' because the hit-and-miss nature of this option warrants a deeper explanation.

### DIFF
--- a/pynicotine/gtkgui/ui/settings/userinterface.ui
+++ b/pynicotine/gtkgui/ui/settings/userinterface.ui
@@ -34,7 +34,7 @@
                   <object class="GtkCheckButton" id="DarkMode">
                     <property name="label" translatable="yes">Prefer dark mode</property>
                     <property name="visible">True</property>
-                    <property name="tooltip_text" translatable="yes">Note that the operating system&apos;s theme may take precedence.</property>
+                    <property name="tooltip_text" translatable="yes">Nicotine+ can only ask GTK to enable the dark theme, it cannot force the light theme, and the operating system&apos;s theme may take precedence.</property>
                     <signal name="toggled" handler="on_theme_changed"/>
                   </object>
                 </child>


pynicotine/gtkgui/ui/settings/userinterface.ui
```diff
@@ -34,7 +34,7 @@
                   <object class="GtkCheckButton" id="DarkMode">
                     <property name="label" translatable="yes">Prefer dark mode</property>
                     <property name="visible">True</property>
-                    <property name="tooltip_text" translatable="yes">Note that the operating system&apos;s theme may take precedence.</property>
+                    <property name="tooltip_text" translatable="yes">Nicotine+ can only ask GTK to enable the dark theme, it cannot force the light theme, and the operating system&apos;s theme may take precedence.</property>
```
mathiascode, line 37:
```suggestion
                    <property name="tooltip_text" translatable="yes">Disabling this option will use the default color scheme of the operating system.</property>
```

mathiascode: I don't know if my suggestion is much better, but I'd like to avoid mentioning technical details that result in more questions for the average user, especially outside Linux-based operating systems (what is GTK?).

slook: In this case, it is very relevant information to mention "GTK" here, because for example in KDE, there is a dedicated control panel for GTK Appearance settings.

hboetes:
Ask the Desktop Environment to use a dark theme for n+, this may or may not work, depending on various things.

PR #1986, opened 2022-04-17
Pylint: enable 'no-member' check

mathiascode:
Disabling the 'no-member' check made maintenance more difficult, as you have to double (triple) check your changes every time you rename a variable in the code base, and hope you didn't overlook anything. As a result, errors have made it into master, as there's no safety net to catch them.

Dynamically added widgets are the reason for disabling this check in the first place, and these have also caused the most issues. This PR explores ways to solve the problem. It might be worth checking what other PyGObject applications do (Gtk.Template is not an option, since PyGObject only gained support for it in 2019).


mathiascode:
Gaphor retrieves the widgets individually by name, clear and comprehensible but rather verbose and cumbersome: https://github.com/gaphor/gaphor/blob/8d7dcfc079eb6201a1cf5f89e82784ac18f8617b/gaphor/ui/greeter.py#L80-L81

mathiascode:
I'm leaning towards a list like this: https://github.com/nicotine-plus/nicotine-plus/pull/1986/commits/078ac5bc78db6efe863c78c9269aebd4ca8ac4b8

It's easy enough to read and add items to.

slook:
This is beyond my understanding. Please can you clarify why it is not acceptable to just reference the widgets by object id in ui files as before?

mathiascode:
The widget IDs/names are 'magically' added as Python attributes here, with setattr(): https://github.com/nicotine-plus/nicotine-plus/blob/8a2496ebfef7dbdabc59b4eeec87e59b2ec2bc9a/pynicotine/gtkgui/widgets/ui.py#L64

Any attributes added with setattr() won't be recognized by static linters, such as Pylint, since these attributes are modified at runtime. You can disable the 'no-member' check for any reference to these attributes, but 1) there are too many of them, and 2) you're back in a position where invalid references to old attributes aren't caught until you run that specific part of the application.

I want to avoid the usage of setattr() and getattr() when feasible, since it makes it harder to verify if an attribute is still in use across the codebase.

PR #2027, opened 2022-05-18
pynicotine.py: don't log new password string value after changing it

slook:
Displaying and logging the value of this setting in clear text doesn't seem appropriate, because it could pose a security risk.

_Confirmation dialog:_
image

_Log:_
`[2022-05-18 18:14:43] Your password has been changed.`

### DIFF
--- a/pynicotine/pynicotine.py
+++ b/pynicotine/pynicotine.py
@@ -920,7 +920,7 @@ class NicotineCore:
         config.sections["server"]["passw"] = password
         config.write_configuration()
 
-        log.add_important_info(_("Your password has been changed."))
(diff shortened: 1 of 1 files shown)


pynicotine/pynicotine.py
```diff
@@ -920,8 +920,7 @@ def change_password(msg):
         config.sections["server"]["passw"] = password
         config.write_configuration()
 
-        log.add_important_info(
-            _("Your password has been changed. Password is %s"), password)
+        log.add_important_info(_("Your password has been changed."))
```
mathiascode, line 923:
```suggestion
        log.add_important_info(_("Your password has been changed"))
```

PR #2031, opened 2022-05-19
nicotine: Improve command line usage help description string

slook:
- Removed: "Nicotine+ is a Soulseek client" string, in favour of the string defined in `config.summary`
+ Added: Footer epilog with `config.author` + `config.website_url` to match the `man nicotine` page

```
$ ./nicotine --help
usage: nicotine [-h] [-c file] [-u dir] [-p dir] [-t] [-d] [-s] [-b ip]
                [-l port] [-r] [-n] [-v]

Graphical client for the Soulseek peer-to-peer network

...

Nicotine+ Team <https://nicotine-plus.org/>
$ 
```

On most windowed posix terminals the URL <link> is clickable.

### DIFF
--- a/pynicotine/__init__.py
+++ b/pynicotine/__init__.py
@@ -40,7 +40,7 @@ def check_arguments():
     )
     parser.add_argument(
         "-u", "--user-data", metavar=_("dir"),
-        help=_("use non-default user data directory")
+         help=_("use non-default user data directory for e.g. list of downloads")
     )
     parser.add_argument(
         "-p", "--plugins", metavar=_("dir"),
--- a/pynicotine/__init__.py
+++ b/pynicotine/__init__.py
@@ -26,7 +26,7 @@ def check_arguments():
     from pynicotine.config import config
 
     parser = argparse.ArgumentParser(
-        description=config.summary, epilog="%s <%s>" % (config.author, config.website_url), add_help=False
+        description=config.summary, epilog=_("Website: %s") % config.website_url, add_help=False
     )
 
     # Visible arguments
--- a/pynicotine/__init__.py
+++ b/pynicotine/__init__.py
@@ -26,11 +26,14 @@ def check_arguments():
     from pynicotine.config import config
 
     parser = argparse.ArgumentParser(
-        description="%s %s - %s" % (config.application_name, config.version, config.summary),
-        epilog="%s <%s>" % (config.author, config.website_url)
+        description=config.summary, epilog="%s <%s>" % (config.author, config.website_url), add_help=False
     )
 
     # Visible arguments
+    parser.add_argument(
+        "-h", "--help", action="help",
+        help=_("show this help message and exit")
+    )
     parser.add_argument(
         "-c", "--config", metavar=_("file"),
         help=_("use non-default configuration file")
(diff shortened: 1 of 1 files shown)


pynicotine/__init__.py
```diff
@@ -24,20 +24,20 @@ def check_arguments():
 
     import argparse
     from pynicotine.config import config
-    parser = argparse.ArgumentParser(description=_("Nicotine+ is a Soulseek client"), add_help=False)
```
mathiascode, line 27: Built-in help message is not translatable.

pynicotine/__init__.py
```diff
@@ -24,20 +24,20 @@ def check_arguments():
...
 
-    # Visible arguments
-    parser.add_argument(
-        "-h", "--help", action="help",
-        help=_("show this help message and exit")
+    parser = argparse.ArgumentParser(
+        description="%s %s - %s" % (config.application_name, config.version, config.summary),
```
mathiascode, line 29: We have --version for this. This should only be a description of the program.

pynicotine/__init__.py
```diff
@@ -24,20 +24,20 @@ def check_arguments():
...
+        epilog="%s <%s>" % (config.author, config.website_url)
```
mathiascode, line 30: Are there other programs that do this? URL might be fine, not sure if author is relevant here.

slook: > other programs that do this?

Some have informational help text or copyright information and such like:

```
$ geany --help
...

Report bugs to https://github.com/geany/geany/issues.

$ 
```

> URL might be fine, not sure if author is relevant here.

The URL looks odd on it's own, the `man nicotine` page is the inspiration.

pynicotine/__init__.py
```diff
@@ -37,7 +40,7 @@ def check_arguments():
     )
     parser.add_argument(
         "-u", "--user-data", metavar=_("dir"),
-        help=_("use non-default user data directory for e.g. list of downloads")
```
mathiascode, line 40: If we remove the -p argument, would it be worth mentioning plugins as an example here?

slook: Yes good idea, possibly, the string I have at the moment is:

"run with alternative user profile directory"

Ultimately, the string will be used on `--profile`, but I am tentative to implement `-p` straight away, incase somebody is routinely using this option to start the program, mixing and migrating these options cause havoc (although, there is not any critical data loss).

"alternative user directory for data and plugins"

pynicotine/__init__.py
mathiascode: > The URL looks odd on it's own, the `man nicotine` page is the inspiration.

That section mentions the author of the manpage though. I think something like `Website: https://nicotine-plus.org` would make more sense here.

slook: Yes, I had thought about that it would have been my first choice, but unfortunately we don't currently have a translatable string for "Website", which was why I went for author instead. It can be changed but the translators won't have time for the release of 3.2.3.

I suggest we merge this now and then change it to "Website: %s" at a later date.

mathiascode: I don't think this is an important change for 3.2.3. You can include a Website string here, and I'll just backport the description change to 3.2.3.

PR #2032, opened 2022-05-20
SLSKPROTOCOL.md: rendering problem with HTML table

slook:
https://nicotine-plus.org/doc/SLSKPROTOCOL.html#sending-login-example

This table does not render properly on the Nicotine+ website, because the page's HTML flattens the table as a paragraph into `<p>` tags instead of `<table>` tags. It's fine when viewing the markdown page from within the GitHub website.

I don't know if the issue is caused by italic `<em>` "`*Data*`" `</em>` text in the table headers, or if it's the lack of `</p>` line breaks after the "`*Message:*`" titles, so this PR tries both fixes, but I don't have an easy way of testing how GitHub Pages will convert the MD into HTML until this PR is merged.

This has been broken since commit https://github.com/nicotine-plus/nicotine-plus/commit/8a9490ae921ee562034b8c1b922fe678be976d0e#r74084767

### DIFF
--- a/doc/SLSKPROTOCOL.md
+++ b/doc/SLSKPROTOCOL.md
@@ -239,18 +239,21 @@ We send this to the server right after the connection has been established. Serv
 ### Sending Login Example
 
 *Message:*
-| *Data*    | Message Length | Message Code | Username Length | Username                | Password Length | Password                |
+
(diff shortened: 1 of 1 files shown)


slook:
> testing how GitHub Pages will convert the MD into HTML

Ideal, the markup works properly now :)

PR #2034, opened 2022-05-20
pynicotine: remove "-t" and "-d" command line options for Tray Icon

slook:
The -t and -d command line options are not very useful and perhaps confusing to the user (as to this being a temporary state and not permanent config change). Also the command line arguments have no effect in certain modern environments such as macOS.

- Removed: Obsolete "-t" command line option and "--enable-trayicon" argument
- Removed: Obsolete "-d" command line option and "--disable-trayicon" argument

In the future the removed options could be better used for example:
+ `-d file` `--download file` specified by an slsk:// URL then exit (or stay running)
+ `-d folder` `--download folder` count the files and size of a remote folder and prompt to download it
+ `-t` print a list of currently active `--transfers` or do some `--terminal` console related function
+ `-t abort` stop and clear all active  transfers
+ `-t abort user` stop and clear active and queued transfers for a specified user
+ `-t pause` stop all active and queued transfers
+ `-t resume` start all paused transfers and retry all failed transfers

### DIFF
--- a/pynicotine/__init__.py
+++ b/pynicotine/__init__.py
@@ -44,14 +44,6 @@ def check_arguments():
         help=_("use non-default directory for plugins")
     )
     parser.add_argument(
-        "-t", "--enable-trayicon", action="store_true",
-        help=_("enable the tray icon")
-    )
-    parser.add_argument(
-        "-d", "--disable-trayicon", action="store_true",
-        help=_("disable the tray icon")
-    )
-    parser.add_argument(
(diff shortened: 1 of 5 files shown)


mathiascode:
I don't see the point of keeping the long arguments. The fewer undocumented arguments we have, the better.

mathiascode:
Thanks. I'll let this change brew in 3.3.0.dev1 for a while. If there are strong arguments for keeping these options, we can possibly restore them.

PR #2041, opened 2022-05-21
SLSKPROTOCOL.md: Server Code 1 reason "Bad Password" -> "INVALIDPASS"

slook:
Correction to documented server generated reason string "Bad Password" was incorrect...

Receive Login Failure

-    **bool** failure **0**
-    **string** reason _Almost always:_ **INVALIDPASS** _(sometimes it's a banned message or another error)._


mathiascode:
Better, but I think we should create a list of possible reasons. I don't think ban messages are ever sent this way.

mathiascode:
As far as I know:

- INVALIDPASS: wrong password for user
- INVALIDUSERNAME: username too long or contains invalid characters

Issue #1728, opened 2021-12-01
Support paths longer than 260 characters on Windows


redactedscribe:
Nicotine+ version: 3.2.0.rc2
Operating System/Distribution: Windows 11

### Describe the bug
```
Error while scanning metadata for file C:\XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX.mp3: [Errno 2] No such file or directory: 'C:\\XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX.mp3'
```
First path is 280 characters, and the second is 284 (due to escaped backslashes `\\` within the path (only the first is shown. the rest are censored)).

### Expected behavior
Scan paths longer than 260 characters.

### Additional context
I thought I had mentioned this somewhere before but I couldn't find the issue.

mathiascode:
It was briefly mentioned in https://github.com/nicotine-plus/nicotine-plus/issues/1380

In order to have widespread support for long paths in Nicotine+, we have to start using the "\\?\" prefix for paths when necessary. This needs to be thoroughly tested though, since there are probably things that will break if the prefix is present.

[...]

PR #2046, opened 2022-05-22
transfers.py: prevent incomplete filename from exceeding 255 bytes

AtticFinder65536:
From the first commit:
>e86229103aef8ee95027e391b5989eee5ea3441e fixed an issue where
downloading a file with a filename of 255 characters would fail
because a prefix was added to its filename in the incomplete directory
which would create a filename over 255 characters, which isn't possible
on many systems.
The restriction on most systems is 255 bytes, not characters, so this
could still fail when multi-byte characters are present. Convert
the filename to bytes when truncating (prefix doesn't need this because
its byte length is known).

This will result in the extension being truncated if needed, unlike before. If we want the extension then we would need to truncate it as well otherwise a file with a filename containing a long extension can still fail to download. We could also try truncating the perfect amount to fill 255 bytes, but I don't care about both of these so haven't bothered doing them.

### DIFF
--- a/pynicotine/transfers.py
+++ b/pynicotine/transfers.py
@@ -1036,12 +1036,12 @@ class Transfers:
                     md5sum = md5()
                     md5sum.update((i.filename + i.user).encode('utf-8'))
 
-                    base_name, extension = os.path.splitext(clean_file(i.filename.replace('/', '\\').split('\\')[-1]))
+                    filename = i.filename.replace('/', '\\').split('\\')[-1]
                     prefix = "INCOMPLETE" + md5sum.hexdigest()
 
-                    # Ensure file name doesn't exceed 255 characters in length
+                    # Ensure filename doesn't exceed 255 bytes in length
                     incomplete_name = os.path.join(
-                        incompletedir, prefix + base_name[:255 - len(prefix) - len(extension)] + extension)
+                        incompletedir, prefix + filename.encode('utf-8')[:213].decode('utf-8', 'ignore'))
                     file_handle = open(incomplete_name.encode('utf-8'), 'ab+')  # pylint: disable=consider-using-with
 
                     try:
--- a/pynicotine/transfers.py
+++ b/pynicotine/transfers.py
@@ -1036,7 +1036,7 @@ class Transfers:
                     md5sum = md5()
                     md5sum.update((i.filename + i.user).encode('utf-8'))
 
-                    filename = i.filename.replace('/', '\\').split('\\')[-1]
+                    filename = clean_file(i.filename.replace('/', '\\').split('\\')[-1])
                     prefix = "INCOMPLETE" + md5sum.hexdigest()
 
                     # Ensure filename doesn't exceed 255 bytes in length


slook:
It seems excessive to have such a long hash in the first place, is there not a way to reduce this down to a more reasonable length, say 8 characters, that would still be enough to uniquely identify identical files from different users?

mathiascode:
> It seems excessive to have such a long hash in the first place, is there not a way to reduce this down to a more reasonable length, say 8 characters, that would still be enough to uniquely identify identical files from different users?

I'm not going to break backwards compatibility and carry more migration code for the next 10 years.

slook:
> going to break backwards compatibility

Since the Incomplete files are created and then deleted during runtime, I cannot see how this would be a breaking change? 

Perhaps the only issue might arise if the user upgraded Nicotine+ whilst there are downloads in progress, in which case the transfer would start again from the beginning and the old Incomplete file would never be automatically deleted.

pynicotine/transfers.py
```diff
@@ -1036,12 +1036,12 @@ def file_download_init(self, msg):
                     md5sum = md5()
                     md5sum.update((i.filename + i.user).encode('utf-8'))
 
-                    base_name, extension = os.path.splitext(clean_file(i.filename.replace('/', '\\').split('\\')[-1]))
```
slook, line 1039: I have a feeling that `clean_file()` is required because it is possible the filename may include characters which the OS does not allow.

AtticFinder65536: Sure, I can add that. I was wondering, does Nicotine+ replace any of the characters in [ILLEGALPATHCHARS](https://github.com/nicotine-plus/nicotine-plus/blob/cbd09acb6506f6e58c192702e9a41a244d790879/pynicotine/utils.py#L41) on moving the complete file to the download directory? What about in displaying filenames in the GUI? Because I use a system that supports any byte except '/' and NUL, I would want to avoid replacing characters whenever possible.

slook: > Sure, I can add that

Thanks, better to be on the safe side, especially considering this is only a temporary file nobody cares if the filename is not precisely correct.

> does Nicotine+ replace any of the characters in [ILLEGALPATHCHARS](https://github.com/nicotine-plus/nicotine-plus/blob/cbd09acb6506f6e58c192702e9a41a244d790879/pynicotine/utils.py#L41) on moving the complete file to the download directory?

Yes, `clean_file()` uses [ILLEGALFILECHARS](https://github.com/nicotine-plus/nicotine-plus/blob/cbd09acb6506f6e58c192702e9a41a244d790879/pynicotine/utils.py#L42) which does contain all of the characters in ILLEGALPATHCHARS and '\\', '/' which is called from `get_download_destination()`:

https://github.com/nicotine-plus/nicotine-plus/blob/cbd09acb6506f6e58c192702e9a41a244d790879/pynicotine/transfers.py#L1755-L1761

> I would want to avoid replacing characters whenever possible

Perhaps the constants containing the characters could be customized, maybe with `sys.platform` or some other logic in utils.py to determine the exact capabilities of the filesystem. However, this would give different results on different systems leading to file duplication or other confusions, whereas at the moment it is the same limitations for everybody.

AtticFinder65536: > Perhaps the constants containing the characters could be customized, maybe with `sys.platform` or some other logic in utils.py to determine the exact capabilities of the filesystem. However, this would give different results on different systems leading to file duplication or other confusions, whereas at the moment it is the same limitations for everybody.
> 
> Another way to do it might be to always attempt to first create the file with the raw name, and then catch an exception to fall back to the clean_file name if required.

I'll implement it privately or maybe in a public fork then, and later submit upstream in case it's a welcome change for everybody to use.

AtticFinder65536:
>Once tested, this PR would help with https://github.com/nicotine-plus/nicotine-plus/issues/1728 (but won't close it)

I should've mentioned, I have tested this PR and it works great. Now I can download files with long CJK filenames for example. So I think it should be ok to merge now. Unless we want the extension to be visible (from what I know the only use case for having the filename here in the first place is so that someone peeking in the incomplete dir can know which file is which, eg. you want to play a file that is partially complete and the sender is offline).. in that case we could truncate the extension to a maximum of 5 bytes so that the user can see '.flac' '.alac' '.opus' etc., then basename is 213 minus 5.
I don't know how this PR will help with #1728, it seems that is about pathname length which is not used in the incomplete file name, and `incomplete_name` is not used anywhere else other than inside the incomplete directory.

slook:
> Unless we want the extension to be visible ... can know which file is which

It is nice if the icons are correct for the filetype, on some systems the file icon displayed is still selected based upon the filename extension, and also like you say this breaks the association when launching a file from a file manager, so this might be considered an irritating regression for anybody who is in the habit of doing so, as they will end up trying to play a jpeg or something which probably doesn't sound very nice!

> we could truncate the extension to a maximum of 5 bytes so that the user can see '.flac' '.alac' '.opus' etc., then basename is 213 minus 5.

What would it take for you to implement this? Would you be willing to revisit it? I think it could be worthwhile.

> I don't know how this PR will help with https://github.com/nicotine-plus/nicotine-plus/issues/1728, it seems that is about pathname length

Right I see, my mistake. I had thought that pathname length in the context of Windows meant the entire length of the path including filename, such as to be affected depending upon the location of the Incomplete folder. This is not my area of expertise, and it seems we are stuck on that issue since it has been open for a long time now (and was a known issue from many years ago) perhaps you would be suitably skilled to have a look at this as well?! I wonder what other Python applications do to get around this issue, surely it must be a common problem that has a solution we can get ideas from. That could help a lot of users.

> I have tested this PR and it works great

Ideal :) Perhaps there will be a chance to get this reviewed soon, is what I meant... at which point it is sure to be a welcome contribution that will be appreciated by many! Meanwhile, if you preserve the filename extension that would be a good way forward imo.

mathiascode:
I'd like to preserve the file extension as well.

> > I don't know how this PR will help with #1728, it seems that is about pathname length
> 
> Right I see, my mistake. I had thought that pathname length in the context of Windows meant the entire length of the path including filename, such as to be affected depending upon the location of the Incomplete folder. This is not my area of expertise, and it seems we are stuck on that issue since it has been open for a long time now (and was a known issue from many years ago) perhaps you would be suitably skilled to have a look at this as well?! I wonder what other Python applications do to get around this issue, surely it must be a common problem that has a solution we can get ideas from. That could help a lot of users.

I had a partial fix for that issue, but it's an annoying issue that I haven't revisited it in a while. We need to use the "\\?\" prefix for file paths on Windows only, and also verify that we don't mix regular slashes and backslashes in the same path, since the long path API is strict about this.

AtticFinder65536:
Closing in favour of #2053, that one keeps the extension and all 255 bytes are used.

slook:
Thank you @AtticFinder65536 if it wasn't for your help we wouldn't have come to any solution on this.

PR #2049, opened 2022-05-24
Main Window: Show important log messages with InfoBar instead of MessageDialog

slook:
- Removed: Don't use MessageDialog for such messages, because a UI obstruction is caused if they originate whilst another dialog is open (which is most likely since those messages are related mainly to login and password success / failures).
+ Added: InfoBar to display 'important_info' and 'important_error' log level messages relating to login and connection issues.
- Changed: Don't display a duplicate of an important message in the status bar (but still write it into the log file/console).

_For clarity, the code added to `mainwindow.ui` is as follows..._
```
        <!-- Status Bar -->
        <child>
          <object class="GtkBox">
            <property name="orientation">vertical</property>
            <child>
              <object class="GtkRevealer">
                <property name="transition-type">slide-up</property>
                <child>
                  <object class="GtkInfoBar" id="info_bar">
                  </object>
                </child>
              </object>
            </child>
```
... then the entire Status Bar block is indented such that the InfoBar is nested above the Status Bar (it was not possible to add the InfoBar in it's own separate GtkBox container because then there is a 1px visual artifact line).

image

In the future, it could be worthwhile adding option buttons to infobar.py in order to replace the Invalid Password dialog, and possible use for other serious connection errors or urgent issues that require immediate user intervention.

### DIFF
--- a/pynicotine/gtkgui/frame.py
+++ b/pynicotine/gtkgui/frame.py
@@ -52,6 +52,7 @@ from pynicotine.gtkgui.widgets.filechooser import FileChooser
 from pynicotine.gtkgui.widgets.iconnotebook import TabLabel
 from pynicotine.gtkgui.widgets.dialogs import MessageDialog
 from pynicotine.gtkgui.widgets.dialogs import OptionDialog
+from pynicotine.gtkgui.widgets.infobar import InfoBar
 from pynicotine.gtkgui.widgets.notifications import Notifications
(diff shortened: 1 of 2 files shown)


slook:
Closing, since it is agreed that attaching an ugly InfoBar to the main window is undesirable and shall not be required anyway to solve a problem that shouldn't exist in the first place.

@mathiascode Please copy your above points into https://github.com/nicotine-plus/nicotine-plus/discussions/1881 to keep track of the improvements that should be made elsewhere.

PR #2050, opened 2022-05-25
Important log messages: Improve user experience during first run Setup Assistant

slook:
This PR is intended to smooth out first run login and connection error messages, whilst making a start on partially resolving some of issues described in https://github.com/nicotine-plus/nicotine-plus/discussions/1881#discussioncomment-2814439 as well as coincidentally stumbling upon a closely related fix for https://github.com/nicotine-plus/nicotine-plus/discussions/1881#discussioncomment-2798560

+ Added: Indefinite delay before `_update_port_mapping()` to prevent wiping out important log messages during first run welcome page, because the UPnP status was out of context here

+ Moved: During `FastConfigure()` first run, the UPnP() module now waits while need_config, so set prior to "port" page such that UPnP status_bar message appears at the appropriate time in context, which is after having set the username and password. This is helpful to keep status text relevant when UPnP is not available on the network in both GUI and headless mode.
- ~~Changed: Don't disable the "Connect" menu item, because it will be able to trigger FastConfigure() if `need_config()`~~

- Changed: Warning about "INVALIDUSERNAME" or other login error in status_bar instead of a UI blocking popup dialog

+ ~~Moved: Check for need_config during connect() instead of start() to handle missing configuration more gracefully~~
- ~~Removed: Handle triggering of `on_fast_configure()` from core with a ui_callback to `connect()` instead~~

_(crossed out items are changes that have now been better implemented in master with commit https://github.com/nicotine-plus/nicotine-plus/commit/7ce72584845d57e47ee747e5fef74a66d1e72303)_

Unfortunately all the above fixes are dependent upon eachother to varying degrees, so I have had to lump it all in to one PR. I've tried to keep it as focused as possible, imo it is good enough to be merged into master/3.2.x for wider testing, or perhaps just used for planning a better implementation as you see fit.

More fixes for smoother password changing etc. are also in the works that are closely related to the above fixes, but I do not intend to add any further commits to this PR unless you need any changes made, any suggestions welcome.

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -61,8 +61,8 @@ jobs:
         shell: msys2 {0}
     env:
       ARCH: ${{ matrix.arch }}
-      NICOTINE_GTK_VERSION: 4
-      NICOTINE_LIBADWAITA: 1
+      NICOTINE_GTK_VERSION: 3
+      NICOTINE_LIBADWAITA: 0
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
--- a/packaging/windows/setup.py
+++ b/packaging/windows/setup.py
@@ -250,6 +250,7 @@ from pynicotine.config import config  # noqa: E402  # pylint: disable=import-err
 
 setup(
     name=config.application_name,
+    description=config.application_name,
     author=config.author,
     version=re.sub(r".(dev|rc)(.*)", "", config.version),
     options={
@@ -274,14 +275,16 @@ setup(
             applications_shortcut=True
         )
     },
+    packages=[],
     executables=[
         Executable(
             script=os.path.join(PYNICOTINE_PATH, "nicotine"),
             target_name=config.application_name,
             base=GUI_BASE,
             icon=os.path.join(PYNICOTINE_PATH, "packaging/windows/nicotine.ico"),
+            copyright=config.copyright,
             shortcut_name=config.application_name,
-            shortcut_dir="StartMenuFolder"
+            shortcut_dir="ProgramMenuFolder"
         )
     ],
 )
--- a/po/ca.po
+++ b/po/ca.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2022-04-22 02:04+0300\n"
-"PO-Revision-Date: 2022-05-22 17:52+0000\n"
+"PO-Revision-Date: 2022-05-25 19:17+0000\n"
 "Last-Translator: Maite Guix <maite.guix@gmail.com>\n"
 "Language-Team: Catalan <https://hosted.weblate.org/projects/nicotine-plus/"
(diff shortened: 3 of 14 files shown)


mathiascode:
> * CANT FIX: This PR causes a slight visual regression whereby the Setup Assistant dialog appears deactiveated (greyed-out) on first run, but this is better than the off-centre problem and fine from the menu (probably related to something in dialogs.py).

This bricks the interface on GNOME. The setup assistant appears before the main window, and is placed behind it, but the main window isn't interactive.

slook:
Thanks for trying it :) By the way I like your refactoring in the application callbacks etc, this modular approach seems like it is a good way forward.

slook:
> Please create a discussion for your intended changes before too much work is put into it.

Done. Thanks, I would very much appreciate your input @mathiascode see https://github.com/nicotine-plus/nicotine-plus/discussions/1879#discussioncomment-2827187 (3.2.x) and https://github.com/nicotine-plus/nicotine-plus/discussions/1881#discussioncomment-2827130 (master) for some ideas.

mathiascode:
> I have marked this PR ready for review not really for merging, but you might be interested in the behaviour of the UPnP wait > I have marked this PR ready for review not really for merging, but you might be interested in the behaviour of the UPnP wait loop (which serves to make the status_bar text less confusing) and to highlight the issue about lack of validation in Setup Assistant that is a bad idea to set it without prompting.

I agree that port forwarding should be delayed, mainly so the user can select a port (it's something I've wanted to do for a long time), but I'd have to think about the best approach. I don't think your solution is the correct one here, for two reasons:
- Logic related to username and password doesn't belong in upnp.py
- It keeps a thread alive for no good reason. The thread should be created when necessary.

mathiascode:
> I have marked this PR ready for review not really for merging, but you might be interested in the behaviour of the UPnP wait > I have marked this PR ready for review not really for merging, but you might be interested in the behaviour of the UPnP wait loop (which serves to make the status_bar text less confusing) and to highlight the issue about lack of validation in Setup Assistant that is a bad idea to set it without prompting.

I agree that port forwarding should be delayed, mainly so the user can select a port (it's something I've wanted to do for a long time), but I'd have to think about the best approach. I don't think your solution is the correct one here, for two reasons:
- Logic related to username and password doesn't belong in upnp.py
- It keeps a thread alive for no good reason. The thread should be created when necessary.

slook:
Both your points are good ones, and I knew my method is a hack to demonstrate a conceptual idea only. Clearly there are larger changes required elsewhere if this strategy is to be considered for further development.

Perhaps there already exists a related core thread that is already alive which could be used for this purpose?

If this turns out not to be viable, then I it could well be more graceful to simply refine the initialization entry points of the relevant modules to fire at the appropriate times.

It makes me wonder, what it would take to include the port range settings into need_config() , maybe it could be split into `need_login_config()` and `need_port_config()`, leave the ports set to None at startup for first run and pre-fill the TextEntry boxes with `config.defaults` on fastconfigure.show(), but that is beyond the scope of our discussion for now.

To attempt the first login, can we just call core.connect() on_switch_page() , or is some other method needed to handle this?

PR #2053, opened 2022-05-27
transfers.py: limit incomplete file names to 255 bytes

mathiascode:
@AtticFinder65536 Does this solution work fine on your end?

### DIFF
--- a/pynicotine/transfers.py
+++ b/pynicotine/transfers.py
@@ -48,6 +48,7 @@ from pynicotine.utils import clean_file
 from pynicotine.utils import clean_path
 from pynicotine.utils import get_result_bitrate_length
 from pynicotine.utils import load_file
+from pynicotine.utils import truncate_string_byte
 from pynicotine.utils import write_file_and_backup
 
 
@@ -1035,13 +1036,17 @@ class Transfers:
                     from hashlib import md5
                     md5sum = md5()
                     md5sum.update((i.filename + i.user).encode('utf-8'))
-
-                    base_name, extension = os.path.splitext(clean_file(i.filename.replace('/', '\\').split('\\')[-1]))
                     prefix = "INCOMPLETE" + md5sum.hexdigest()
 
-                    # Ensure file name doesn't exceed 255 characters in length
-                    incomplete_name = os.path.join(
-                        incompletedir, prefix + base_name[:255 - len(prefix) - len(extension)] + extension)
+                    # Ensure file name doesn't exceed 255 bytes in length
+                    base_name, extension = os.path.splitext(clean_file(i.filename.replace('/', '\\').split('\\')[-1]))
+                    base_name_limit = 255 - len(prefix) - len(extension.encode('utf-8'))
+                    base_name = truncate_string_byte(base_name, base_name_limit)
+
+                    if base_name_limit < 0:
+                        extension = truncate_string_byte(extension, 255 - len(prefix))
+
+                    incomplete_name = os.path.join(incompletedir, prefix + base_name + extension)
                     file_handle = open(incomplete_name.encode('utf-8'), 'ab+')  # pylint: disable=consider-using-with
 
                     try:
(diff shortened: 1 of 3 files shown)


AtticFinder65536 (approved):
I tested a bunch of filenames in a python script containing the relevant code, and it works (every `os.path.join(prefix + base_name + extension)` at the end of my script had a byte length of 255 (sometimes a few less because incomplete multi-byte characters are ignored when encoding)). The extension is always visible. When the extension is over 255 characters only the extension is visible and also truncated. A lot better than my solution.

slook:
I suppose extention means any filename containing one or full-stops `.` so has the solution been tested on such files?

AtticFinder65536:
>I suppose extension could mean any filename containing one or more full-stops . so has the solution been tested on such files?
>...

The extension variable becomes `.extension`, and the full filename inside the incomplete directory becomes `INCOMPLETEffffffffffffffffffffffffffffffffThis is the first part of the filename. This is the second. - (and, finally)... The third.extension`. [Python documentation demonstrates how it will look like](https://docs.python.org/3.10/library/os.path.html#os.path.splitext)

Here is what I used to test filenames as they will be processed by Nicotine+ in this PR:
```
#!/usr/bin/env python
import os
import os.path

ILLEGALPATHCHARS = ['?', ':', '>', '<', '|', '*', '"']
ILLEGALFILECHARS = ILLEGALPATHCHARS + ['\\', '/']
REPLACEMENTCHAR = '_'

def clean_file(filename):
    for char in ILLEGALFILECHARS:
        filename = filename.replace(char, REPLACEMENTCHAR)
    return filename

import sys
filename = sys.argv[1]

print("filename: " + filename)
print("byte length of filename: " + str(len(filename.encode('utf-8', 'ignore'))))
prefix = "INCOMPLETEffffffffffffffffffffffffffffffff"
print("prefix: " + prefix)
print("byte length of prefix: " + str(len(prefix.encode('utf-8', 'ignore'))))

def truncate_string_byte(string, byte_limit, encoding='utf-8'):
    """ Truncates a string to fit inside a byte limit """
    return string.encode(encoding)[:max(byte_limit, 0)].decode(encoding, 'ignore')

base_name, extension = os.path.splitext(clean_file(filename.replace('/', '\\').split('\\')[-1]))
print("base_name: " + base_name)
print("byte length of base_name: " + str(len(base_name.encode('utf-8', 'ignore'))))
print("extension: " + extension)
print("byte length of extension: " + str(len(extension.encode('utf-8', 'ignore'))))

base_name_limit = 255 - len(prefix) - len(extension.encode('utf-8'))
print("base_name_limit: " + str(base_name_limit))
print("byte length of base_name_limit: " + str(len(str(base_name_limit).encode('utf-8', 'ignore'))))
base_name = truncate_string_byte(base_name, base_name_limit)
print("new base_name: " + base_name)
print("byte length of new base_name: " + str(len(base_name.encode('utf-8', 'ignore'))))

if base_name_limit < 0:
	print("base_name_limit is less than zero")
	extension = truncate_string_byte(extension, 255 - len(prefix))
	print("new extension: " + extension)
	print("byte length of new extension: " + str(len(extension.encode('utf-8', 'ignore'))))

incompletedir = "/home/user/.local/share/nicotine/incomplete"
incomplete_name = os.path.join(incompletedir, prefix + base_name + extension)
print("incomplete_name: " + incomplete_name)
print("byte length of incomplete_name: " + str(len(incomplete_name.encode('utf-8', 'ignore'))))

print("also,")
fullfilename = os.path.join(prefix + base_name + extension)
print("fullfilename: " + fullfilename)
print("byte length of fullfilename: " + str(len(fullfilename.encode('utf-8', 'ignore'))))
```

slook:
Ideal, so that's correct then.

Please can you swap out the tabs for 4 spaces in your above codeblock, the indentation is broken otherwise. It is an interesting script could this be worth implementing into a CI test module?

mathiascode:
Thanks for testing! I've added a unit test.

PR #2056, opened 2022-05-28
NEWS.md: Nicotine+ 3.2.3 Bug Fixes, Changes and Corrections

slook:
+ Added: "IMPORTANT" issue about CPU usage in Uploads tab (we don't really know if this is rare or not, and it is severe).
+ Added: @AtticFinder65536's solution to prevent incomplete filename from exceeding 255 bytes
- Changed: The order of the Corrections section in 3.2.3 into the order of subjective importance.
- Changed: Clarified about "total" time, and changed word order about subfolders of folders containing subfolders.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -31,6 +31,7 @@ Issues closed on GitHub
 
  * UnicodeEncodeError when downloading file ([#1980](https://github.com/nicotine-plus/nicotine-plus/issues/1980))
  * Nicotine+ always on dark mode regardless of my settings ([#1983](https://github.com/nicotine-plus/nicotine-plus/issues/1983))
+ * High CPU Usage (pegs a single core) ([#1998](https://github.com/nicotine-plus/nicotine-plus/issues/1998))
  * Weird behavior when I download a whole folder with multiple subdirectories ([#2004](https://github.com/nicotine-plus/nicotine-plus/issues/2004))
  * Don't freeze/crash without xdg-open ([#2005](https://github.com/nicotine-plus/nicotine-plus/issues/2005))
  * Unicode encode error ([#2015](https://github.com/nicotine-plus/nicotine-plus/issues/2015))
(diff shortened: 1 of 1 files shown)


mathiascode:
> we don't really know if this is rare or not, and it is severe

Safe to say if affects all systems to some degree. I've seen it on my system when working on the transfer treeviews in the past, and Gtk.TreeViewColumnSizing.AUTOSIZE is inefficient as the GTK docs mention.

NEWS.md
```diff
@@ -38,6 +39,7 @@ Issues closed on GitHub
  * Time Left column regression ([#2020](https://github.com/nicotine-plus/nicotine-plus/issues/2020))
  * Clear ALL downloads, Clear ALL uploads not working ([#2023](https://github.com/nicotine-plus/nicotine-plus/issues/2023))
  * User note saves when pressing Cancel ([#2036](https://github.com/nicotine-plus/nicotine-plus/issues/2036))
+ * High CPU Usage (pegs a single core) ([#1998](https://github.com/nicotine-plus/nicotine-plus/issues/1998))
```
mathiascode, line 42: sort according to issue number

slook:
It is cumulative in nature on my system (as in to say the more Uploads in the list then the worse it gets becoming noticeable at 500 or more and then severe at 2000 or more), but it was surprising to find out that the bug presented on a system with Autoclear enabled where there were only a few items at any one time.

mathiascode:
> but it was surprising to find out that the bug presented so severely on a system with Autoclear enabled where there were only a few items at any one time.

The issue report mentioned that there were many active users in the upload list (at least 9 at one point), and assuming these users queue up a few albums, you quickly reach a large number of rows.

slook:
Well yes that's true. It is promising to see that the resource usage is hardly noticeable now that this treeview column limitation has been solved, and now that the time estimates update smoothly every second there is a massive difference to the perceived performance of the application nowadays.

PR #2065, opened 2022-06-01
gui/search.py: on_refilter() optimize in/out

slook:
Moved: Don't do `strip()` or `lower()` on blank strings, instead wait until it we know it is actually required.


mathiascode:
Closing in favor of #2071

PR #2070, opened 2022-06-02
Rework custom icon theme system

mathiascode:
- Stop using GIcon to improve scrolling performance in treeviews
- Allow changing icon theme without restarting Nicotine+


slook:
The tray icon and flags appear correctly (btw i'm still testing with Debian 9 / Python 3.5 GTK 3.22.11 for another week or two.)

mathiascode:
Can you compare scrolling smoothness of the user list in the nicotine room? This change doesn't really affect CPU usage, and I'm not sure we can do anything about it, but it should result in slightly higher FPS in lists with many status icons.

mathiascode:
> The scrolling FPS was never an issue for me

I believe it's only noticeable if the system is under heavy load.

> What about skipping the getting of icon if the Status or Country icon column(s) is not visible?

How would this make a difference? The CPU usage when scrolling is high no matter which columns are visible.

slook:
How would this make a difference?

Logically thinking, scrolling should be lighter on CPU if there are no image render columns, but this isn't the case. It is my guess that even an invisible column(s) might consume CPU.

mathiascode:
I don't believe it's the icons that consume CPU here. It happens in the transfer views as well. Some of it is due to GTK 3 not being hardware accelerated, GTK 4 uses less CPU.

PR #2071, opened 2022-06-03
search.py: various optimizations for search filters

mathiascode:
Since the filter checks are performed on each row, it makes sense to do more work when creating the filters.

### DIFF
--- a/pynicotine/gtkgui/search.py
+++ b/pynicotine/gtkgui/search.py
@@ -909,22 +909,20 @@ class Search(UserInterface):
         if self.active_filter_count == 0:
             return True
 
-        file_path = row[11].lower()
-
         for filter_id, filter_value in self.filters.items():
             if not filter_value:
                 continue
 
-            if filter_id == "filtertype" and not self.check_file_type(filter_value, file_path):
+            if filter_id == "filtertype" and not self.check_file_type(filter_value, row[11].lower()):
                 return False
 
             if filter_id == "filtercc" and not self.check_country(filter_value, row[12].upper()):
                 return False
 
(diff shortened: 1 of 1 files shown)


pynicotine/gtkgui/search.py
```diff
@@ -918,33 +909,35 @@ def check_filter(self, row):
         if self.active_filter_count == 0:
             return True
 
-        filters = self.filters
+        file_path = row[11].lower()
```
slook, line 912: `.lower()` should be done under the two conditions that require it in (`"filterin"`, `"filterout"`), because most of the filter_id's don't use the (potentially quite long) string value in this variable, and much of the time it is most likely (the default) that no filters.items() will be set anyway.

pynicotine/gtkgui/search.py
```diff
@@ -1418,30 +1411,51 @@ def on_refilter(self, *_args):
...
 
         if filter_out:
             try:
-                filter_out = re.compile(filter_out)
+                filter_out = re.compile(filter_out.lower())
             except sre_constants.error:
                 filter_out = None
 
+        if filter_size:
+            filter_size = filter_size.replace(" ", "").split("|")
```
slook, line 1436: These `replace(" ", "")` operations make the `strip()` operations above redundant.

mathiascode: Problem is, if you just have whitespace as the value, say " ", you end up with an empty string in the list after splitting [""].

mathiascode: Reduced it to a single operation in bceb85700c4044cce604bd11947a6bb67c76e1ee

PR #2072, opened 2022-06-04
gtkgui: use timestamp format preferences in more contexts

AtticFinder65536:
Some lines don't use the rooms_timestamp and log_timestamp preference, but I feel like they should. See commit messages.

[privatechat.py uses the private_timestamp pref for disconnected/reconnected messages](https://github.com/nicotine-plus/nicotine-plus/blob/master/pynicotine/gtkgui/privatechat.py#L297) so all messages in private chats use the same user-specified timestamp format, so 9c721c4da801899f947628323399109094de1f5e should be ok, just fixing something that was forgotten. As for 1ba07ffb79ecb44cdc37f08a31caca4a26dc0572 I'm not 100% sure if this is a welcome change, but I personally prefer it. To allow the user to set the timestamp format of the messages in the log history pane, we could create a new config option next to the existing log_timestamp option for it, but I think reusing the log_timestamp option like I've done is better.

I tested a port of this PR to 3.2.2 and it works - all timestamps shown in the GUI are now using my preferred format.

### DIFF
--- a/pynicotine/gtkgui/chatrooms.py
+++ b/pynicotine/gtkgui/chatrooms.py
@@ -668,7 +668,8 @@ class ChatRoom(UserInterface):
 
             if lines:
                 timestamp_format = config.sections["logging"]["rooms_timestamp"]
-                self.chat_view.append_line(_("--- old messages above ---"), self.tag_hilite, scroll=False, timestamp_format=timestamp_format)
+                self.chat_view.append_line(_("--- old messages above ---"), self.tag_hilite, scroll=False,
+                                           timestamp_format=timestamp_format)
 
     def populate_user_menu(self, user, menu, menu_private_rooms):
 
--- a/pynicotine/gtkgui/chatrooms.py
+++ b/pynicotine/gtkgui/chatrooms.py
@@ -879,7 +879,9 @@ class ChatRoom(UserInterface):
 
         if not self.core.network_filter.is_user_ignored(username) and \
                 not self.core.network_filter.is_user_ip_ignored(username):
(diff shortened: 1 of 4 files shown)


pynicotine/gtkgui/chatrooms.py
```diff
@@ -667,7 +667,8 @@ def append_log_lines(self, path, numlines):
                                                timestamp_format="", scroll=False)
 
             if lines:
-                self.chat_view.append_line(_("--- old messages above ---"), self.tag_hilite, scroll=False)
+                timestamp_format = config.sections["logging"]["rooms_timestamp"]
+                self.chat_view.append_line(_("--- old messages above ---"), self.tag_hilite, scroll=False, timestamp_format=timestamp_format)
```
mathiascode, line 671: This exceeds the maximum line length of 120, otherwise this PR looks good.

mathiascode:
Maybe we should use the room timestamp for user join/leave/away messages as well?

AtticFinder65536:
>Maybe we should use the room timestamp for user join/leave/away messages as well?

Yes, I have done that on my system but I forgot to do it here... I'll do that shortly.

mathiascode:
> Yes, I have done that on my system but I forgot to do it here... I'll do that shortly.

Thanks!

> I also updated the translation template. When making the change initially I wasn't sure if I should have done it, I think I looked through some other PRs to find if it was done if they changed lines, but didn't see it done so I didn't do it.

Could you revert this for now? I'll update the template once 3.2.3 has been released.

AtticFinder65536:
>Could you revert this for now? I'll update the template once 3.2.3 has been released.

Done

PR #2076, opened 2022-06-07
Clarify use of autoaway option, make it a button

AtticFinder65536:
> The autoaway option could be disabled by setting it to 0, but this is
> not extremely obvious. Split it into two options: UseAutoAway and
> AutoAwayInterval. Create an away timer only if UseAutoAway is enabled.
> Turn the label into a check button, clarify its functionality to the
> user, and set the lowest possible value to 1 because the button now
> disables it instead of when a value of 0 is used.

I found it a little bit confusing at first ([and someone else did recently](https://www.reddit.com/r/Soulseek/comments/v61o4y/nicotine_plus_randomly_switching_to_away_without/ibcuwl1/?context=3)) so just a small change to clarify its usage.
(button turns it on and off for those who don't want it)


pynicotine/config.py
```diff
@@ -180,7 +180,8 @@ def load_config(self):
                 "ipignorelist": {},
                 "ipblocklist": {},
                 "autojoin": ["nicotine"],
-                "autoaway": 15,
```
slook, line 183: Changing the data type of an existing config setting might lead to runtime errors if the user ever needs to revert back to an older version. Please test for backwards compatibility.

Perhaps ideally the setting could be converted to a tuple `(15, True)`, but I don't think doing this works when rolling back to older versions.

It would be easier for maintenance to keep the existing setting's name and create a new one for the CheckButton.

slook: Runtime errors which are silent and don't throw a critical error are the worst, because it can lead to mysterious behaviour.

slook:
Closing, since it is agreed that no additional button is required.

PR #2078, opened 2022-06-07
textview.py: scroll to bottom when height changes

mathiascode:
@slook Can you test this? It's not perfect (still has visual glitches if the history is large), but it should solve https://github.com/nicotine-plus/nicotine-plus/discussions/1535#discussioncomment-1294120. Works in both GTK 3 and GTK 4.


slook:
Massive improvement :)

The recent messages are now correctly in view after resizing. A very satisfactory fix.

Although the positioning of the scroll when in the middle is still unpredictable, this is minor by comparison to the outright failure of scroll-to-bottom as it was before.

PR #2080, opened 2022-06-08
ui/settings/network.ui: clarify autoaway pref use

AtticFinder65536:
To keep backwards compatibility, the previous behaviour that autoway is disabled when set to 0 should be kept, so no need to make it a button, just clarify the existing behaviour in the UI.


mathiascode:
Keep in mind that strings should generally be short. Your string spans two lines, and will likely span three in some translations, as translated strings are typically longer.

I propose something like this:
`Minutes of inactivity until away (0 to disable):`

If anyone has better suggestions around that length, feel free to mention them.

slook:
I too noticed that the string was long, consider a ToolTip if an expanded description is required.

pynicotine/gtkgui/ui/settings/network.ui
```diff
@@ -251,7 +251,8 @@
                   <object class="GtkLabel">
                     <property name="visible">True</property>
                     <property name="hexpand">True</property>
-                    <property name="label" translatable="yes">Toggle away status after minutes of inactivity:</property>
+                    <property name="label" translatable="yes">Minutes of inactivity before going away (0 to disable):</property>
+                    <property name="tooltip-text" translatable="yes">Once you interact with the application after idling, status will be set to online.</property>
```
slook, line 255: "idling" -> "being away"

pynicotine/gtkgui/ui/settings/network.ui
```diff
@@ -251,7 +251,8 @@
                   <object class="GtkLabel">
                     <property name="visible">True</property>
                     <property name="hexpand">True</property>
-                    <property name="label" translatable="yes">Toggle away status after minutes of inactivity:</property>
+                    <property name="label" translatable="yes">Minutes of inactivity before going away (0 to disable):</property>
+                    <property name="tooltip-text" translatable="yes">Once you interact with the application being away, status will be set to online.</property>
```
slook, line 255: "being away" -> "after being away"

AtticFinder65536: I should get some more sleep.

PR #2131, opened 2022-07-23
Remove link parsing of usernames that are links

Nachtalb:
If a username is an HTTP link or any other kind of link it should not be treated as such, as it can mislead or confuse people. You also don't get the user context menu when right-clicking, which makes it more of a hassle to do any user interactions. 

I don't know if this is the best approach. As this `append_line` method is used in quite a few contexts.

## Before: 
image
image
In the wild:  image

## After:
image
image


mathiascode:
Thanks! I refactored append_line in 68857bd5b0dd311395bffe886b3b88a2d233343f to hopefully make it more readable, and applied a similar fix there.

PR #2132, opened 2022-07-25
theme.py: Dim selected but unfocused list item in TreeViews

slook:
Concept for an accessibility improvement for visually identifying the focused widget list item by means of dimming non-focused items by 50%.

It is useful for the dual view navigation in Browse Shares as well as doing multi-selections in any other TreeView, especially when using the keyboard to do selections...

image

image

Note: The highlighted selections auto-complete dropdown lists are also dimmed.

### DIFF
--- a/pynicotine/gtkgui/widgets/theme.py
+++ b/pynicotine/gtkgui/widgets/theme.py
@@ -218,6 +218,16 @@ def set_global_css():
         -GtkTreeView-vertical-separator: 5;
     }
 
+    treeview:selected {
+        /* Dim selected but unfocused list items */
+        background-color: alpha(@theme_selected_bg_color, 0.5);
+    }
+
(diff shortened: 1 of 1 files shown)


slook:
It not affecting color, it's opacity..

I don't know if there is a more standard way of showing focus or not.

mathiascode:
But you're still referencing a color `@theme_selected_bg_color`, possibly overriding something a theme had done differently. Libadwaita doesn't use `@theme_selected_bg_color` here, and uses opacity for all selected rows, as an example.

> I don't know if there is a more standard way of showing focus or not.

I don't know either, wonder if any other GTK applications do something like this.

slook:
Example in PCManFM http://sourceforge.net/projects/pcmanfm

PR #2152, opened 2022-08-12
Only backup files once per session

slook:
This way, at least it might be possible for a user to recover their old data from the previous session if something went wrong or the user made an error. Also never overwrite the 'config.old' file with config.defaults

Writing through the existing file without moving it beforehand seems to work fine (not tested on all systems).

TODO: Seperate `restore_file()` function, it seems that restoring the old data after a failed save will not be appropriate now.

### DIFF
--- a/pynicotine/config.py
+++ b/pynicotine/config.py
@@ -569,6 +569,11 @@ class Config:
         # Update config values from file
         self.set_config()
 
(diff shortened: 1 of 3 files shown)


slook:
Closing as it works better the way it is.

PR #2161, opened 2022-08-25
Chat Rooms: Include room name in Public feed mention notification

slook:
Include the actual room name in notifications for mentions in the Public feed, and avoid duplicate notifications if the room's chat tab is open.

Fixes https://github.com/nicotine-plus/nicotine-plus/discussions/2145#discussioncomment-3418096

### DIFF
--- a/pynicotine/gtkgui/chatrooms.py
+++ b/pynicotine/gtkgui/chatrooms.py
@@ -805,7 +805,7 @@ class ChatRoom(UserInterface):
 
         self.chatrooms.request_tab_hilite(self.container, mentioned)
 
-        if public and room in self.chatrooms.pages:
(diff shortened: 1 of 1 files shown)


pynicotine/gtkgui/chatrooms.py
```diff
@@ -796,37 +796,41 @@ def ticker_add(self, msg):
...
+    def show_notification(self, login, room, user, text, tag, public=False):
 
         if user == login:
             return
 
         mentioned = (tag == self.tag_hilite)
 
+        self.chatrooms.request_tab_hilite(self.container, mentioned)
+
+        if public and room in self.chatrooms.pages:
```
mathiascode, line 808:
```suggestion
        if public and room in self.core.chatrooms.joined_rooms:
```

PR #2162, opened 2022-08-25
Spamfilter: Ignore repeated chat room messages

slook:
Added: It can block messages that are exact repeats of recent messages, such as when multiple user's plugins all respond automatically to a chat room command.

The default parameter setting of `3` lines covers the use case of multiple Reddit plugins, whereby only the first set of responses are seen.


mathiascode:
I will hold off from merging this for some time, to see if it's still needed after recent Nicotine+ versions have trickled down to more users.

PR #2198, opened 2022-10-08
mainwindow.ui: Room List -> Rooms

slook:
"Room List" is ambiguous with list of Users in a room.

The name of the user list is "Users", so the room list should be "Rooms"


mathiascode:
Looks good. Windows CI has been flaky recently.

PR #2204, opened 2022-10-08
settings/shares.ui: shorten explaination

slook:
The minor detail about "hidden files" is ambiguous with Soulseek 'private files' or buddy-only shares, it's confusing to mention this here.


slook:
Closing in favour of #2205 since there is no compelling reason to have a different string in each dialog.

PR #2205, opened 2022-10-08
settings/shares.ui: change explanation

slook:
Matches the string in Setup Assistant

better alternative to #2204 , The minor detail about "hidden files" is ambiguous with Soulseek 'private files' or buddy-only shares, it's confusing to mention this here.


mathiascode:
@hboetes Thoughts?

hboetes:
> Soulseek users will be able to download from your shares. Please contribute to the Soulseek network by sharing your own collection and by resharing what you downloaded from other users.

PR #2207, opened 2022-10-10
Add separate Shift+Ctrl+H shortcut for Set Away Status

slook:
+ Added: Shift+Ctrl+H key shortcut action for setting Away status
+ Changed: Ctrl+H key shortcut action for setting Online status, instead of toggling Away/Online
- Moved: Action declaration code within set_up_actions() since Away is not an item in the File menu anymore, also grouped Away Mode functions as appropriate.

The purpose of separating these shortcuts is to prevent a server ban for rapid Away/Online status toggling that was possible due to key-hold repeat of the current Away toggle shortcut key.

### DIFF
--- a/pynicotine/gtkgui/frame.py
+++ b/pynicotine/gtkgui/frame.py
@@ -537,9 +537,6 @@ class NicotineFrame(Window):
     def on_disconnect(self, *_args):
         self.core.disconnect()
 
-    def on_away(self, *_args):
-        self.core.set_away_mode(self.core.user_status != UserStatus.AWAY, save_state=True)
-
     def on_soulseek_privileges(self, *_args):
 
         import urllib.parse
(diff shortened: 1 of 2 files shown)


mathiascode:
This is unnecessarily complex in my opinion. I would either change the shortcut to Shift+Ctrl+H to set both away/online, or add a small cooldown so it's no longer possible to ban yourself (or maybe both).

slook:
> change the shortcut to Shift+Ctrl+H to set both away/online

This wouldn't solve the problem, key repeat is still possible no matter which modifier is used.

> add a small cooldown so it's no longer possible to ban yourself

I'd be concerned that a cooldown might make the UI feel unresponsive.

mathiascode:
> This wouldn't solve the problem, key repeat is still possible no matter which modifier is used.

Sure, but it would be more difficult to accidentally ban yourself. We should probably avoid changing existing shortcuts if possible, though.

> I'd be concerned that a cooldown might make the UI feel unresponsive.

We could limit it to the callback for the shortcut.

slook:
Good idea, that does work quite well and does not feel obstructive. Although, there is probably a cleaner way to do this, open to suggestions... see #2208

PR #2210, opened 2022-10-11
shares.py: Warning if share(s) cannot be accessed before a rescan

### LINKED ISSUE #1698: warning if no files are shared

slook:
This would close #1698

+ Added: Log warning if share(s) cannot be accessed before a rescan wipes out the shares database
+ Added: Option dialog prompt for "Retry", "Force Rescan" or "Configure Shares"

Implementation of a confirmation dialog as discussed in #1698

_If no shares are accessible:_
Screenshot_2022-10-19_20-00-10

.
.

_If one share cannot be found:_
```
2022-10-19 20:05:38 2 shares configured
2022-10-19 20:05:38 1 share ready
2022-10-19 20:05:38 1 share not found
2022-10-19 20:05:38 Rescan aborted: Missing public share "PLAY" at /media/user/PLAY
```
Screenshot_2022-10-19_20-01-25

,
.

If more than one share**s** are not accessible::
```
2022-10-19 20:08:20 3 shares configured
2022-10-19 20:08:20 1 share ready
2022-10-19 20:08:20 2 shares not found
2022-10-19 20:08:20 Rescan aborted: 2 shares not found
```
Screenshot_2022-10-19_20-08-37

.
.
_On "Retry" after the missing shares are mounted:_
```
2022-10-19 20:11:39 3 shares configured
2022-10-19 20:11:39 3 shares ready
2022-10-19 20:11:39 Rescanning shares…
2022-10-19 20:11:39 276 folders found before rescan, rebuilding…
2022-10-19 20:11:40 Rescan complete: 276 folders found
```

It seems to be successful at preventing inadvertent database loss, but it is quite obnoxious because the box doesn't have a close (X) or cancel button. However, if there are no ready (or no configured) shares to scan then "Ignore" is offered instead of "Force Rescan", allowing the dialog to be easily dismissed if the user doesn't wish to share anything.

The "Force Rescan" button changes to "Ignore" if there a no available shares to scan then it seems pointless to allow a force scan which would result is zeroing out the database.

The future intention is to integrate the shares checking mechanism into the `/listshares` command which would require some sort of output, and  also there is some long-winded experimental code which aims to provide inspiration for improved flexibility as to how multiple share "groups" (i.e. share_type) could potentially be addressed in a more logical way.

### DIFF
--- a/pynicotine/shares.py
+++ b/pynicotine/shares.py
@@ -834,38 +834,65 @@ class Shares:
         return False
 
     @staticmethod
+    def count_shares(shared_folders):
+        """ Return the total number of configured shared folders """
+
+        num_shares = 0
+
+        for shares in shared_folders:
+            for virtual, folder, *_unused in shares:
+                num_shares += 1
+
+        return num_shares
+
+    @staticmethod
     def check_shares(shared_folders):
         """ Return shared folder paths that are inaccessible """
 
         errors = []
 
-        for virtual, folder, *_unused in shared_folders[1]:
-            if not os.access(str(folder), os.R_OK):
-                errors.append(f"Cannot access share {virtual} with real path {folder}")
+        for shares in shared_folders:
+            for virtual, folder, *_unused in shares:
+                if not os.access(str(folder), os.R_OK):
+                    errors.append(f"Cannot access share {virtual} with real path {folder}")
 
         return errors
+
+    def confirm_rescan(self, shared_folders):
+        """ Verify if all shares are mounted before continuing to rescan """
+
+        errors = self.check_shares(shared_folders)
+
+        if not errors:
+            return True
+
+        num_errors, num_shares = len(errors), self.count_shares(shared_folders)
+        title = f"Failed to access {num_errors} of {num_shares} configured shares"
+        message = '\n'.join(errors)
+
+        log.add(message)
+
+        if num_errors >= 2:
+            log.add(title)
+
+        if self.ui_callback:
+            # Prompt to retry or force rescan
+            return self.ui_callback.confirm_rescan_dialog(title, message)
+
+        return False
 
     def rescan_shares(self, init=False, rescan=True, rebuild=False, use_thread=True, force=False):
 
         if self.rescanning:
             return None
 
-        self.rescanning = True
         shared_folders = self.get_shared_folders()
-        errors = self.check_shares(shared_folders)
 
-        if errors:
-            log.add('\n'.join(errors))
+        if not force:
+            rescan = (rescan and self.confirm_rescan(shared_folders))
 
-            if len(errors) >= 2:
-                log.add(f"Failed to access {len(errors)} shares")
-
-            if not force and rescan and self.ui_callback:
-                pass  # TODO: Confirm abort/retry/continue (force rescan)
-
-            rescan = (force and rescan)
-            
         # Hand over database control to the scanner process
+        self.rescanning = True
         self.close_shares(self.share_dbs)
 
         scanner, scanner_queue = self.build_scanner_process(shared_folders, init, rescan, rebuild)
--- a/pynicotine/shares.py
+++ b/pynicotine/shares.py
@@ -858,26 +858,26 @@ class Shares:
         return False
 
     def list_shares(self, group=None):
-        """ Returns ONE list with all readable shares and missing shares.
-        group specifies which shares to check: "normal", "buddy", etc """
+        """ Returns ONE string with all readable shares and missing shares.
+        group specifies which shares to list: "normal", "buddy", etc. """
         # TODO: Unused, need new command in core_commands to run this
 
         reads, fails = self.check_shares(shares=None, group=group)
         num_reads = len(reads)
         num_fails = len(fails)
         num_total = num_reads + num_fails
+        total_shares_line = "0 " + _("shares") + " " + _("configured")
 
         if not num_total:
-            head_line = "No configured shares"
-            fails.insert(0, head_line)
+            fails.insert(0, total_shares_line)
 
-        # Coalesce all "Ready " and "Missing " items toegether
-        all_shares = '\n\n'.join(reads) + "\n" + '\n\n'.join(fails)
+        # Coalesce the summary line toegether with all "Ready " and "Missing " items
+        all_shares = total_shares_line + ":\n\n" + '\n\n'.join(reads) + "\n" + '\n\n'.join(fails)
 
         return all_shares
 
     def check_shares(self, shares=None, group=None):
-        """ Returns TWO lists of: readable shares, and missing shares.
+        """ Returns TWO lists of strings: readable shares, and missing shares.
         group specifies which shares to check: "normal", "buddy", etc. """
 
         if shares is None:
@@ -892,9 +892,9 @@ class Shares:
 
             for virtual, folder, *_unused in table:
                 if os.access(folder, os.R_OK):
-                    reads.append(f"Ready {group_name} share \"{virtual}\" at:\n{folder}")
+                    reads.append(_(f"Ready {group_name} share \"{virtual}\" at:\n{folder}"))
                 else:
-                    fails.append(f"Missing {group_name} share \"{virtual}\" at:\n{folder}")
+                    fails.append(_(f"Missing {group_name} share \"{virtual}\" at:\n{folder}"))
 
             if group is not None:
                 # Only check a specific group (ie just "normal" or "buddy" shares, etc)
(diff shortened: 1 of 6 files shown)


hboetes:
I would use: "Cannot access share {virtual} at {folder}"

hboetes:
Looks nice so far.

- Isn't there that clever hack to automagically say i.e. 1 share/2 shares?
- unindex... technically correct, I think some people would be scratching their head. I propose "remove share"

hboetes (approved):
Call in the General.

pynicotine/shares.py
```diff
@@ -833,15 +833,135 @@ def process_scanner_messages(self, scanner, scanner_queue):
...
+            return 0
+
+        if group in (1, "buddy", "private", "trusted"):
+            return 1
+
+        return -1
+
+    @staticmethod
+    def _get_group_name(group=0):
+        """ Convienience function to identify a specific shares database by [index] or alias """
```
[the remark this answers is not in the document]
slook: Nothing is convenient about the shares module!

slook:
> Isn't there that clever hack to automagically say i.e. 1 share/2 shares?

There are lots of strings in the program which have this problem, if you know of an elegant method of approaching this then that would be great.

I have avoided showing the summary count lines "Cannot access 1 of 1 configured shares" where there is only 1 missing share is obviosuly more useful to display the actual missing share name and path (primarily so that the most relevant information is retained in the status bar).

> unindex... technically correct, I think some people would be scratching their head. I propose "remove share"

Agreed, I'm struggling with this one. Unfortunately your proposal is too ambiguous with deleting the share from the filesystem. For now I have gone with "exclude"?

slook:
@mathiascode Do you know the reason why the test_startup CI fails at finding " 0 " in the output? I have seen this error elsewhere many times recently so I'm unsure as to whether or not it is failing due to the changes in this PR or not (there is still a " 0 " in the output). How about changing it to "Nicotine+" or just "+ " would that pickup the loading line?

mathiascode:
> @mathiascode Do you know the reason why the test_startup CI fails at finding " 0 " in the output? I have seen this error elsewhere many times recently so I'm unsure as to whether or not it is failing due to the changes in this PR or not (there is still a " 0 " in the output). How about changing it to "Nicotine+" or just "+ " would that pickup the loading line?
> 
> This PR introduces a new string which outputs " 1 " of " 1 " in the log during the headless ci-mode run, so I tried changing it to that in [b2149a0](https://github.com/nicotine-plus/nicotine-plus/commit/b2149a0db5d9855d6d0fd603f732984954f8ff6f) but it didn't make any difference.

I don't know the reason, but you can run the tests locally with `python -m unittest` or `pytest`. Add a temporary print statement in the test to see what output contains.

mathiascode:
https://github.com/nicotine-plus/nicotine-plus/commit/46c7256d8d0bb465305c8b77b9d49647a906754e will provide more helpful error messages in this case.

slook:
Thanks, that does help yes.

AssertionError: ' 0 ' not found in "b'[2022-10-18 21:21:23] No configured shares\\n[2022-10-18 21:21:23] No configured shares\\n\\n'"
[130](https://github.com/nicotine-plus/nicotine-plus/actions/runs/3277002631/jobs/5393730293#step:7:131)

pynicotine/shares.py
```diff
@@ -833,15 +857,129 @@ def process_scanner_messages(self, scanner, scanner_queue):
...
+
+        num_reads = len(reads)
+        num_fails = len(fails)
+        num_total = num_reads + num_fails
+
+        reads_shares_word = _("share") if num_reads == 1 else _("shares")
+        fails_shares_word = _("share") if num_fails == 1 else _("shares")
+        total_shares_word = _("share") if num_total == 1 else _("shares")
+
+        total_shares_line = f"{num_total} {total_shares_word} " + _("configured")
```
mathiascode, line 920: The largest issue with this kind of string building is that the word order will be different in different languages. It's difficult for translators to a) realize that these strings should be translated in the context of a sentence, and b) piece them together. There could also be conflicts with individual words elsewhere in the application in the future.

mathiascode, line 920: The largest issue with this kind of string building is that the word order will be different in different languages. It's difficult for translators to a) realize that these strings should be translated in the context of a sentence, and b) piece them together. There could also be conflicts with individual words elsewhere in the application in the future.

mathiascode:
> There are lots of strings in the program which have this problem, if you know of an elegant method of approaching this then that would be great.

The only way of doing this is by using [ngettext](https://docs.python.org/3/library/gettext.html#gettext.ngettext) instead of an underscore, but there are some problems:

1. Almost all of our needs for plural strings are in context menus, and we would have to update the menu items every time we display a context menu, which is annoying.
2. GtkBuilder (for our UI files) doesn't support plural strings, as far as I know.

I think it would be worth checking if we can rewrite some of our strings in ways that avoid the need for plural strings completely. If we want to use plural strings, they need to be supported everywhere for consistency.

And no, you can't simply determine that a plural string should be used if a number is larger than 1, because several languages have their own plural rules... https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html

slook:
_17 source strings viewed as list:_
https://hosted.weblate.org/browse/nicotine-plus/nicotine-plus/en/?q=check:optional_plural

mathiascode:
> GtkBuilder (for our UI files) doesn't support plural strings, as far as I know.

This isn't a large issue though, you can update the labels in the Python code instead.

In any case, we don't have a lot of unpluralized strings, so I'm not entirely opposed to using ngettext in these cases. However, I would prefer to use it only when there is no other option. A handful of our existing strings would sound good enough if they only used the plural form, for example.

slook:
So for example, we can perhaps assume that a folder might ordinarily be expected to contain multiple "Subfolders" in the case of "Download Folder & Subfolder(s)", and in the case of the "User(s)" menu (probably those actions are rarely performed on a multi selection), it could be changed to "User Actions", since there are always multiple actions in that menu.

mathiascode:
> So for example, we can perhaps assume that a folder might ordinarily be expected to contain multiple "Subfolders" in the case of "Download Folder & Subfolder(s)", and in the case of the "User(s)" menu (probably those actions are rarely performed on a multi selection), it could be changed to "User Actions", since there are always multiple actions in that menu.

Precisely.

I will review this PR more closely soon, but looking at the screenshot alone, I feel like we could already save ourselves unnecessary work here. Does the user need to see the exact number of failed shares in the first sentence, or is it enough to change it to something like "Cannot access the following configured shares:" and simply provide the list?

slook:
> but looking at the screenshot

The screenshot was outdated, I've just posted updated ones showing 0, 1 and 2 missing shares.

slook:
> a small message

When testing with anymore than half a dozen shares, the buttons were not visible on screen, so if there are more than 5 items then the message body is trucated...

Screenshot_2022-10-19_20-17-50

> What would be nice ... e.g. check if removable drives are plugged in, network shares are mounted, folder permissions etc.

Agreed. The "Retry to check again" string can be revised to include something like this.

mathiascode:
> > is it enough to change it to something like "Cannot access the following configured shares:" and simply provide the list?
> 
> 
> 
> No, because with anymore than half a dozen shares, the buttons were not visible on screen because the OptionDialog() can grow tall beyond the screen boundary, so to work around this problem if there are more than 5 items then the message body gets trucated, hence all the importance of having the folder counts clearly displayed in the title.

How much value does a count provide if you don't know which folders have a problem?

We could probably add a scrolling widget to the dialog, it's needed for the critical error dialog anyway.

slook:
> We could probably add a scrolling widget to the dialog, it's needed for the critical error dialog anyway

I was thinking exactly the same thing. I'd be keen to see what that looks like, unfortunately I don't know how to do that.

If you add a ScrolledWindow to the dialog, then I will remove the trunctation code from here and post a screenshot.

mathiascode:
> If you add a ScrolledWindow to the dialog, then I will remove the trunctation code from here and post a screenshot.

Added in 1cc6bd828f5574aaee18ebb36e44082da13272f7

pynicotine/shares.py
```diff
@@ -834,15 +858,120 @@ def process_scanner_messages(self, scanner, scanner_queue):
...
+
+        # Check if any shares are missing and log counts
+        (num_reads, num_fails, num_total, summary_total,
+         summary_reads, summary_fails, _reads_text, fails_text, fails) = self.count_shares(shares)
+
+        if num_reads and not num_fails:
+            # Continue initializing shares, and do the rescan now
+            return True
+
+        log.add(_("Rescan aborted") + ": " + (fails_text.replace(":\n", " ") if num_fails == 1 else summary_fails))
```
mathiascode, line 936: In the future, please avoid creating strings like this. It becomes difficult to understand what the message is, without jumping back and forth in the code to see what fails_text is.

We also can't be sure that every language wants a colon that looks exactly like this at this specific location (French uses a space before punctuation, for example).

mathiascode:
Here's the approach I would go with, for improved maintainability: https://github.com/nicotine-plus/nicotine-plus/pull/2237

slook:
Closing this PR since your approach is way cleaner :)

pynicotine/shares.py
slook: Indeed it is the product of an over-complicated design. I learn more from doing this project that these over-engineered string hacks are hardly ever suitable for all cases.

That's interesting I did not know about the French language punctuation.

mathiascode: It takes a bit of time to get used to, but the more you work on different projects, the more you appreciate a simple solution that does just enough, over of a complex solution with all kinds of bells and whistles. All about finding a balance that both the developer/maintainer and user appreciates.

Rule of thumb for translated strings is to always try to use full sentences.

PR #2220, opened 2022-10-21
Implement event scheduler

mathiascode:
Alternative to threading.Timer that only uses a single thread.


pynicotine/scheduler.py
```diff
@@ -0,0 +1,81 @@
...
+        callback()
+        self._add_repeat(delay, callback, event_id)
+
+    def _add_repeat(self, delay, callback, event_id):
+        self._events[event_id] = self._scheduler.enter(
+            delay=delay, priority=1, action=self._callback_repeat, argument=(delay, callback, event_id))
+
+    def add(self, delay, callback, repeat=False):
+
+        event_id = ''.join(random.choice(string.digits) for _ in range(8))
```
slook, line 52: For what purpose is it needed to select an event randomly, as opposed to selecting items in a linear sequence? Why number 8 is it to with with max_sockets or something else?

mathiascode: This was taken from popupmenu.py, but you're right, there's no need for a random ID.

PR #2229, opened 2022-10-24
config.py: move "Unknown config option" into debug log

### LINKED ISSUE #2228: Unknown config option 'filterlength' in section 'searches'
On Manjaro Linux I get the following message when I option the program.
```
Unknown config option 'filterlength' in section 'searches'
Loading GTK 3.24.34
Loading Python 3.10.8 (main, Oct 13 2022, 21:13:48) [GCC 12.2.0]
Loading Nicotine+ 3.2.6
```
I've just updated to 3.2.6 and I don't recall seeing this message before the update.

slook:
Closes #2228 

It might be a good idea to move this log entry into the debug log, since it is not really an important enough problem for it to be in the standard log.

### DIFF
--- a/pynicotine/config.py
+++ b/pynicotine/config.py
@@ -660,13 +660,13 @@ class Config:
 
                 # Check if config section exists in defaults
                 if i not in self.defaults and i not in self.removed_options:
-                    log.add(_("Unknown config section '%s'"), i)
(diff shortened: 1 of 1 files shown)


mathiascode (commented):
Debug log strings should not be translatable.

slook:
Changed to non-translatable string.

PR #2237, opened 2022-10-31
Abort rescan if shares are unavailable

### LINKED ISSUE #1698: warning if no files are shared

mathiascode:
Closes #1698

### DIFF
--- a/pynicotine/cli/application.py
+++ b/pynicotine/cli/application.py
@@ -101,6 +101,10 @@ class Application:
     def on_critical_error_threading(args):
         raise args.exc_value
 
+    def shares_unavailable(self, shares):
+        for virtual_name, folder_path in shares:
+            log.add("• \"%s\" %s" % (virtual_name, folder_path))
+
     def show_scan_progress(self):
         # Not implemented
         pass
--- a/pynicotine/gtkgui/frame.py
+++ b/pynicotine/gtkgui/frame.py
@@ -1468,13 +1468,43 @@ class NicotineFrame(Window):
     def on_get_user_info(self, *_args):
         self.userinfo.on_get_user_info()
 
-    """ Browse Shares """
+    """ Shares """
 
     def on_get_shares(self, *_args):
         self.userbrowse.on_get_shares()
 
     def on_load_from_disk(self, *_args):
         self.userbrowse.on_load_from_disk()
+
+    def shares_unavailable_response(self, _dialog, response_id, _data):
+
+        if response_id == 2:  # 'Retry'
+            self.core.shares.rescan_shares()
+
+        elif response_id == 3:  # 'Force Rescan'
+            self.core.shares.rescan_shares(force=True)
+
+    def shares_unavailable(self, shares):
+
+        shares_list_message = ""
+
+        for virtual_name, folder_path in shares:
+            shares_list_message += "• \"%s\" %s\n" % (virtual_name, folder_path)
+
+        def create_dialog():
+            OptionDialog(
+                parent=self.window,
+                title=_("Shares Not Available"),
+                message=_("Verify that external disks are mounted and folder permissions are correct."),
+                long_message=shares_list_message,
+                first_button=_("_Cancel"),
+                second_button=_("_Retry"),
+                third_button=_("_Force Rescan"),
+                callback=self.shares_unavailable_response
+            ).show()
+
+        # Avoid dialog appearing deactive if invoked during rescan on startup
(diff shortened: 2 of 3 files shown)


mathiascode:
Screenshot (strings are not necessarily final):

pynicotine/cli/application.py
```diff
@@ -101,6 +101,10 @@ def on_critical_error(self, _exc_type, exc_value, _exc_traceback):
     def on_critical_error_threading(args):
         raise args.exc_value
 
+    def shares_unavailable(self, shares):
+        for virtual_name, folder_path in shares:
+            log.add("• %s (%s)" % (virtual_name, folder_path))
```
slook, line 106: `"• \"%s\" at %s % (virtual_name, folder_path)`

Suggest use " quotes instead of brackets for the virtual name, because the planned shares commands will require names to be surrounded in quotes and it's helpful to have these matching syntax.

pynicotine/gtkgui/frame.py
```diff
@@ -1468,14 +1468,45 @@ def on_update_user_info(self, *_args):
...
+
+        elif response_id == 3:  # 'Force Rescan'
+            self.core.shares.rescan_shares(force=True)
+
+    def shares_unavailable(self, shares):
+
+        shares_list_message = ""
+
+        for virtual_name, folder_path in shares:
+            shares_list_message += "• %s (%s)\n" % (folder_path, virtual_name)
```
slook, line 1492: The string in the gui might as well match the one used in the headless cli (unless there's a compelling reason for the folder_path and virtual_name to be in a different order?).

`"• \"%s\" at %s % (virtual_name, folder_path)`

pynicotine/gtkgui/frame.py
```diff
@@ -1468,14 +1468,45 @@ def on_update_user_info(self, *_args):
...
+        def create_dialog():
+            OptionDialog(
+                parent=self.window,
+                title=_("Shares Not Available"),
+                message=_("The following shares are not available. Verify that external disks are mounted and "
+                          "folder permissions are correct."),
+                long_message=shares_list_message,
+                first_button=_("_Cancel"),
+                second_button=_("_Retry"),
+                third_button=_("_Force Rescan"),
```
slook, line 1503: The Force Rescan button can and should be hidden if there is nothing available to scan, because there is no point in offering to completely wipe out the index, since the primary purpose of all this is to prevent inadvertent data loss.

`third_button=_("_Force Rescan") if shares else "",`

If the user intends to zero their index on purpose, then it will be possible for the rescan command to be run with the force argument in any case.

slook (approved):

pynicotine/gtkgui/frame.py
```diff
@@ -1468,14 +1468,45 @@ def on_update_user_info(self, *_args):
...
+
+        for virtual_name, folder_path in shares:
+            shares_list_message += "• %s (%s)\n" % (folder_path, virtual_name)
+
+        def create_dialog():
+            OptionDialog(
+                parent=self.window,
+                title=_("Shares Not Available"),
+                message=_("The following shares are not available. Verify that external disks are mounted and "
+                          "folder permissions are correct."),
```
slook, line 1499: > strings are not necessarily final

"~_The following shares are not available._~ Verify that any external disks are correctly mounted and accessible."

pynicotine/gtkgui/frame.py
mathiascode: I got the order mixed up.

pynicotine/gtkgui/frame.py
mathiascode: I'm not quite sure what you mean. The warning (in my branch) only appears if you have shared something, and at least one of the shared folders (with a virtual name) are unavailable.

If you're talking about showing a warning if a user hasn't shared any folders (0 folders, 0 files), I don't think it belongs in this dialog.

mathiascode: > Since the dialog is unaware of the total number of configured shares, then it's understandable that this complexity is not desired at the moment. This current implementation seems to satisfy the requirement nicely.

It could be added, but I'm not sure how common it would be for all shares to be unavailable at the same time. Leaning towards not adding this for now.

PR #2238, opened 2022-10-31
Result Filters: Spaced digit conditions and presets

### LINKED ISSUE #2141: Bitrate filter does not work as expected
Nicotine+ version:  3.2.2
Operating System/Distribution: Arch Linux

### Describe the bug

The "Bitrate" results filter does not seem to work as expected:

- I would expect that typing `<320` and pressing enter should hide all results with bitrates of 320 or above. But 320kbps mp3s are still visible.
- Typing `>192` and pressing enter should hide all results with bitrates of 192 or below. They are still visible though.
- Typing `>192 <320` and pressing enter should show only results with bitrates above 192 and below 320. Again, this is not what happens.

It seems that Nicotine+ is interpreting `>` as `>=` and `<` as `<=`, since `<319` seems to hide 320kbps results as expected; `>193` also seems to hide 192kbps.

However, then I would expect a query like `>193 <319` to hide 320kbps and 192kbps results. Instead, I still see them. Is it possible to filter to a specific range?

Maybe I am misunderstanding how this filter works; if so, a more informative tooltip may be helpful. I could've sworn previous versions had a better tooltip but in 3.2.2 it just says "Bitrate" when I hover my mouse over the input field.

Either way, the bitrate field has always felt a bit buggy.

### Expected behavior

- `>` and `<` operate as greater then/less than, rather than greater than or equals/less than or equals.
- Multiple (space-separated) entries in the Bitrate field should be `AND`-ed together like they are in the other filters.
- A better tooltip with bitrate filtering syntax and examples of bitrate filtering.

1. Do a search.
2. Open "Result filters".
3. Try any of the above examples.

### Additional context

Let me know if you need any other information.

slook:
Closes #2141 

+ Added: Enable ` ` (space) and `&` (ampersand) characters to seperate filter conditions (plus also `,` (comma) in country)
+ Fixed: Handle conditions where the value is an exact match but neither `!` or `=` is the operator
+ ~Changed: `<` and `>` is now mathmatically correct, i.e. less/more than instead of less/more or equal to~
+ ~Added: `<=` and `>=` operators to be specified, also allow `==` and `!=` syntax~
+ ~Fixed: Return match on exact match even if the filter isn't factorized (i.e on Bitrate or raw seconds)~
+ Changed: Result Filter Help and tooltip strings to reflect recent changes and illustrate new multiple condition features
+ Changed: Presets to inlcude ranges, moved so that most recently used item is always at top, enables easier toggling
+ ~Added: A video filetype preset. Removed: a couple of obscure filetypes and outdated items from the filter presets.~
+ Changed: Exclude non-audio files (with zero duration or bitrate) when using digit based filter (unless exact match of zero)

_Any amount of spaces or pipes can be entered between multiple filter conditions without breaking the field. One space after an operator can be entered before the digit without it being orphaned, allowing for a much more flexible syntax._

This PR is ready to be merged for user testing if desired, can revise or revert any modifications as required.


pynicotine/gtkgui/search.py
```diff
@@ -1452,20 +1448,21 @@ def on_refilter(self, *_args):
             except re.error:
                 filter_out = None
 
+        # Allow the use of seperator pipes | OR spaces & ampersands
         if filter_size:
-            filter_size = filter_size.split("|")
+            filter_size = re.sub(r"[\s&]", '|', filter_size).split("|")
```
slook, line 1453: There is a slight problem where multiple `|` characters appear if more than one seperator is used between conditions and are stored in the presets like such. For example space+"` & `"+space produces "`|||`". Empty conditions are therefore being passed into check_filter()

WIP a regex that can split better.

slook: Resolved - Split at | pipes ampersands & space(s) but don't split operators spaced before digit

`(?:[|&\s])+(?<![<>!=]\s)`

Screenshot_2022-11-05_23-45-39

pynicotine/config.py
```diff
@@ -315,10 +315,10 @@ def load_config(self):
                 "filtercc": [],
                 "filterin": [],
                 "filterout": [],
-                "filtersize": [],
-                "filterbr": [],
-                "filtertype": [],
-                "filterlength": [],
+                "filtersize": [">50MiB", ">20MiB|<50MiB", ">10MiB|<20MiB", ">5MiB|<10MiB", "<=5MiB"],
+                "filterbr": ["!0", "128|<224", ">192|<=320", "=320|=1411", ">320"],
+                "filtertype": ["flac|wav|aiff|cue|iso", "mp3|m4a|aac|ogg|opus", "mp4|m4v|mkv|mov|264", "!mp3|!wma]"],
```
mathiascode, line 320: Could you add WavPack back to the first (lossless) filter? Not sure about including iso here, is it used for anything other than commercial SACDs?

mathiascode: Shouldn't we remove pipes from these default filters?

mathiascode, line 320: Does anyone actually use the .264 extension?

mathiascode, line 320: Could you include webm in the video filters?

mathiascode:
@hboetes How do the strings in this PR look?

pynicotine/gtkgui/ui/search.ui
```diff
@@ -239,7 +239,7 @@
                               <object class="GtkEntry">
                                 <property name="width-chars">12</property>
                                 <property name="placeholder-text" translatable="yes">File type…</property>
-                                <property name="tooltip-text" translatable="yes">File type, e.g. flac|wav|ape or !mp3|!m4a</property>
+                                <property name="tooltip-text" translatable="yes">File type, e.g. flac wav !mp3 !m4a</property>
```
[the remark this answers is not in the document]
slook: My intention is to clarify that "or" is not part of the syntax, but indeed the resulting example is non-sensical.

pynicotine/config.py
```diff
@@ -315,10 +315,10 @@ def load_config(self):
                 "filtercc": [],
                 "filterin": [],
                 "filterout": [],
-                "filtersize": [],
-                "filterbr": [],
-                "filtertype": [],
-                "filterlength": [],
+                "filtersize": [">50MiB", ">20MiB|<50MiB", ">10MiB|<20MiB", ">5MiB|<10MiB", "<=5MiB"],
```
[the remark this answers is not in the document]
slook: Why are you not in support of the defaults being in config?

My primary intention is to have the most recently used item at the top of the list.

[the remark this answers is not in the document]
mathiascode: I'd like to have a few quick access filters. Once the history grows large enough, these defaults become useless for this purpose.

pynicotine/config.py
slook: The string results in a list entry that is too long to fit in the combobox dropdown.

slook: The issue will be resolved by https://github.com/nicotine-plus/nicotine-plus/pull/2383

hboetes:
> @hboetes How do the strings in this PR look?

They look great, clear, to the point, and nothing overcomplicated.

pynicotine/gtkgui/ui/popovers/searchfilterhelp.ui
```diff
@@ -427,8 +427,7 @@
                 <property name="visible">True</property>
                 <property name="xalign">0</property>
                 <property name="margin-start">12</property>
-                <property name="label" translatable="yes">Uses country codes defined by ISO 3166-2:
-    US will only show results from users with IP addresses in the United States. !GB will hide results that come from users in the United Kingdom (Great Britain).</property>
+                <property name="label" translatable="yes">Multiple countries can be specified with commas, spaces or | pipe seperators.</property>
```
mathiascode, line 430: Perhaps we should only mention spaces everywhere in the documentation (but keep support for pipes). Less information to remember, and spaces seem like the intuitive choice.

slook: It does help to simply the language since there is no need to discuss the separator characters.

pynicotine/gtkgui/search.py
```diff
@@ -556,20 +557,18 @@ def update_filter_comboboxes(self):
...
-            if presets:
-                for value in presets:
-                    widget.append_text(value)
-
-                widget.append_text("")  # Separator
-
             for value in config.sections["searches"][filter_id]:
                 widget.append_text(value)
 
+            presets = self.FILTER_PRESETS.get(filter_id)
```
mathiascode, line 563: I'd prefer to have fewer presets instead of moving them to the bottom of the list.

pynicotine/gtkgui/search.py
```diff
@@ -1503,20 +1505,26 @@ def on_refilter(self, *_args):
             except re.error:
                 filter_out = None
 
+        # Split at | pipes ampersands & space(s) but don't split operators spaced before digit
+        seperator_pattern = re.compile(r"(?:[|&\s])+(?<![<>!=]\s)")
```
mathiascode, line 1509: You can compile these patterns in a class constant instead of doing it here every time.

pynicotine/gtkgui/search.py
mathiascode: Nevermind, there are only 5 of them. I'd say that's fine to keep on the top.

pynicotine/gtkgui/search.py
```diff
@@ -300,9 +300,10 @@ class Search:
         ("archive", FileTypes.ARCHIVE)
     )
     FILTER_PRESETS = {
-        "filterbr": ("0", "128", "160", "192", "256", "320"),
-        "filtersize": (">10MiB", "<10MiB", "<5MiB", "<1MiB", ">0"),
-        "filtertype": ("audio", "image", "video", "text", "archive", "!executable", "audio|image|text")
+        "filterbr": ("!0", "128 <224", ">190 <=320", "=320 =1411", "320"),
+        "filtersize": (">50MiB", ">20MiB <=50MiB", ">10MiB <=20MiB", ">5MiB <=10MiB", "<=5MiB"),
+        "filtertype": ("audio", "image", "video", "text", "archive", "!executable", "audio image text"),
+        "filterlength": (">12:00", ">8:00 <=12:00", ">5:00 <=8:00", "!0 <=5:00", "=0")
```
mathiascode, line 306: Not sure if it's worth listing the `=0` filter by default, now that we have the generic file type filters.

Regarding `!0 <=5:00` wouldn't it be more intuitive if simply `<=5:00` worked, and files without a duration (or 0) were excluded automatically? Right now it also lists non-audio files, which feels wrong.

pynicotine/gtkgui/ui/popovers/searchfilterhelp.ui
```diff
@@ -394,7 +393,8 @@
                     <property name="visible">True</property>
                     <property name="xalign">0</property>
                     <property name="margin-start">12</property>
-                    <property name="label" translatable="yes">To exclude non-audio files use !0 in the duration filter.</property>
+                    <property name="label" translatable="yes">Non-audio files can be excluded:
```
mathiascode, line 396: `Non-audio files can be excluded` not relevant anymore, we have generic file type filters.

pynicotine/gtkgui/search.py
slook: Thank you, I didn't know you could do that.

pynicotine/gtkgui/search.py
slook: The advantage of having the most recently used at top is that it makes it much easier to toggle the filter on/off/on/off.

Not only is it easier to do for the user, but also easier to code such a function to act on the TextEntry icon handler.

pynicotine/gtkgui/search.py
slook: Agreed, what was offered here is a solution to a problem that shouldn't exist in the first place.

pynicotine/gtkgui/search.py
```diff
@@ -300,10 +300,13 @@ class Search:
         ("archive", FileTypes.ARCHIVE)
     )
     FILTER_PRESETS = {
-        "filterbr": ("0", "128", "160", "192", "256", "320"),
-        "filtersize": (">10MiB", "<10MiB", "<5MiB", "<1MiB", ">0"),
-        "filtertype": ("audio", "image", "video", "text", "archive", "!executable", "audio|image|text")
+        "filterbr": ("!0", "128 <224", ">190 <=320", "=320 =1411", "320"),
```
mathiascode, line 303: =1411 seems a bit too specific (and won't work for results from Nicotine+ clients until 3.3.0). I'd probably go with `=320` and `>320` for the last two presets.

slook: > `=320` and `>320` for the last two presets

Not sure about the last preset should be `320`?

mathiascode: The idea for >320 was a lossless filter.

pynicotine/gtkgui/search.py
mathiascode: Not sure about this yet. It's something to explore for another PR.

pynicotine/gtkgui/search.py
slook: I feel like the list of presets lacks an overall "high-quality" filter.

If one considers anything that is _320Kbps or above_ to be a minimum standard, then there is no present for that here.

slook: How about:
`        "filterbr": ("!0", "128 <=256", "192 <=320", "320 <=768", ">768"),`

mathiascode (approved):

Issue #110, opened 2020-03-05
Max number of downloads uploads sametime


Mike-Reddick:
Please implement the feature to limit the maximum number of downloads, uploads at same time (concurrent) the same feature which have the original client.

kiplingw:
Thanks @Mike-Reddick. If you'd like to submit a PR, I will review it as soon as I am able.

Mike-Reddick:
Sorry I'm not a programmer, but I do not use Nicotine just because this limitation, is very important this feature.

mathiascode:
There's a setting to change the max number of upload slots in Settings -> Transfers, but not for downloads yet.

Mike-Reddick:
@mathiascode , thanks the attention, would be fine a download limiter also

mathiascode:
This requires the introduction of locally queued downloads, i.e. downloads that are added to our own list, but not requested from the uploader yet, similar to Soulseek NS.

khabib227:
This feature would be awesome!

slook:
>  I believe this shouldn't be too difficult to implement

Whereabouts would the limiter need to be checked in the Transfers module, and what is involved to create locally queued downloads?

mathiascode:
> > I believe this shouldn't be too difficult to implement
> 
> Whereabouts would the limiter need to be checked in the Transfers module, and what is involved to create locally queued downloads?

It needs to look something like this (all downloads from users are queued remotely until the number of users reaches the download slot limit, the rest of the users are queued locally): https://github.com/mathiascode/nicotine-plus/tree/download-user-limit

slook:
> It needs to look something like this (all downloads from users are queued remotely until the number of users we've queued from reaches the download slot limit, the rest of the users are queued locally)

Why doesn't it make sense to always queue everything locally first, then send transfer requests after checking if any limit criteria is set and satisfied?

mathiascode:
> Why doesn't it make sense to always put everything into the local queue first, then wait to send transfer requests only after checking if any limit criteria is set and satisfied?

- The uploader has no way of aborting any downloads that are queued locally on the other end
- We can't get queue positions as long as downloads are queued locally

slook:
Good points, however it needs to be considered if it might seem to be unreasonably disruptive to get a queue position only to then create circumstances whereby the transfer may not begin for an unknown amount of time. This would surely be likely to lead to the peers client getting jammed up.

[...]

PR #2246, opened 2022-11-03
Implement download slot limit

mathiascode:


mathiascode:
It works, but more polish is needed. It's very wasteful with QueueUpload requests, mainly because check_download_queue isn't very clever at the moment.

pynicotine/transfers.py
```diff
@@ -2085,6 +2113,28 @@ def check_queue_upload_allowed(self, user, addr, filename, real_path, msg):
...
+        if not self.allow_new_downloads():
+            return
+
+        user = None
+
+        for download in reversed(self.downloads):
+            if download.status != "Locally Queued":
+                continue
+
+            if user is None:
```
slook, line 2129: With this condition here, only one user will get checked. Is this intended?

pynicotine/transfers.py
```diff
@@ -2085,6 +2113,28 @@ def check_queue_upload_allowed(self, user, addr, filename, real_path, msg):
...
+                user = download.user
+
+            elif user != download.user:
+                continue
```
slook, line 2133: Continuing here will result in the same user being checked again, due to the above condition.

slook:
The the downloads list includes an item which has previously encountered the "Connection timeout" status, then the checking of Locally Queued stalls.

slook:
> It works, but more polish is needed.

The feature does seem to serve the intended purpose as discussed in #110 quite well overall.

Given the complexity of transfers it inherently causes some quirky behaviours between peers that are inevitably going to lead to future issues, but given the refactoring that has been done hopefully this shall not be too difficult to maintain.

slook:
Considering the pivotal changes this makes to the upload/download queuing, it would be a good idea to rebase and merge this PR as soon as possible, otherwise it will create even more work if it becomes too divergent from the master branch.

We will only find out if there are bugs in it if the testers try it out.

> more polish is needed. It's very wasteful with QueueUpload requests...

Unless you already have some specific ideas for changes that could help to address this issue, then it would be better to refine this at a later date.

mathiascode:
It's still flawed. It can get stuck for a longer time if the uploader uses the round robin queue system.

slook:
Might that be due to the prefer_user mechanism? Perhaps it's better just to go by the natural order for now.

slook:
It's difficult to read the code methodically, I think it might be better to split transfers.py into downloads.py and uploads.py core modules, such as in https://github.com/slook/nicotine-plus/tree/transfers-downloads-uploads https://github.com/slook/nicotine-plus/commit/a56bd07606cfb4526664875a0257b077bd47362c , wherein it becomes much more easily apparent where optimizations are required.

mathiascode:
@slook I agree that this should be done eventually. Keeping track of what code is related to downloads and uploads, as well as which direction messages go and which end initiated them, is difficult.

mathiascode:
> It's difficult to read the code methodically, I think it might be better to split transfers.py into downloads.py and uploads.py core modules, such as in https://github.com/slook/nicotine-plus/tree/transfers-downloads-uploads [slook@a56bd07](https://github.com/slook/nicotine-plus/commit/a56bd07606cfb4526664875a0257b077bd47362c) , wherein it becomes much more easily apparent where optimizations are required.

Completed in https://github.com/nicotine-plus/nicotine-plus/pull/2648

slook:
This seems like it would be totally impossible to rebase now the modules are split. Shall I close this PR?

mathiascode:
Yes, I don't have time to work on this for now.

PR #2259, opened 2022-11-15
Implement event emitter for communication between components

mathiascode:
Instead of the core directly calling methods in the frontend, the frontend can listen to events emitted by the core.

As a result of this:

- The core is completely isolated from the frontend, preventing developers from accidentally depending on frontend code in the core
- The frontend no longer needs to implement every callback in specific classes, improving maintainability and allowing for more flexibility when structuring the code

### DIFF
--- a/pynicotine/__init__.py
+++ b/pynicotine/__init__.py
@@ -112,7 +112,6 @@ def rescan_shares():
     from pynicotine.logfacility import log
 
     config.load_config()
-    core.init_components()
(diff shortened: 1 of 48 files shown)


slook:
Invisible Downloads / Uploads histories, and Private Chat tabs not loaded.

mathiascode:
Transfers and chats haven't been fully migrated to the event system yet.

slook:
It does load up alright now

mathiascode:
Here goes nothing!

PR #2266, opened 2022-11-26
Bump version to 4.0.0.dev1

mathiascode:
Due to the move to GTK 4 by default (which is essentially completely different than GTK 3 due to various rewrites and the move to GPU rendering), larger changes in the Nicotine+ core, new plugin command system, and other planned changes, I think bumping the major version would be a good idea: https://semver.org/

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,8 +1,8 @@
-nicotine (3.3.0-dev4) jammy; urgency=medium
+nicotine (4.0.0-dev1) jammy; urgency=medium
 
   * Development version.
 
- -- Mat <mail@mathias.is>  Fri, 25 Nov 2022 21:05:54 +0300
+ -- Mat <mail@mathias.is>  Sat, 26 Nov 2022 12:02:15 +0300
 
 nicotine (3.2.6-1) jammy; urgency=medium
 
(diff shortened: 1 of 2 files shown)


slook:
> other planned changes

The version shouldn't be bumped until those changes are actually implemented.

mathiascode:
> The version shouldn't be bumped until those changes are actually implemented

Fair enough.

> and even then, the differences visible to most end users probably arn't going to be that significant.

Yes, no significant visual differences. However, I'm not confident that GTK 4 will work without breakage on every system out there (especially older hardware) only two years after the initial GTK 4 release, since some people like to stick to old LTS distro releases.

slook:
> not confident that GTK 4 will work without breakage

The Nicotine+ major version number should not be bumped until GTK 3 support is completely removed from Nicotine+ in however many years, or perhaps when Nicotine+ will only work with Python 4.x.

Meanwhile, we will need a robust fallback mechanism or other user accessible means of selecting GTK 3 before/during runtime.

mathiascode:
I'll revisit this when we get closer to a possible release.

mathiascode:
>until GTK 3 support is completely removed

If we want to support Debian 11 ELTS, June 2026.

> We don't want new users to think that GTK 4 is a runtime requirement.

No, but GTK 4 will be the default and prioritized version moving forward.

mathiascode:
The next release won't contain as many user-facing or breaking changes as originally estimated, so I'm abandoning this idea.

PR #2271, opened 2022-11-30
Remove command alias system

mathiascode:
The command alias system predates the plugin system, and was designed for running external processes/scripts to fetch content to display in a chat message, e.g. the current system time. The ability to run external processes/scripts was removed a while back due to security concerns, leaving us with a system that can send static chat messages and run other Nicotine+ commands.

Implementation-wise, the alias system sits in an awkward position, competing with the new plugin command system and conflicting with planned features such as command completions and descriptions for all commands. The code is also awkward to understand and maintain.

If the functionality the alias system provided is desired in the future, I'd rather have it implemented in a much simpler way as a plugin, instead of belonging to the Nicotine+ core.


slook:
The detail in the commit message does explain why I never found any useful case for the command alias system.

I think objections will be unlikely.

mathiascode:
In hindsight, I should've removed the system back when I removed the possibility of running external processes. There was no code documentation though, and I didn't fully understand the purpose of the feature until much later when I read the old user manual (from 2006).

PR #2279, opened 2022-12-13
downloads.py and uploads.py: Select file in the manager if possible

slook:
+ Added: Select a single file on Open in File Manager action if a 'File Manager command' handler is configured

Only tested on linux with `thunar $`. Ideally, such behaviour would be the default on all platforms, but executing shell commands is a pickle at the moment. A more unified approach to this is required.

This functionality is very convenient especially if the folder contains a lot of files.


mathiascode:
> Only tested on linux with `thunar $`

Have you tested with no file manager specified in Nicotine+ (system default)?

mathiascode:
Ah, right, I forgot that both folders and files are opened through the same API. I'll have to see if anything can be done to highlight files when opening folders in this case.

slook:
Some file managers have a `--select` command line argument, but I don't think doing a multi-selection warrants such specific effort, the focus here was to deal with cases where a single file is referenced.

slook:
Closing, since I have forgotten what way it is supposed to work for it to be properly rebased.

PR #2281, opened 2022-12-15
core_commands: move all commands into new plugin

slook:
+ ~Added: `/help` command with fuzzy query search functionality~
+ ~Added: `/rescan` command optional `force` flag argument, command also available in headless mode~
+ ~Added: Headless mode core commands for users and network filters~
**_TODO:_** share commands need core implementation)
+ ~Changed: Make the `[user]` arguments optional when executing in a private chat interface (except for `/usearch`)~
+ Added: Accept "long strings with spaces" as quoted arguments, but don't require quotes where argument is unambiguous
+ ~Changed: Don't switch to the tab on entering `/msg` command, because that's what the /pm command is for~
+ ~Changed: Don't clear the gui input text entry unless the command is successfully consumed~
+ ~Changed: `/ban` and `/ignore` commands can accept either username or IPv4 address as argument to add appropriate filter~
+ ~Changed: `/unban` and `/unignore` commands removes both username-based and IP-based filters at the same time~
+ ~Changed: `/ip` command get address from core and print to output if available, fallback to request only if user is unknown~
+ ~Added: `/ip` command can accept IPv4 address as argument to lookup known addresses and output associated username~

- Removed: `/ctcpversion` command + Added: `/ctcp` command with `version` sub-command
- ~Removed: `/ignoreip` command (was undocumented) + Changed: `/ignore` command to also accept an IPv4 address~
- ~Removed: `/toggle` command + Added: `/plugin` command with `toggle` `enable` `disable` sub-commands~
- ~Removed: `/p` and `/part` aliases from the `/leave` chat room command~
- ~Removed: `/ad` alias from the `/add` buddy command~

+ ~Fixed: Zero "`0.0.0.0`" IP addresses returned by GetPeerAddress() could overwrite another unrelated pending request~
----
_Headless `help` command output:_
```
help
Listing 16 available commands:

Nicotine+ Commands:
	/help, /? [query]	-  List available commands
	/away, /a	-  Toggle away status
	/quit, /q, /exit [-force]	-  Quit Nicotine+

Configure Shares:
	/rescan [-force]	-  Rescan shares
	/shares, /ls [public] [buddy]	-  List shares
	/share <public|buddy> <folder path>	-  Add share
	/unshare <public|buddy> <virtual name>	-  Remove share

core_commands:
	/hello, /echo, /greet [something..]	-  Print something

Now Playing:
	/now	-  Display the Now Playing script's output

Users:
	/add, /buddy <user>	-  Add user to buddy list
	/rem, /unbuddy <buddy>	-  Remove user from buddy list

Network Filters:
	/ip <user or ip>	-  Show IP address of user or username from IP
	/ban <user or ip>	-  Stop connections from user or IP address
	/unban <user or ip>	-  Remove user or IP address from ban lists

Plugin Commands:
	/plugin <toggle|enable|disable> <plugin_name>	-  Load or unload a plugin

Port Checker:
	/port <user>	-  Check firewall state of user

Type /help [query] to list similar commands

```

----
_Chat `/help` command output:_
```
Listing 33 available commands:

Nicotine+ Commands:
	/help, /? [query]	-  List available commands
	/away, /a	-  Toggle away status
	/quit, /q, /exit [-force]	-  Quit Nicotine+

Configure Shares:
	/rescan [-force]	-  Rescan shares
	/shares, /ls [public] [buddy]	-  List shares
	/share <public|buddy> <folder path>	-  Add share
	/unshare <public|buddy> <virtual name>	-  Remove share

core_commands:
	/hello, /echo, /greet [something..]	-  Print something

Chat:
	/clear, /cl	-  Clear chat window
	/ctcp <version|ping> [user]	-  Send client-to-client protocol query
	/me <something..>	-  Say something in the third-person

Chat Rooms:
	/join, /j <room>	-  Join chat room
	/say <room> <message..>	-  Say message in specified chat room
	/leave, /l [room]	-  Leave chat room

Private Chat:
	/msg, /m <user> <message..>	-  Send private message to user
	/pm <user>	-  Open private chat window for user
	/close, /c <user>	-  Close private chat

Now Playing:
	/now	-  Display the Now Playing script's output

Users:
	/add, /buddy <user>	-  Add user to buddy list
	/rem, /unbuddy <buddy>	-  Remove user from buddy list
	/info, /whois, /w <user>	-  Show user profile information and interests
	/browse, /b <user>	-  Browse files of user

Network Filters:
	/ip <user or ip>	-  Show IP address of user or username from IP
	/ban <user or ip>	-  Stop connections from user or IP address
	/unban <user or ip>	-  Remove user or IP address from ban lists
	/ignore <user or ip>	-  Silence chat messages from user or IP address
	/unignore <user or ip>	-  Remove user or IP address from chat ignore lists

Search Files:
	/search, /s <query>	-  Start global file search
	/rsearch, /rs <query>	-  Search files in joined rooms
	/bsearch, /bs <query>	-  Search files of all buddies
	/usearch, /us <"user name"> <query>	-  Search a user's shared files

Plugin Commands:
	/plugin <toggle|enable|disable> <plugin_name>	-  Load or unload a plugin

Port Checker:
	/port <user>	-  Check firewall state of user

Type /help [query] to list similar commands
```


pynicotine/plugins/core_commands/__init__.py
```diff
@@ -26,39 +26,637 @@ def __init__(self, *args, **kwargs):
...
+                "usage_private_chat": ["[user]"]
+            },
+            "unban": {
+                "callback": self.unban_user_command,
+                "description": _("Remove user from ban list"),
+                "group": _("Users"),
+                "usage": ["<user>"],
+                "usage_private_chat": ["[user]"]
+            },
+            "block": {  # new
```
mathiascode, line 170: Suggest ban-ip

pynicotine/plugins/core_commands/__init__.py
```diff
@@ -26,39 +26,637 @@ def __init__(self, *args, **kwargs):
...
+                "callback": self.block_user_ip_command,
+                "description": _("Stop all connections from same IP as user"),
+                "group": _("Network Filters"),
+                "usage": ["<user>"],
+                "usage_private_chat": ["[user]"]
+            },
+            "unblock": {  # new
```
mathiascode, line 177: Suggest unban-ip

pynicotine/plugins/core_commands/__init__.py
```diff
@@ -26,39 +26,637 @@ def __init__(self, *args, **kwargs):
...
+            },
+            "unignore": {
+                "callback": self.unignore_user_command,
+                "description": _("Remove user from chat ignore list"),
+                "disable": ["cli"],
+                "group": _("Users"),
+                "usage": ["<user>"],
+                "usage_private_chat": ["[user]"]
+            },
+            "ignoreip": {
```
mathiascode, line 200: ignore-ip

pynicotine/plugins/core_commands/__init__.py
```diff
@@ -26,39 +26,637 @@ def __init__(self, *args, **kwargs):
...
+                "callback": self.ignore_user_ip_command,
+                "description": _("Silence chat messages from IP address of user"),
+                "disable": ["cli"],
+                "group": _("Network Filters"),
+                "usage": ["<user>"],
+                "usage_private_chat": ["[user]"]
+            },
+            "unignoreip": {  # new
```
mathiascode, line 208: unignore-ip

pynicotine/plugins/core_commands/__init__.py
```diff
@@ -26,39 +26,637 @@ def __init__(self, *args, **kwargs):
...
+                "callback": self.unignore_user_ip_command,
+                "description": _("Remove user's IP address from chat ignore list"),
+                "disable": ["cli"],
+                "group": _("Network Filters"),
+                "usage": ["<user>"],
+                "usage_private_chat": ["[user]"]
+            },
+            "ipignore": {  # new
```
mathiascode, line 216: Having both a ignoreip and ipignore command that do different things is very confusing.

pynicotine/plugins/core_commands/__init__.py
```diff
@@ -26,39 +26,637 @@ def __init__(self, *args, **kwargs):
...
+            },
+            "usearch": {
+                "aliases": ["us"],
+                "callback": self.search_user_command,
+                "description": _("Search a user's shared files"),
+                "disable": ["cli"],
+                "group": _("Search Files"),
+                "usage": ["<\"user name\">", "<query>"]
+            },
+            "listshares": {
```
mathiascode, line 287: list-shares

pynicotine/plugins/core_commands/__init__.py
```diff
@@ -26,39 +26,637 @@ def __init__(self, *args, **kwargs):
...
+                "callback": self.list_shares_command,
+                "description": _("List shares"),
+                "group": _("Shares")
+            },
+            "addshare": {
```
mathiascode, line 292: share-folder (perhaps?)

pynicotine/plugins/core_commands/__init__.py
```diff
@@ -26,39 +26,637 @@ def __init__(self, *args, **kwargs):
...
+                "callback": self.add_share_command,
+                "description": _("Add share"),
+                "group": _("Shares"),
+                "usage": ["<public|private|buddy>", "<\"virtual name\">", "<\"folder path\">", ""]  # "" max 3 args
+            },
+            "removeshare": {
```
mathiascode, line 298: unshare-folder (perhaps?)

pynicotine/plugins/core_commands/__init__.py
```diff
@@ -26,39 +26,637 @@ def __init__(self, *args, **kwargs):
...
+                "description": "Print something",
+                "group": _("Message"),
+                "usage": ["[something..]"]
+            },
+            "away": {
+                "aliases": ["a"],
+                "callback": self.away_command,
+                "description": _("Toggle away status")
+            },
+            "plugin": {
```
mathiascode, line 53: I think I'd prefer separate commands here.
- toggle-plugin
- enable-plugin
- disable-plugin

pynicotine/plugins/core_commands/__init__.py
```diff
@@ -26,39 +26,637 @@ def __init__(self, *args, **kwargs):
...
+                "disable": ["cli"],
+                "group": _("Chat"),
+                "usage": [""]  # "" disallow any args
+            },
+            "ctcp": {
+                "callback": self.ctcp_command,
+                "description": _("Send client-to-client protocol query"),
+                "disable": ["cli"],
+                "group": _("Chat"),
+                "usage": ["<version|time|ping>", "[user]"]
```
mathiascode, line 77: I don't like the idea of sending times. I'm concerned it would expose your approximate country when behind a VPN.

pynicotine/plugins/core_commands/__init__.py
slook: What about a single `share` command with `add` and `remove`/`rem` and `list` sub-commands?

pynicotine/plugins/core_commands/__init__.py
slook: If the input looks like an IP address, then we could call the relevant function accordingly. The only downside would be that you wouldn't be able to ban a user whose username is like an ip address.

pynicotine/plugins/core_commands/__init__.py
mathiascode: I'm concerned about adding too many arguments to commands, and the help message becomes confusing.

pynicotine/plugins/core_commands/__init__.py
```diff
@@ -26,39 +26,637 @@ def __init__(self, *args, **kwargs):
...
+            "listshares": {
+                "callback": self.list_shares_command,
+                "description": _("List shares"),
+                "group": _("Shares")
+            },
+            "addshare": {
+                "callback": self.add_share_command,
+                "description": _("Add share"),
+                "group": _("Shares"),
+                "usage": ["<public|private|buddy>", "<\"virtual name\">", "<\"folder path\">", ""]  # "" max 3 args
```
mathiascode, line 296: What is "private"?

pynicotine/plugins/core_commands/__init__.py
slook: My intention is to move CTCP functions into a core_commands module (ctcp_commands.py), and expose it's options in a Core Commands plugin settings settings dialog. The replies can be handled in there with an incomming_chat_event() too.

pynicotine/plugins/core_commands/__init__.py
slook: that's what you originally put in your first cli-commands-test draft.

pynicotine/plugins/core_commands/__init__.py
slook: The enable|disable functions act more like load and unload, because the setting is not permanently saved.

pynicotine/plugins/core_commands/__init__.py
mathiascode: I probably intended to put buddy.

pynicotine/plugins/core_commands/__init__.py
slook: Thats very true. Is there anything that prevents us from having commands containing spaces, such that each sub-command could then have it's own documention?

pynicotine/plugins/core_commands/__init__.py
slook: Do you know how the core functions are going to be implemented? At the moment, adding and removing shares is not callable as they are handled by either preferences.py or setup assistant.

pynicotine/plugins/core_commands/__init__.py
slook: If it were a seperate function, then it would be disabled by default and users would be unlikely to enable it. We dont want the core_commands to be disableable.

pynicotine/plugins/core_commands/__init__.py
slook: Do you know how the core functions are going to be implemented? At the moment, adding and removing shares is not callable as they are handled by either preferences.py or setup assistant.

pynicotine/plugins/core_commands/__init__.py
slook: I have opted for combined `/ignoreip` and `/unignoreip` commands that accept either `username` or `ip address` as an argument.

The `is_ip_address()` staticmethod used to do this could be placed into utils.py to be used from preferences.py and networkfilter.py in several places where we need to check if a string is an IPv4 address format or not.

pynicotine/plugins/core_commands/__init__.py
slook: This command does not (currently) support accepting an IP address as argument, so the proposed name would be misleading.

pynicotine/plugins/core_commands/__init__.py
slook: This command does not (currently) support accepting an IP address as argument, so the proposed name would be misleading.

pynicotine/plugins/core_commands/__init__.py
slook: The name matches the original spec of the existing command as it was before.

pynicotine/plugins/core_commands/__init__.py
slook: The name of this new experimental command is based on the same naming convention as historically used.

It will probably be removed for now because implementing it requires significant effort that is beyond the scope of this PR.

pynicotine/plugins/core_commands/__init__.py
slook: The name has not changed since the name is specified as it was when you first wrote the command definitions.

In this case `unshare` would be enough.

pynicotine/plugins/core_commands/__init__.py
slook: `shares`, `ls`

pynicotine/plugins/core_commands/__init__.py
mathiascode: Yes, but I disagree with my past choices here.

pynicotine/plugins/core_commands/__init__.py
mathiascode: I suppose we can use subcommands if they are simple enough.

pynicotine/plugins/core_commands/__init__.py
mathiascode: I've moved away from the "block" terminology in the GUI over the past year or so.

pynicotine/plugins/core_commands/__init__.py
mathiascode: Let's hope our commands remain short enough then.

slook: Agreed, it's not reasonable to expect the user to deal with a whole range of commands each with subtle differences.

It is going to be necessary to allow for more flexible arguments, even if this making some assumptions, such as:

- If we specify an `ip_address`, then the intention is to ignore/unignore all usernames associated with that address.
- If we specify a `username`, then the intention may be inferred also to ignore/unignore all users at that same address.

The `<user>` argument now validates on either a "username" or an "ip_address" type of input for both the `/ignore` and `/unignore` commands, but a core function is missing from networkfilter.py to do unignore by ip address.

pynicotine/plugins/core_commands/__init__.py
slook: Agreed, it's not reasonable to expect the user to deal with a whole range of commands each with subtle differences.

It is going to be necessary to allow for more flexible arguments, even if this involves making some assumptions, such as:

    If we specify an ip_address, then the intention is to ban/unban all usernames associated with that address.
    If we specify a username, then the intention may be inferred also to ban/unban all users at that same address.

The <user> argument now validates on either a "username" or an "ip_address" type of input for both the `/ban` and `/unban `commands, but a core functionality is missing from networkfilter.py to enable ban/unban without knowing a username (yet it should be possible because `ipblocklist` does accept a zero-length string for the username, and perhaps we can try look up an associated username for `banlist` if we have an active/current connection).

slook: Resolved, both username-based and IP-based bans now get removed with a single `/unban` command.

pynicotine/plugins/core_commands/__init__.py
slook: Resolved, both username-based and IP-based ignore rules now get removed with a single `/unignore` command.

pynicotine/plugins/core_commands/__init__.py
```diff
@@ -26,39 +26,633 @@ def __init__(self, *args, **kwargs):
...
+            user = args
+
+        # Remove username-based filter as well as IP-based filter if exists
+        unignored_ip_address = self.core.network_filter.unignore_user_ip(user)
+        self.core.network_filter.unignore_user(user)
+
+        self.output(_("Unignored %s") % (unignored_ip_address or user))
+
+    @staticmethod
+    def is_ip_address(ip_address):
```
mathiascode, line 532: You could move this function to networkfilter.py.

pynicotine/networkfilter.py
```diff
@@ -60,27 +60,33 @@ def _request_ip(self, user, action, list_type):
...
+            # Try to get a username from currently active connections
+            user = self.get_known_username(ip_address) or ""
+
+        elif user and not ip_address:
+            # Try to get an address from currently active connections
+            ip_address = self.get_known_ip_address(user)
 
-        ip_address, _port = core.user_addresses[user]
+            if ip_address is None:
+                return self._request_ip(user, "add", list_type)
```
mathiascode, line 80: Return value should be an IP address (or None), but you're returning a boolean value here.

mathiascode:
> /ignoreip command (was undocumented)

I've never noticed this until now. Wonder why someone added a single IP-related command and never listed it in the help window.

I'm glad the command help popover will soon be generated dynamically, so we don't have to deal with these kind of issues anymore.

pynicotine/networkfilter.py
slook: Yes, good point I wasn't really sure what to do in this case, handling of offline users in here certainly isn't ideal.

For now, we return a username, this at least avoids a boolean being expressed in the output for unknown/offline users, albeit at the expense of code readability.

slook:
> Wonder why someone added a single IP-related command and never listed it

As it was before, the module didn't keep very good track of IP addresses, it seems that the IP-based filtering feature was probably added after the username-based mechanism.

> popover will soon be generated dynamically

That would be wonderful, I was going to suggest removing the popover and make the button be a shortcut for the text-based `/help` command. Generating dynamic content would certainly warrent help_command() being it's own help_command.py module.

pynicotine/networkfilter.py
slook: We can check if the return value `is_ip_address()`, this is only used in ban_user_ip() to close a banned connection. 

Beforehand, I don't think we even checked it wasn't None until the CloseConnectionIP() message to see if any of the conn_obj items were equal to None in the core (or having wildcard values containing `*` asktricks), so the return chain didn't really work properly anyway, and did result in some message pack error on certain inputs. That issue is solved now.

If you can think of a better way of doing this then feel free to change it or suggest something, this is the best I can come up with.

slook: > Return value should be an IP address (or None)

It's more desirable to press ahead with creating the filter with a dummy address, because then this allows it to be updated by the callback when the user comes on-line (or when reconnecting if the filter is added whilst offline).

No return is needed here at all @mathiascode

mathiascode:
> As it was before, the module didn't keep very good track of IP addresses, it seems that the IP-based filtering feature was probably added after the username-based mechanism.

I added the networkfilter module about two years ago, but it was essentially a rewrite of scattered ban/ignore-related code in Nicotine+. I wouldn't be surprised if some things weren't ideal.

https://github.com/nicotine-plus/nicotine-plus/commit/e12ce3aea46ebfd59e34aa5e83f8a81a3507095b

> It going to be interesting to see how you transpose output() into XML tags, good luck with that! Suppose we could just bung the whole raw output into a text_view widget, but that's a topic for another time.

No need to use GtkBuilder and XML, I'll just create GTK label widgets in Python code.

> How should we proceed with listing, adding and removing the shares for the `ls`, `share` and `unshare` commands?

Another PR.

slook:
>> How should we proceed with listing, adding and removing the shares for the ls, share and unshare commands?

> Another PR.

Done in #2313

PR #2287, opened 2022-12-22
networkfilter.py: support explicit ip addresses and handle offline users

slook:
This PR includes what is needed to support the user filtering related CLI core_commands being developed in #2281

+ Added: Support for adding IP addresses directly into the IP Ban/Ignore filters, without needing to know a username
+ Added: Try to establish a username, if known, whilst adding a IP ban/ignore filter in the Preferences or `/ignoreip` command

As a result of the intended implementation, it has been necessary to address the following issues:

+ Changed: Prefer locally stored `core.user_addresses` where possible to avoid unnecessary excessive GetPeerAddress() calls
+ Fixed: Several issues that made it impossible to reliably ban/ignore more than one offline user at a time
+ Fixed: Prevent duplicate username lookups for improved performance when getting and setting cached ban/ignore states
+ Fixed: Don't replace a valid filter with zero `0.0.0.0` IP address or `""` empty username data if the user is offline
+ Fixed: Don't try to close zero 0.0.0.0 IP connections when adding an IP-ban for an offline user, or closing Preferences dialog
+ Fixed: Avoid core error when attempting to close newly added banned IP address containing `*` wildcard characters

### DIFF
--- a/pynicotine/gtkgui/dialogs/preferences.py
+++ b/pynicotine/gtkgui/dialogs/preferences.py
@@ -858,22 +858,13 @@ class IgnoredUsersPage:
 
         ip_address = dialog.get_entry_value()
 
-        if not ip_address or ip_address.count(".") != 3:
+        if not core.network_filter.is_ip_address(ip_address):
             return
 
-        for chars in ip_address.split("."):
-            if chars == "*":
-                continue
-
-            if not chars.isdigit():
-                return
-
-            if int(chars) > 255:
-                return
-
         if ip_address not in self.ignored_ips:
-            self.ignored_ips[ip_address] = ""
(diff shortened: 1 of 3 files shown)


mathiascode:
I'll review this as soon as possible.

mathiascode:
Changes are fine in general. Not sure if I want the placeholder stuff, but I'll open a separate PR.

slook:
Thank you. I agree the pending placeholder stuff is not the most graceful to look at, but it does work better than before.

I really dont like the `?.?.?.?` question marks, I can't remember exactly why `0.0.0.0` was unsuitable, there was some sort of conflict somewhere.

If you can come up with a better way of tracking offline users who are pending to be banned/unbanned/ignored/unignored, then feel free to rip into it.

PR #2288, opened 2022-12-22
core_commands: rescan and shares list commands

slook:
+ Added: `/rescan` command optional `force` flag argument, command is now also available in headless mode
+ Added: `/shares` command (with `/ls` alias) to list public, buddy or all configured shares


pynicotine/shares.py
```diff
@@ -838,16 +838,21 @@ def process_scanner_messages(self, scanner, scanner_queue):
         return False
 
     def check_shares_available(self):
+        return self.list_shares(only_unavailable=True)
+
+    def list_shares(self, only_unavailable=False):
```
mathiascode, line 843: This method seems somewhat unnecessary. I'd simply use get_shared_folders() in the plugin command.

slook: Done

PR #2294, opened 2022-12-30
Fix bug where path is null for certain downloads

jackBonadies:
On 3.3.0-dev4, there is a bug where the transfer.path is only set when one uses the "Download Files / Folders to..." option.  If one does not, then the transfer.path will be None.  This causes a problem where, when right clicking on a transfer > "Open in File Manager" it opens to the root download folder (rather than the downloaded folder name or the username if the 'store completed to username subfolders' setting is on).

The fix I have made for this is to set the transfer.path to the final saved folder that we move the transfer to when it finishes downloading.  If the path is already set (in the case of "Download Files / Folders to..." option) it changes nothing.

Please feel free to modify or discard this pull request if there is a different fix in mind.

### DIFF
--- a/pynicotine/transfers.py
+++ b/pynicotine/transfers.py
@@ -2034,6 +2034,7 @@ class Transfers:
         transfer.current_byte_offset = transfer.size
         transfer.sock = None
         transfer.token = None
+        transfer.path = folder
 
         core.statistics.append_stat_value("completed_downloads", 1)
 


mathiascode:
I've tried to improve the situation in 7543ad5771f5d0f2797db8f20fc9248fbdd0e928 and 3fbcb01719f238fbf3e2ff1b137c8e02e25a9bd7. Could you test master and verify that it behaves as you would expect?

jackBonadies:
Just tested it out and yes those 2 commits fix the behavior.

> As far as I know, the path not being set for non-folder downloads was intentional. It doesn't make much sense though.

I couldn't see any side effects of always setting it.  The problem with not setting it is that one loses the default directory at the time of download.

PR #2299, opened 2023-01-02
Prioritize user profiles in user actions

mathiascode:
The idea is that the user description becomes more prominent, in order to learn more about a user before interacting with them. User profiles contain buttons to then perform various actions, like sending a private message.

- Move user profile item to the top of user menus
- Double-clicking usernames in chatroom list opens user profile
- Rename user info -> user profile

### DIFF
--- a/pynicotine/gtkgui/widgets/trayicon.py
+++ b/pynicotine/gtkgui/widgets/trayicon.py
@@ -1,4 +1,4 @@
-# COPYRIGHT (C) 2020-2022 Nicotine+ Contributors
+# COPYRIGHT (C) 2020-2023 Nicotine+ Contributors
 #
 # GNU GENERAL PUBLIC LICENSE
 #    Version 3, 29 June 2007
@@ -102,8 +102,8 @@ class BaseImplementation:
         self.create_item()
 
         self.send_message_item = self.create_item(_("Send Message"), self.on_open_private_chat)
-        self.lookup_info_item = self.create_item(_("Request User's Info"), self.on_get_a_users_info)
-        self.lookup_shares_item = self.create_item(_("Request User's Shares"), self.on_get_a_users_shares)
+        self.lookup_info_item = self.create_item(_("View User Profile"), self.on_get_a_users_info)
+        self.lookup_shares_item = self.create_item(_("Browse Shares"), self.on_get_a_users_shares)
 
         self.create_item()
 
@@ -232,8 +232,8 @@ class BaseImplementation:
 
         EntryDialog(
             parent=self.application.window,
-            title=_("Request User Info"),
-            message=_('Enter the name of the user whose info you want to see:'),
+            title=_("View User Profile"),
+            message=_('Enter the name of the user whose profile you want to see:'),
             callback=self.on_get_a_users_info_response,
             droplist=sorted(core.userlist.buddies)
         ).show()
@@ -252,7 +252,7 @@ class BaseImplementation:
 
         EntryDialog(
             parent=self.application.window,
-            title=_("Request Shares List"),
+            title=_("Browse Shares"),
             message=_('Enter the name of the user whose shares you want to see:'),
             callback=self.on_get_a_users_shares_response,
             droplist=sorted(core.userlist.buddies)
--- a/pynicotine/userinfo.py
+++ b/pynicotine/userinfo.py
@@ -1,4 +1,4 @@
-# COPYRIGHT (C) 2021-2022 Nicotine+ Contributors
+# COPYRIGHT (C) 2021-2023 Nicotine+ Contributors
 #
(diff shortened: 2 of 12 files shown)


slook:
> Add button to show user profile when browsing shares

This button seems mis-placed here.

The toolbar might as well be a single full-width toolbar, rather than one for each pane. A "Download File" button (or "Upload File" button for self-browse; or "Down/Up-load Folder" for folder selection) would be more suited here instead.

mathiascode:
> > Add button to show user profile when browsing shares
> 
> This button seems mis-placed here.

What do you suggest? I'd like to link the browse page with the user info page somehow, but I haven't found any way that feels right.

> The toolbar might as well be a single full-width toolbar, rather than one for each pane. The search text entry should be centered rather than aligned to one of the treeview widgets, because its function does apply to both folders and files.

It would fix the search bar, but the expand/collapse all button feels awkward if it's positioned all the way to the right.

> A "Download File" button (or "Upload File" button for self-browse; or "Down/Up-load Folder" for folder selection) would be more suited here instead. Possibly considering also, label widgets for the search position, selected file count, total selected size, duration etc. If such items were to be added to your proposed full-width toolbar in future, then shortcut buttons for unrelated peripheral functions might be more confusing than helpful.

Such actions would have to go into a bottom bar, if we ever add one in the future.

mathiascode:
Trying out some ideas:

slook:
> some ideas

The layout is heading in the right direction, but the "All Info" button label doesn't align with your proposed "profile" terminology.

> What do you suggest?

I don't have any concrete suggestions, I have considered all sorts but it's difficult to come up with a layout that doesn't have downsides in one way or another.

> It would fix the search bar, but the expand/collapse all button feels awkward if it's positioned all the way to the right.

What about on the left? What about considering a collapsible Search bar (somewhat like the Result Filters toolbar)?

> Such actions would have to go into a bottom bar, if we ever add one in the future.

I'd be concerned about this taking up excessive amounts of vertical screen real-estate, a permanent secondary toolbar would have to be heck of a useful for it to warrant the space.

hboetes:
My suggestion would be to add a 'user info'  and 'chat user' button in the tab bar on the right. So in the same row as the 'cptblack' but on the right. Similar buttons would also be nice in the info and private chat tabs.

This will make it a lot easier to jump to the information most often used. Currently you'd have to right-click and then select the item you want to see.

slook:
> button in the tab bar on the right

Afaik the tab bar area is not viable for placing UI elements, whilst we have hacked the tab widget to provide navigation for new message notifications in chats, such an implementation would be problematic as the widget is provided by the Gtk library.

mathiascode:
> > It would fix the search bar, but the expand/collapse all button feels awkward if it's positioned all the way to the right.
> 
> What about on the left? What about considering a collapsible Search bar (somewhat like the Result Filters toolbar)?

Long-term it would be nice to have the search result filters in the shares browse view too.

> I'd be concerned about this taking up excessive amounts of vertical screen real-estate, a permanent secondary toolbar would have to be heck of a useful for it to warrant the space on a low res widescreen aspect display. Therefore, we should aim to fit everything needed on a single toolbar if at all possible.

Yeah, I think leaving out download/upload buttons would be the best idea.

> perhaps it is better to merge this PR with no changes to the Browse Shares toolbar(s) for now.

Agreed, the other changes are already a nice improvement. We'll see where we go from there.

mathiascode:
Moved the layout changes to https://github.com/mathiascode/nicotine-plus/tree/user-profiles-old

@hboetes Do the strings in the PR look good to you?

pynicotine/gtkgui/widgets/trayicon.py
```diff
@@ -232,7 +232,7 @@ def on_get_a_users_info(self, *_args):
 
         EntryDialog(
             parent=self.application.window,
-            title=_("Request User Info"),
+            title=_("View User Profile"),
             message=_('Enter the name of the user whose info you want to see:'),
```
slook, line 236: `'Enter the name of the user whose info you want to see:'`
 -> 
`'Enter the name of the user whose profile you want to see:'`

pynicotine/userinfo.py
```diff
@@ -118,7 +118,7 @@ def _user_info_request(self, msg):
         self.requested_info_times[user] = request_time
 
         if core.login_username != user:
-            log.add(_("User %(user)s is reading your user info"), {'user': user})
+            log.add(_("User %(user)s is viewing your user profile"), {'user': user})
```
slook, line 121: Potentially ambiguous and superfluous use of the word "user" here.

`"User %(user)s is viewing your profile"`

hboetes:
> @hboetes Do the strings in the PR look good to you?

@slook already found two nits, with which I agree, apart from that it's looking good.

PR #2302, opened 2023-01-05
Version 3.2.8

mathiascode:


slook:
Copyright header comments and about strings etc shall need to be updated to `2023`.

PR #2313, opened 2023-01-09
core_commands: share and unshare commands

slook:
+ Added: `/share` command, with validation to check `folder_path` exists, is able to switch _public <> buddy_ share groups
+ Added: `/unshare` command, matches `virtual_name` or full `folder_path` in any `share_group`, does work easily
+ Fixed: Don't allow virtual names to contain trailing spaces or quotes when editing
+ Fixed: Allow public share to be added in Setup Assistant even if existing buddy-only share needs remapping to do it
+ Changed: Don't forget a custom virtual name if an already existing shared folder path swaps groups or is added anew
+ Added: Clearing the dialog text entry sets the virtual name to default based on folder path base name
+ Changed: Improved handling of adding duplicate folder paths, only allow one share per folder, swap groups as appropriate

The intention of moving share related code into the core is to reduce code duplication whilst removing dependency on gtkgui modules for these functions. I wanted to take this opportunity to de-duplicate and prevent any future need for code duplication.

The code works but it lacks clarity so feedback suggestions etc are welcomed. Review points to highlight:

- It is desirable that the `/share` command can be used to switch an existing shared folder to another share group. For this to work as desired some changes to handling of duplicate share entries were necessary. It would be very awkward to have to do `/unshare` first, albeit not really the end of the world that warrants so much complexity in and of itself.

- It is possible that the `/share` command will be most used for assigning existing shares into another share group, rather than simply for adding new shared folders, especially considering the name of the "share" command implies that it should be capable of doing this.

- For free we get to fix some silent fail type bugs in the GUI dialogs too, which makes everything more friendly to use in general when adding folder paths that are already shared.

- It does not seem likely to be a valid use case for having the same folder path configured in multiple groups. In the case of a different virtual name, remove the old share to result in a rename operation being done.

- The new code uses `group_index` rather than `group_name` to internally assign "public" or "buddy" groups, respectively. Not much needs to be changed here to enable more share groups, such as "trusted" and any number of custom user groups could be possible.

- A minor issue exists when adding an already existing folder path using the Preferences dialog, the old entry remains in the list until the dialog is re-opened.

These commands were complex to implement cleanly, because the Preferences module uses it's own shared folder lists, which is useful if the dialog is destroyed with Cancel. It could be argued if the ability to revert is really essential or not. This PR keeps the existing behaviour of the dialogs. 

Every scenario is tested and all functions seem to be mostly correct, but feedback or suggestions for an alternate implementation/approach etc are welcomed. Obviously, something much simpler would be preferable.


mathiascode:
Closing in favor of https://github.com/nicotine-plus/nicotine-plus/pull/2594

PR #2322, opened 2023-01-15
mainwindow.py: avoid continuous "size-allocate" signal calls on GTK 3

slook:
+ Fixed: An issue where the "size-allocate" signal continuously calls many times per second on Xfce causing some excess resource usage in Xorg.

Moving/reszing does trigger the `"notify::is-active"` signal on GTK 3 so we can use that instead in order to save the window size and position only when needed.

Not tested on other window managers, working on Xfce 4.16, GTK 3.24.24... it might be an idea to also make a call `on_shutdown()` using `self.window` (but there might be a risk that the window object is not realized).

### DIFF
--- a/pynicotine/gtkgui/mainwindow.py
+++ b/pynicotine/gtkgui/mainwindow.py
@@ -351,7 +351,6 @@ class MainWindow(Window):
         else:
             self.window.connect("delete-event", self.on_close_request)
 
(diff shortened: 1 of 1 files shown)


mathiascode:
Thanks for the idea, I've implemented it in https://github.com/nicotine-plus/nicotine-plus/commit/3ef44c88afaf1e42e847f283a0379936d83a0f9d.

PR #2324, opened 2023-01-17
settings/userinterface.ui: combine tab prefs in a single section

mathiascode:
Whenever I browse the "User Interface" settings page, I'm always confused by the tab-related preferences being split between the general section and a "Secondary Tabs" section. It doesn't feel intuitive to jump between them when you want to change tab preferences.

Combine the preferences in a single section, as such:

### DIFF
--- a/pynicotine/gtkgui/ui/settings/userinterface.ui
+++ b/pynicotine/gtkgui/ui/settings/userinterface.ui
@@ -63,13 +63,6 @@
                   </object>
                 </child>
                 <child>
-                  <object class="GtkCheckButton" id="TabSelectPrevious">
-                    <property name="label" translatable="yes">Restore the previously active main tab at startup</property>
-                    <property name="tooltip-text" translatable="yes">By default the leftmost tab is activated at startup</property>
-                    <property name="visible">True</property>
-                  </object>
-                </child>
-                <child>
                   <object class="GtkBox">
                     <property name="visible">True</property>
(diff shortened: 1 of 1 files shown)


slook:
It could be condensed further ff the 5 "x position" labels were instead checkboxes, such that the Visible Main Tabs in presented as a list with ComboBox for those which customizable Secondary tab bar positions.

pynicotine/gtkgui/ui/settings/userinterface.ui
```diff
@@ -490,134 +306,458 @@
...
-                    <property name="mnemonic-widget">UserInfoPosition</property>
-                    <property name="wrap">True</property>
-                    <property name="wrap-mode">word-char</property>
+                    <property name="homogeneous">True</property>
+                    <property name="spacing">12</property>
+                    <child>
+                      <object class="GtkLabel">
+                        <property name="visible">True</property>
+                        <property name="xalign">0</property>
+                        <property name="label" translatable="yes">User info tab bar position:</property>
```
slook, line 437: `info` -> `Profiles`

mathiascode:
How does this look?

slook:
Nice, to complete the idea suggest a new "Tab bar colors:" sub-section to contain the 3 items "Regular", "Changed" and "Highlighted" at the end of the Tabs section (below the positions).

mathiascode:
I think it's better to keep all color choosers the same.

PR #2327, opened 2023-01-18
pluginsystem.py: handle command name conflicts

slook:
+ Added: Log a warning if a plugin command already exists in another module for each interface
+ Fixed: Don't remove the wrong command if there was a conflicting command in another plugin
+ Changed: Variable name `command_type` -> `command_interface`

This PR only logs conflicting commands that are skipped, but there could be an opportunity here (while in place of if) to add a counter suffix - i.e: `/command` -> `/command2` - which could help to make the handler resilient to conflicting command names, or alternaviely adding a prefix - i.e: `/plugin_name.command` - but that would mean tracking actual added command names rather than defined command names (to allow correct `disable_plugin()` behaviour), and the resulting prefixed/suffixed names for the same command could end up being different if there are any `disabled_interfaces`. This PR only logs conflicting commands that are skipped.

Conflicting aliases have not been tested, assume the help output will simply be wrong.


pynicotine/plugins/core_commands/__init__.py
```diff
@@ -50,7 +50,7 @@ def __init__(self, *args, **kwargs):
                 "usage_chatroom": ["<user>"],
                 "usage_private_chat": ["[user]"]
             },
-            "sample": {
+            "port": {
```
slook, line 53: Debug

PR #2336, opened 2023-01-22
pluginsystem.py: give core_commands a human_name

slook:
+ Added: Human name of "core_commands" plugin to application_name + "Commands", i.e: "Nicotine+ Commands"

The string appears for the main group in the `/help` output as well as for the title of `self.log` messages.

Without a `human_name` value, the plugin log function is not really usable, if we were to ever want to use that.


mathiascode:
Perhaps we should just add a PLUGININFO file.

slook:
Done

mathiascode:
PLUGININFO files should support translations, so you could use the underscore syntax for the name.

slook:
done

mathiascode:
Thanks!

PR #2337, opened 2023-01-22
pluginsystem.py: support boolean value to disable command usage

slook:
- Removed: `"disable"` item from command `data`
- Added: Boolean can be specified as value for `"usage_chatrooms"`, `"usage_private_chat"` and `"usage_cli"` items,
defaults to `True` (if nothing is specified then the command is available).

It would be cleaner to iterate through the command interfaces to avoid code duplication in `enable_plugin()`, but this illustrates the concept which minimizes options in plugin commands and may be more intuitive for plugin developers to implement.

### DIFF
--- a/pynicotine/plugins/core_commands/__init__.py
+++ b/pynicotine/plugins/core_commands/__init__.py
@@ -44,43 +44,43 @@ class Plugin(BasePlugin):
             "me": {
                 "callback": self.me_command,
                 "description": _("Say something in the third-person"),
-                "disable": ["cli"],
                 "group": _("Chat"),
-                "usage": ["<something..>"]
+                "usage": ["<something..>"],
+                "usage_cli": False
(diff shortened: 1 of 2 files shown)


mathiascode:
I don't find this more intuitive myself. It doesn't seem like a good idea to allow both boolean and list values (from a type checking perspective), and it's not obvious that you can disable a command by specifying False.

slook:
Thank you for explaining your reasoning. I don't know if I agree that the current system is intuitive, but I do agree that it is a pain from a type checking point of view.

mathiascode:
I don't know if the current way is intuitive or not, it can probably be improved. `usage` is a bad/incorrect term for specifying command parameters though, it's too broad. If we used `parameters` or another more fitting term, this PR wouldn't make much sense.

PR #2344, opened 2023-01-25
core_commands: msg command and shlex split

slook:
+ Moved: `/msg` command into core_commands plugin
+ Added: Accept "long strings with spaces" as quoted arguments, but don't require quotes if arguments are unambiguous
+ Changed: Don't switch to the tab on entering `/msg` command, because that's what the /pm command is for

Lexical (shell-like) argument splitting allows support for quoted "user names". Quotes are not mandatory if the argument is the first or last one, nor if the username is a single word, i.e; the following inputs all validate into 2 arguments:

`/msg "user name" "message text"`  <- both arguments quoted, ok.
`/msg user name "message text"`  <- only final argument quoted, ok.
`/msg "user name" message text`  <- only first argument quoted, ok.
`/msg username message text`  <- no quotes required for single-word usernames, ok.

_Review/testing points:_

- 3 or more `num_required_args` are tested working with the new `split_args()` function. The `/msg` only needs to accept two arguments, but other commands might need more. A helpful error message is shown on the interface in the case of a mismatch, such as "Too many arguments". Setting `num_required_args=0` returns unlimited splits.

- It is very likely that the `<user>` argument will contain spaces, because many users have such long names. Some commands are useless without support for arguments that contain spaces.

- The proposed implementation is a useful utility that should be available to plugin developers to process any string in the way they want, with any amount of required arguments (such as perhaps from secondary input() prompt or another source).

- Closer integration for argument parsing might be considered into the pluginsystem core, but caution is needed to ensure that optional arguments don't throw an error that prevents the plugin command from being triggered.

- Strings containing `'`_inline `"`quotes`"` can be `"`protected`"` with alternate quotation marks_`'`, and other features (but unfortunately there is no _maxsplit_ feature). See [shlex module Python docs](https://docs.python.org/3.8/library/shlex.html#shlex.split) for details.

### DIFF
--- a/pynicotine/gtkgui/widgets/textentry.py
+++ b/pynicotine/gtkgui/widgets/textentry.py
@@ -113,19 +113,6 @@ class ChatEntry:
             if arg_self:
                 core.request_ip_address(arg_self)
 
-        elif cmd in ("/m", "/msg"):
-            if args:
-                args_split = args.split(" ", maxsplit=1)
-                user = args_split[0]
-                msg = None
-
-                if len(args_split) == 2:
-                    msg = args_split[1]
-
-                if msg:
-                    core.privatechat.show_user(user)
-                    core.privatechat.send_message(user, msg)
-
         elif cmd in ("/us", "/usearch"):
             args_split = args.split(" ", maxsplit=1)
 
--- a/pynicotine/plugins/core_commands/__init__.py
+++ b/pynicotine/plugins/core_commands/__init__.py
@@ -61,6 +61,14 @@ class Plugin(BasePlugin):
                 "usage_chatroom": ["<user>"],
                 "usage_private_chat": ["[user]"]
             },
+            "msg": {
+                "aliases": ["m"],
+                "callback": self.msg_command,
+                "description": _("Send private message to user"),
+                "disable": ["cli"],
+                "group": _("Private Chat"),
+                "usage": ["<user>", "<message..>"]
+            },
             "pm": {
                 "callback": self.pm_command,
                 "description": _("Open private chat"),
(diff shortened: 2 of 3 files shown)


slook:
```
help sample
Listing 1 available commands matching "sample":

Nicotine+ Commands:
	/sample, /demo <choice1|choice2> <second argument> [something else..]  -  Sample command description

sample choice1 second arg something else
Hello, testing 3 arguments: >choice1<  >second<  >arg something else<

sample choice1 "second arg" something else
Hello, testing 3 arguments: >choice1<  >second arg<  >something else<

sample choice1 second arg
Hello, testing 3 arguments: >choice1<  >second<  >arg<

sample choice1 "second arg"
Hello, testing 3 arguments: >choice1<  >second arg<  >False<

sample choice1 second arg "something else"
Hello, testing 3 arguments: >choice1<  >second arg<  >something else<

sample choice1 "second arg" "something" "else"
Too many arguments, 3 required, 4 given (3 quoted)
Hello, testing 3 arguments: >False<  >False<  >False<

sample choice1 "second arg" 'some "thing" else'
Hello, testing 3 arguments: >choice1<  >second arg<  >some "thing" else<

sample choice1 'the "second" argument' something else
Hello, testing 3 arguments: >choice1<  >the "second" argument<  >something else<

sample choice1 "second mismatch quoted arg' something else
No closing quotation
Missing <choice1|choice2> argument
Usage: /sample <choice1|choice2> <second argument> [something else..]

sample choice1 "second mismatch quoted arg' something else"
Hello, testing 3 arguments: >choice1<  >second mismatch quoted arg' something else<  >False<

sample choice1 "second mismatch quoted arg' something" else
Hello, testing 3 arguments: >choice1<  >second mismatch quoted arg' something<  >else<

```

mathiascode:
I intend to merge this for 3.3.0, just haven't gotten around to reviewing the changes yet.

slook:
The actual `split_args()` method works alright, but I don't like the style of implementation, it needs to be refactored.

mathiascode:
Postponing this in order to release 3.3.0 sooner, hopefully we can get it into a 3.3.x point release.

mathiascode:
What would be ideal is to have `split_args()` take a single argument, `args`, and automatically split them according to the command parameters. Quotation marks could be used to influence how long an argument is, if it contains spaces.

slook:
That would be ideal but it would only work for commands whereby all of the arguments are required and none optional.

PR #2345, opened 2023-01-25
core_commands: ip command

slook:
+ Moved: `/ip` command into core_commands plugin
+ Changed: `/ip` command get address from core and print to output if available, fallback to request only if user is unknown
+ Added: `/ip` command can accept IPv4 address as argument for looking up an online address to output associated username

There is no text around the raw output strings, which makes copy & paste nice and easy when working with the network filters in a terminal, but it could be debated as to if more verbosity could be desirable.

- The future plan is to also output the Online/Offline status and Banned/Ignored states, as well as listing any stored IP addresses/usernames associated with a particular user/address.

-  For now this PR brings all the functionality that is required to replace the existing `/ip` command.

### DIFF
--- a/pynicotine/plugins/core_commands/__init__.py
+++ b/pynicotine/plugins/core_commands/__init__.py
@@ -94,6 +94,13 @@ class Plugin(BasePlugin):
                 "usage": ["<room>"],
                 "usage_chatroom": ["[room]"]
             },
+            "ip": {
+                "callback": self.ip_address_command,
+                "description": _("Show IP address or username"),
+                "group": _("Network Filters"),
+                "usage": ["<user or ip>"],
+                "usage_private_chat": ["[user]", "[ip]"]
+            },
             "rescan": {
                 "callback": self.rescan_command,
                 "description": _("Rescan shares"),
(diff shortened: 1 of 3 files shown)


pynicotine/plugins/core_commands/__init__.py
```diff
@@ -245,6 +252,25 @@ def leave_command(self, args, room=None, **_unused):
...
+    def ip_address_command(self, args, user=None, **_unused):
+
+        if self.core.network_filter.is_ip_address(args):
+            self.output(self.core.network_filter.get_known_username(args))
+            return
+
+        if args:
+            user = args
+
+        known_ip_address = self.core.network_filter.get_known_ip_address(user) or False
```
mathiascode, line 266: `or False` is not necessary here. None, empty lists, empty strings etc. are already falsy.

slook: I'm tempted to use your new `get_user_ip_addresses()` function instead, it would almost work but there are extraneous parameters in the way as there's no need to modify the filter. I'd want to get all previous_ip_addressess from both filter lists.

PR #2347, opened 2023-01-25
networkfilter.py: remember previous IP addresses of user

mathiascode:
- If an IP-banned user changes their IP address, keep the filter for the old IP address in place until the user is unbanned.
- Various code cleanups

### DIFF
--- a/pynicotine/core.py
+++ b/pynicotine/core.py
@@ -333,6 +333,7 @@ class Core:
         self.user_status = slskmessages.UserStatus.OFFLINE
 
         # Clean up connections
+        self.user_addresses.clear()
         self.user_statuses.clear()
         self.watched_users.clear()
 
--- a/pynicotine/gtkgui/dialogs/preferences.py
+++ b/pynicotine/gtkgui/dialogs/preferences.py
@@ -888,7 +888,7 @@ class IgnoredUsersPage:
             return
 
         if ip_address not in self.ignored_ips:
-            user = core.network_filter.get_known_username(ip_address) or ""
+            user = core.network_filter.get_online_username(ip_address) or ""
             self.ignored_ips[ip_address] = user
             self.ignored_ips_list_view.add_row([ip_address, user])
 
@@ -1031,7 +1031,7 @@ class BannedUsersPage:
             return
 
         if ip_address not in self.banned_ips:
-            user = core.network_filter.get_known_username(ip_address) or ""
+            user = core.network_filter.get_online_username(ip_address) or ""
             self.banned_ips[ip_address] = user
             self.banned_ips_list_view.add_row([ip_address, user])
             self.ip_ban_required = True
--- a/pynicotine/gtkgui/widgets/popupmenu.py
+++ b/pynicotine/gtkgui/widgets/popupmenu.py
@@ -397,9 +397,9 @@ class UserPopupMenu(PopupMenu):
         self.actions[_("Ignore User")].set_state(
             GLib.Variant("b", core.network_filter.is_user_ignored(self.user)))
         self.actions[_("Ban IP Address")].set_state(
-            GLib.Variant("b", core.network_filter.get_cached_banned_user_ip(self.user) or False))
+            GLib.Variant("b", core.network_filter.is_user_ip_banned(self.user)))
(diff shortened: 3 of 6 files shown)


mathiascode:
I'm using just the username as the placeholder instead of `?.?.?.?`, seemed a bit cleaner to handle. It doesn't look ideal in the IP address column, but perhaps usernames could be replaced with "Unknown" before appending to the list view somehow.

slook:
> perhaps usernames could be replaced with "Unknown" before appending to the list view somehow

The idiom `user in ip_addresses` is misleading to read in the code, as well as looking a bit dumb in the list. Suggest adding something like `f"{user} (Unknown)"` instead of just `user`.

> Let me know if I broke something that's needed for CLI.

Nothing seems to be broken, only very minor changes are required to my draft branch.

mathiascode:
> The idiom `if user and user in ip_addresses` is misleading to read in the code, as well as looking a bit dumb in the list. Suggest adding something like `f"{user} (Unknown)"` instead of just `user`, would be a good idea, yes.

How about `username in ip_addresses` and a comment next to it?

I was mainly thinking about replacing the whole username with `Unknown` in the preferences GUI. If we just append `(Unknown)`, we will have problems with translations anyway.

pynicotine/core.py
```diff
@@ -333,6 +333,7 @@ def _server_disconnect(self, manual_disconnect=False):
         self.user_status = slskmessages.UserStatus.OFFLINE
 
         # Clean up connections
+        self.user_addresses.clear()
```
slook, line 336: Crash on quit, after interacting with a private chat of a non-existant user who is never online:

```
Error in sys.excepthook:
  File "/home/user/Git/slook/nicotine-plus/pynicotine/gtkgui/mainwindow.py", line 362, in on_window_active_changed
    self.privatechat.clear_notifications()
  File "/home/user/Git/slook/nicotine-plus/pynicotine/gtkgui/privatechat.py", line 133, in clear_notifications
    self.window.application.notifications.clear("private", user=user)
  File "/home/user/Git/slook/nicotine-plus/pynicotine/gtkgui/widgets/notifications.py", line 71, in clear
    if core.notifications.remove_hilite_item(location, item):
  File "/home/user/Git/slook/nicotine-plus/pynicotine/notifications.py", line 57, in remove_hilite_item
    if item not in self.chat_hilites[location]:
KeyError: 'private'

Original exception was:
  File "/home/user/Git/slook/nicotine-plus/pynicotine/gtkgui/mainwindow.py", line 362, in on_window_active_changed
    self.privatechat.clear_notifications()
  File "/home/user/Git/slook/nicotine-plus/pynicotine/gtkgui/privatechat.py", line 133, in clear_notifications
    self.window.application.notifications.clear("private", user=user)
  File "/home/user/Git/slook/nicotine-plus/pynicotine/gtkgui/widgets/notifications.py", line 71, in clear
    if core.notifications.remove_hilite_item(location, item):
  File "/home/user/Git/slook/nicotine-plus/pynicotine/notifications.py", line 57, in remove_hilite_item
    if item not in self.chat_hilites[location]:
KeyError: 'private'
/usr/lib/python3/dist-packages/gi/overrides/Gio.py:43: Warning: Source ID 9479 was not found when attempting to remove it
  return Gio.Application.run(self, *args, **kwargs)
```

mathiascode:
Or maybe something like this, if we don't want to mess with the listview append code and also want to save space:

Screenshot from 2023-01-26 02-40-25

pynicotine/core.py
mathiascode: That's a different issue.

slook:
Yes that fits better

pynicotine/core.py
slook: Moved to discussion https://github.com/nicotine-plus/nicotine-plus/discussions/2145#discussioncomment-4783141

mathiascode:
> What happens when you hit Remove on a user who has multiple address entries, does it just remove the one item or all of the matching ones?

One item

mathiascode:
Yep. In the future, we could sort the listviews by username to make it easier to find IP addresses for the same username.

PR #2353, opened 2023-01-29
core_commands: say command

slook:
+ Added: `/say` command into core_commands plugin, as it was traditionally in previous versions of Nicotine
+ Added: Don't attempt to send chat room message unless joined, fail with error message if not

_Note: Only single-word room names are currently supported._ 

If we don't want to include this command in the project right now, then it can be excluded with the `"disabled"` property, (or perhaps be placed into a hidden help group, or an optional plugin?). rather than it being omitted entirely.

### DIFF
--- a/pynicotine/plugins/core_commands/__init__.py
+++ b/pynicotine/plugins/core_commands/__init__.py
@@ -102,6 +102,13 @@ class Plugin(BasePlugin):
                 "usage": ["<room>"],
                 "usage_chatroom": ["[room]"]
             },
+            "say": {
+                "callback": self.say_command,
+                "description": _("Say message in specified chat room"),
+                "disable": ["cli"],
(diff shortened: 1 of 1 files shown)


mathiascode:
> as it was traditionally in previous versions of Nicotine

When was this a thing? I don't recall.

slook:
I probably remember it from Museek/Mucous:

https://github.com/eLvErDe/museek-plus/blob/36c3e9bfb491ee5f6a4a31e73ba38c888b482407/mucous/pymucous/MucousHelp.py#L49

It was implemented as a way of providing some chat functionality without implementing a full-blown chat interface. Suppose you were only viewing the [Chat] log catagory on the CLI, then it might be more relevant there.

My guess is that it may have been removed to prevent bot spam abuse, but I doubt a bot plugin would use this command to do that anyway.

PR #2354, opened 2023-01-29
pluginsystem.py: sanity check before send_public

slook:
+ Added: Avoid sending a chat message to a room that isn't joined
+ Added: Avoid sending a blank message

It is not possible to send a message to a non-joined room, so we shouldn't allow the possibility of sending the server erroneous requests.


mathiascode:
Not sure why we aren't using core.chatrooms.send_message() here.

mathiascode:
https://github.com/nicotine-plus/nicotine-plus/commit/94850df43723420375d0b120c36f8a0fdadc82d7

PR #2363, opened 2023-02-01
Avoid math division operation in main loop sleep timer

slook:
+ Changed: Use explicit time values instead of doing numerous math divide calculations, to avoid unnecessary work.
+ Changed: Sleep for 16ms instead of 16.666~ms (59 per second instead of 60 per second), to avoid a potential rounding operation being needed further down the stack, time.sleep() might not support nanosecond resolution on all systems.

This PR makes no measurable difference to performance, any gain of perceived efficiency is purely theoretical.

### DIFF
--- a/pynicotine/slskproto.py
+++ b/pynicotine/slskproto.py
@@ -184,7 +184,7 @@ class SoulseekNetworkThread(Thread):
     CONNECTION_BACKLOG_LENGTH = 4096
     SOCKET_READ_BUFFER_SIZE = 1048576
     SOCKET_WRITE_BUFFER_SIZE = 1048576
-    SLEEP_MIN_IDLE = 0.016
+    SLEEP_MIN_IDLE = 0.016  # ~60 times per second
 
     def __init__(self, queue, user_addresses):
         """ queue is deque object that holds network messages from Core. """
--- a/pynicotine/slskproto.py
+++ b/pynicotine/slskproto.py
@@ -2257,7 +2257,7 @@ class SoulseekNetworkThread(Thread):
             self._calc_loops_per_second(current_time)
 
             # Don't exhaust the CPU
-            time.sleep(0.02)
+            time.sleep(0.016)
 
         # Networking thread aborted
         self._manual_server_disconnect = True
--- a/pynicotine/slskproto.py
+++ b/pynicotine/slskproto.py
@@ -2257,7 +2257,7 @@ class SoulseekNetworkThread(Thread):
             self._calc_loops_per_second(current_time)
 
             # Don't exhaust the CPU
-            time.sleep(1 / 60)
+            time.sleep(0.02)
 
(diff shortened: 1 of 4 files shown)


mathiascode:
> Now that network events are consumed in batches, the interval can be increased further to about 50ms (20 per second)

Changed in f829f287f89d822861a0e592403672231e60ea5a

I wouldn't go lower in the network loop, since you'll most likely start impacting transfer speeds.

slook:
Suggest then that the value should be `0.016`?

mathiascode:
This seems like unnecessary work and complexity for no reason, when there aren't even any measurements available. I would leave the current code alone, unless you have proof that improvements need to be made.

slook:
From doing a debug measurement using a variable, doing the division operation takes about half of one millisecond (thus, sleep time is over ~17ms rather than the intended target of 16.66~ms per loop).

It's impossible to decern if this extra time is spent doing the math operation or may be more likely the process of assignment to a variable (that is needed in order to perform the debug time measurement), but either way, this experiment proves that doing any sort of operation in this loop has a high consequence (due to it being performed ~60 times per second x 0.5ms = 30ms wasted per second), so it would not be desirable to implement a dynamic idler here and since that is the case, then even if any benefit is marginal there is no reason not to specify an explicit millisecond value for the idle sleep timer.

mathiascode:
You could add a constant with `1 / 60` as the value here and see if there's any difference: https://github.com/nicotine-plus/nicotine-plus/blob/f829f287f89d822861a0e592403672231e60ea5a/pynicotine/slskproto.py#L186-L187

slook:
> see if there's any difference

It seems more possible for the process to reach all the way down to 0% in py-spy.

mathiascode:
Tested with this code:

```
t1 = time.time()
time.sleep(1 / 60)
print(time.time() - t1)
```

The value is usually 0.0167... on my system, so there seems to be a slight inaccuracy.

With 1 / 60, self._loops_per_second is usually 58-59.
With 0.0162, self._loops_per_second is 60 on my system.

slook:
Also on my system, the difference is not really measurable. This optimization is intended with older or low-spec systems in mind.

Having the constant doesn't seem to carry any overhead, and allows for the future possibility of implementing a deeper sleep feature.

pynicotine/slskproto.py
```diff
@@ -184,6 +184,7 @@ class SoulseekNetworkThread(Thread):
     CONNECTION_BACKLOG_LENGTH = 4096
     SOCKET_READ_BUFFER_SIZE = 1048576
     SOCKET_WRITE_BUFFER_SIZE = 1048576
+    SLEEP_MIN_IDLE = 0.016
```
mathiascode, line 187:
```suggestion
    SLEEP_MIN_IDLE = 0.016  # ~60 times per second
```

PR #2371, opened 2023-02-06
core_commands: plugin command

slook:
- Removed: `/toggle` command (was undocumented)
+ Added: `/plugin` command with `toggle` and `info` sub-commands
+ Changed: toggle_plugin() function does update config with enabled plugins

### DIFF
--- a/pynicotine/pluginsystem.py
+++ b/pynicotine/pluginsystem.py
@@ -440,9 +440,13 @@ class PluginHandler:
         enabled = plugin_name in self.enabled_plugins
 
         if enabled:
-            self.disable_plugin(plugin_name)
+            if self.disable_plugin(plugin_name):
+                config.sections["plugins"]["enabled"].remove(plugin_name)
         else:
-            self.enable_plugin(plugin_name)
+            if self.enable_plugin(plugin_name):
+                config.sections["plugins"]["enabled"].append(plugin_name)
+
+        config.write_configuration()
 
     def load_plugin(self, plugin_name):
 
--- a/pynicotine/gtkgui/widgets/textentry.py
+++ b/pynicotine/gtkgui/widgets/textentry.py
@@ -113,10 +113,6 @@ class ChatEntry:
             core.now_playing.display_now_playing(
                 callback=lambda np_message: self.send_message(self.entity, np_message))
 
-        elif cmd == "/toggle":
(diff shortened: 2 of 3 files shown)


pynicotine/pluginsystem.py
```diff
@@ -440,9 +440,13 @@ def toggle_plugin(self, plugin_name):
         enabled = plugin_name in self.enabled_plugins
 
         if enabled:
-            self.disable_plugin(plugin_name)
+            if self.disable_plugin(plugin_name):
+                config.sections["plugins"]["enabled"].remove(plugin_name)
```
mathiascode, line 444: I feel like this should go into the enable/disable_plugin methods instead, perhaps with a `permanent` boolean parameter that's false when loading plugins on startup and disabling them on shutdown.

slook: > when loading plugins on startup and ~disabling~ [unloading] them on shutdown

It seems that this should be the intention of having the seperate load/unload methods, but instead these methods have become dependent upon eachother such that a refactor would be required.

> `permanent/persistent`

`remember`?

mathiascode: I don't remember if we used to have load/unload methods, it's been too long.

> it would be better to avoid having to do this if possible.

We need something similar for the preferences dialog as well. It's currently possible to enable/disable plugins, and end up in an inconsistent state by pressing Cancel instead of OK. We need to write the list of enabled plugins immediately.

slook: Yeh I was looking at that too. The separate tracking that goes on in Preferences might as well be removed, as Plugins is one category that immediate config changes is unavoidable (yet that behaviour is desirable in most cases imo).

mathiascode: Addressed in 5408c40ffa67364ef480c538929bbdff1bfc19d5

mathiascode: Is there any need for the "Enable plugins" toggle? Not sure in what situation you would want to disable all plugins.

slook: Yes, there might be certain situations where the configured plugins are not wanted perhaps just for one session.

The advantage of having the global plugin toggle is that the user's selected combination of plugins are remembered.

PR #2374, opened 2023-02-06
pylint: disable no-member in Plugin modules

slook:
Disable the `no-member` check for plugin modules where namespaces are manipulated during runtime and thus existing member attributes cannot be deduced by static analysis.

Allows the use of `self.parent` reference in core Plugin modules, as documented:
https://github.com/nicotine-plus/nicotine-plus/blob/23180d9623e9c98f5cb2ad2d4adf877d07e6d7f0/pynicotine/pluginsystem.py#L56

May suggest this might also be appropriate for the `pynicotine.widgets.dialogs` module.

- Alternatively, it looks like we could specify the PluginHandler and Plugin Class in the setting above.

Afaik these only affect membership-related tests.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -399,7 +399,8 @@ ignored-classes=optparse.Values,thread._local,_thread._local
 # (useful for modules/projects where namespaces are manipulated during runtime
 # and thus existing member attributes cannot be deduced by static analysis). It
 # supports qualified module names, as well as Unix pattern matching.
-ignored-modules=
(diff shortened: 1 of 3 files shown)


mathiascode:
Fixed in https://github.com/nicotine-plus/nicotine-plus/commit/23aee2e2724854d03567436d4d95da34ffb61d88

mathiascode:
> May suggest this might also be appropriate for the pynicotine.widgets.dialogs module.

This is a different issue (GTK 3 vs GTK 4).

PR #2382, opened 2023-02-10
Implement file type column with icons

mathiascode:
I was curious how much effort it would take to implement this. Turns out it's quite straightforward, and doesn't seem to impact performance.

### DIFF
--- a/pynicotine/gtkgui/search.py
+++ b/pynicotine/gtkgui/search.py
@@ -367,20 +367,20 @@ class Search:
 
         self.column_offsets = {}
         self.column_numbers = list(range(self.resultsmodel.get_n_columns()))
-        color_col = 18
+        has_free_slots_col = 18
         self.cols = cols = initialise_columns(
             self.window, "file_search", self.tree_view,
-            ["id", _("ID"), 50, "number", color_col],
-            ["user", _("User"), 200, "text", color_col],
+            ["id", _("ID"), 50, "number", has_free_slots_col],
+            ["user", _("User"), 200, "text", has_free_slots_col],
             ["country", _("Country"), 25, "icon", None],
-            ["speed", _("Speed"), 120, "number", color_col],
-            ["in_queue", _("In Queue"), 110, "number", color_col],
-            ["folder", _("Folder"), 400, "text", color_col],
-            ["file_type", _("File Type"), 40, "icon", None],
-            ["filename", _("Filename"), 400, "text", color_col],
-            ["size", _("Size"), 100, "number", color_col],
-            ["bitrate", _("Bitrate"), 100, "number", color_col],
-            ["length", _("Duration"), 100, "number", color_col]
+            ["speed", _("Speed"), 120, "number", has_free_slots_col],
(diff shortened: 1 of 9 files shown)


slook:
This looks lovely, what a difference!

Minimum width is slightly too narrow:
image

pynicotine/gtkgui/search.py
```diff
@@ -482,18 +487,19 @@ def create_model(self):
...
+            str,                  # (9)  h_bitrate
+            str,                  # (10) h_length
+            GObject.TYPE_UINT,    # (11) bitrate
+            str,                  # (12) fullpath
+            str,                  # (13) country
+            GObject.TYPE_UINT64,  # (14) size
+            GObject.TYPE_UINT,    # (15) speed
+            GObject.TYPE_UINT,    # (16) queue
+            GObject.TYPE_UINT,    # (17) length
+            bool                  # (18) free slots
```
slook, line 502: These numeric column id's seem like they are a real pain to change, and makes a mess of the diff everytime we add something or want to make changes to the GUI.

Perhaps instead they can be mapped using a class or something, such as they could be altered in one place.

pynicotine/gtkgui/transferlist.py
```diff
@@ -134,14 +137,17 @@ def __init__(self, window, transfer_type):
...
         cols["path"].set_sort_column_id(1)
-        cols["filename"].set_sort_column_id(2)
-        cols["status"].set_sort_column_id(3)
-        cols["queue_position"].set_sort_column_id(13)
-        cols["percent"].set_sort_column_id(5)
-        cols["size"].set_sort_column_id(10)
-        cols["speed"].set_sort_column_id(12)
-        cols["time_elapsed"].set_sort_column_id(14)
-        cols["time_left"].set_sort_column_id(15)
+        cols["file_type"].set_sort_column_id(2)
```
slook, line 140: It's debatable if sorting by the label value of FILE_TYPE_ICON_LABELS rather than the actual file extension is desirable or not. The sort order seems to be a bit random.

Perhaps it would be useful if the tooltip did show the extention, such as "mp3 (Audio)".

pynicotine/gtkgui/search.py
mathiascode: This would be a nice change in the TreeView class, so we could always reference columns by the string column_ids instead of the numerical ones.

pynicotine/gtkgui/transferlist.py
mathiascode: I personally like having the files in generic file type groups when sorting, it's easier to separate music and videos from images and other documents/text files.

mathiascode:
Labeling all unknown files as "Document" is a bit lazy. Perhaps "Text" and "Other" would be fine?

slook:
"Document" is better than "Other" imo.

Is there a Text icon? Don't forget the `nfo` text file type.

mathiascode:
> Is there a Text Document icon with lines of writing on it? Don't forget the `nfo` text file type.

There is. :)

I've added common document/text file extensions that provide useful information for most people. Let me know if I forgot to add something.

PR #2383, opened 2023-02-11
search.py: implement generic file type filters

mathiascode:
Allows you to use filters such as "audio|image", "audio|text|!mp3", "!video|!image".

### DIFF
--- a/pynicotine/gtkgui/search.py
+++ b/pynicotine/gtkgui/search.py
@@ -56,6 +56,7 @@ from pynicotine.gtkgui.widgets.treeview import show_file_path_tooltip
 from pynicotine.gtkgui.widgets.treeview import show_file_type_tooltip
 from pynicotine.gtkgui.widgets.ui import UserInterface
 from pynicotine.logfacility import log
+from pynicotine.shares import FileTypes
 from pynicotine.slskmessages import SEARCH_TOKENS_ALLOWED
 from pynicotine.slskmessages import FileListMessage
 from pynicotine.utils import factorize
@@ -290,10 +291,18 @@ class Searches(IconNotebook):
 
 class Search:
 
+    FILTER_GENERIC_FILE_TYPES = (
+        ("audio", FileTypes.AUDIO),
+        ("executable", FileTypes.EXECUTABLE),
+        ("image", FileTypes.IMAGE),
+        ("video", FileTypes.VIDEO),
+        ("text", FileTypes.DOCUMENT_TEXT),
+        ("archive", FileTypes.ARCHIVE)
+    )
     FILTER_PRESETS = {
         "filterbr": ("0", "128", "160", "192", "256", "320"),
         "filtersize": (">10MiB", "<10MiB", "<5MiB", "<1MiB", ">0"),
-        "filtertype": ("flac|wav|ape|aiff|wv|cue", "mp3|m4a|aac|ogg|opus|wma", "!mp3")
+        "filtertype": ("audio", "image", "video", "text", "archive", "!executable", "audio|image|text")
     }
 
     def __init__(self, searches, text, token, mode, mode_label, show_page):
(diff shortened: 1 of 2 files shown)


pynicotine/gtkgui/widgets/theme.py
```diff
@@ -314,7 +314,6 @@ def set_global_style():
 
 CUSTOM_ICON_THEME_NAME = ".nicotine-icon-theme"
 FILE_TYPE_ICON_LABELS = {
-    "application-x-executable-symbolic": _("Executable"),
```
slook, line 317: What has this category been removed?

mathiascode: There are too few file extensions, and I don't want to give special treatment to executables. You can just use "exe" as the filter if you want these files.

slook: A valid use for having both distinctive icon and an obvious method of filtering out executables does warrant special treatment, most users may be aware of the risk of such files but not all of them.

mathiascode: Then the question becomes, what counts as an executable? Some of them don't have a file extension, and we have a bunch of package formats included in the archive-related extension list at the moment. If .exe is the main concern, it would be shorter to type than "executable" anyway.

I'm somewhat concerned that people would still (mis)use the filter to actually find the files, even if the intention is to exclude.

A distinctive icon is fine, but which extensions should have this treatment?

pynicotine/gtkgui/search.py
```diff
@@ -551,7 +552,7 @@ def update_filter_comboboxes(self):
                 presets = (">10MiB", "<10MiB", "<5MiB", "<1MiB", ">0")
 
             elif filter_id == "filtertype":
-                presets = ("flac|wav|ape|aiff|wv|cue", "mp3|m4a|aac|ogg|opus|wma", "!mp3")
+                presets = ("audio", "image", "video", "text", "archive", "audio|image|text", "video|text")
```
slook, line 555: A demonstration of the not `!` operator is needed, suggest `!executable`.

pynicotine/gtkgui/widgets/theme.py
mathiascode: Should we perhaps rename it to "Software" and include all package formats there?

slook: See "Potentially Dangerous File Extensions": https://www.lifewire.com/list-of-executable-file-extensions-2626061

Things like `bat` `scr`, perhaps even scripts like `py` etc.

pynicotine/gtkgui/search.py
mathiascode: I think I would be fine with this after all, when using the revised extension list in my other comment.

pynicotine/gtkgui/widgets/theme.py
slook: > Maybe Software is too generic

"Program", "Package", "Installer"

mathiascode: On second thought, I think "Executable" drives the point home faster (exe, executable, can be run, possibly dangerous).

mathiascode: > Isn't there a library we can use to match extensions with mime types known by the OS?

Perhaps https://docs.python.org/3/library/mimetypes.html

mathiascode: Audio, image and video work fine with the mimetypes module, since these mimetypes start with `audio/`, `image/` and `video/`.

Everything else falls apart, since it either starts with `application/`, or includes a lot of junk we don't want to show (`text/`, which includes e.g. json, xml, config files, source files, which is most certainly junk on the Soulseek network).

mathiascode: Seems like the list is also manually updated on other OSes than Windows, so some extensions will be missing in older Python versions: https://github.com/python/cpython/commits/3.11/Lib/mimetypes.py

I think I'll stick with our curated lists for now.

PR #2385, opened 2023-02-11
nowplaying.py: remove "Playing now" string in $n

slook:
- Removed: Intro text from the `$n` string output.

User reports that otherwise it prevents Now Playing Search from working properly. If the intro text is wanted for use in chat it will have to be added somewhere else.

Fixes https://github.com/nicotine-plus/nicotine-plus/discussions/2380 discussion

### DIFF
--- a/pynicotine/nowplaying.py
+++ b/pynicotine/nowplaying.py
@@ -153,7 +153,7 @@ class NowPlaying:
             self.title["artist"] = artist = lastplayed["artist"]["#text"]
             self.title["title"] = title = lastplayed["name"]
             self.title["album"] = album = lastplayed["album"]["#text"]
-            self.title["nowplaying"] = f"{_('Last played')}: {artist} - {album} - {title}"
+            self.title["nowplaying"] = f"{artist} - {album} - {title}"
 
         except Exception:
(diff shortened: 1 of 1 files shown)


mathiascode:
Seems like this was completely wrong. Outgoing strings should not be translated, and any additional text should be added by customizing the message in preferences.

pynicotine/plugins/now_playing_sender/__init__.py
```diff
@@ -112,7 +112,7 @@ def send_now_playing(self):
             playing = self.core.now_playing.get_np()
 
             if playing:
-                self.send_public(room, playing)
+                self.send_public(room, f"{_('Playing now')}: {playing}")
```
mathiascode, line 115: Revert this

slook:
> Outgoing strings should not be translated, and any additional text should be added by customizing the message in preferences.

I don't know if the intention was to provide receiving users with a method of spam-filtering out unwanted messages from now playing scripts, which is not possible to do if the generated strings are always entirely unique.

Perhaps alternatively, using `send_automatic_message()` might be considered in order to provide a common text anchor with which a spam filter might be applied.

Although, ignoring the sending user's would seem to be more appropriate in such cases.

mathiascode:
I doubt it was the intention, considering two different strings were used.

PR #2386, opened 2023-02-12
now_playing_sender: now command

slook:
+ Moved: `/now` command into now_playing_sender plugin
+ Added: Optional `local` or `broadcast` argument, if there is no argument then send to current chat by default
+ Added: Ability to print the `/now` command output in headless CLI echo (chat functions are unavailable, local echo only)
+ Added: Support for maximum plugin command arguments to be limited by terminating `usage` with empty string `""` (the `local` and `broadcast` choice argument is not relevant in the `cli` interface)
+ Fixed: Support for a single plugin command choice argument to be optional
+ Changed: PLUGININFO Name to remove reference to MPRIS as it is no longer mandatory for "Now Playing Sender" because the /now command provides functionality even for non-MPRIS users
+ Fixed: Only get the now playing string once, rather than recursing in the loop for each listed chat room
- Removed: Special exclusion for Windows and macOS, the plugin might partially operate by using the `/now` command, since platform specific functions are bypassed within the plugin itself.
- Removed: Intro text from `$n` output, any additional text should be added by customizing the message in preferences.

### DIFF
--- a/pynicotine/plugins/now_playing_sender/__init__.py
+++ b/pynicotine/plugins/now_playing_sender/__init__.py
@@ -16,8 +16,12 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from gi.repository import Gio
+import sys
+
 from pynicotine.pluginsystem import BasePlugin
+
+if sys.platform not in ("win32", "darwin"):
+    from gi.repository import Gio
 
 
 class Plugin(BasePlugin):
@@ -35,12 +39,28 @@ class Plugin(BasePlugin):
                 "type": "list string"
             }
         }
-
-        self.last_song_url = ""
+        self.commands = {
+            "now": {
+                "callback": self.now_playing_command,
+                "description": _("Announce the song playing now"),
+                "group": _("Now Playing"),
+                "usage": ["[local|broadcast]"],
+                "usage_cli": [""]
+            }
+        }
+        self.bus = None
+        self.signal_id = None
         self.stop = False
+        self.dbus_mpris_service = ""
+        self.last_song_url = ""
+
+    def loaded_notification(self):
+
+        if sys.platform in ("win32", "darwin"):
+            # MPRIS is not available on Windows and macOS
+            return
 
         self.bus = Gio.bus_get_sync(bus_type=Gio.BusType.SESSION)
-        self.signal_id = None
         self.dbus_mpris_service = "org.mpris.MediaPlayer2."
 
         self.add_mpris_signal_receiver()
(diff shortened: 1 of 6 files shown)


mathiascode:
> * Added: Support for maximum plugin command arguments to be limited by terminating `usage` with empty string `""` (the `send` and `broadcast` arguments must be rejected if in the `cli` interface)

I don't think this is the way to go. It's very confusing at first glance. Do we really need this functionality?

mathiascode:
These changes can't go in until all the now playing-related code is in the same plugin, otherwise you will have a broken experience out of the box with no /now command available.

pynicotine/plugins/now_playing_sender/__init__.py
```diff
@@ -151,4 +163,30 @@ def song_change(self, _connection, _sender_name, _object_path, _interface_name,
...
+        else:
+            # Send in current chat only
+            self.send_message(playing)
+
+    def now_playing_command(self, args, **_unused):
+        """ /now command """
+
+        self.send_now_playing(
+            broadcast=(args == "broadcast"),
+            local=(not args)
```
mathiascode, line 191: The purpose of this command for 20 years has been to broadcast what you're listening to. I'm not willing to make it local by default, otherwise the whole feature might as well be removed.

slook: Good point. The default action should be send (to current chat only).

slook:
>> maximum plugin command arguments to be limited by terminating usage with empty string `""`

> Do we really need this functionality?

Yes, since you removed the "Invalid option" gates in https://github.com/nicotine-plus/nicotine-plus/commit/533b9b2b58c6fc62c521170fc8b742efda96a4e2 then it raises the question of instead providing a built-in method of rejecting unwanted or wrong flag arguments would be preferred over having to handle these in the plugin functions.

It is not proper to allow for example "/quit force extra argument text" to run, because in this case it would be expected that the command will fail because "extra argument text" is not a valid option for the /quit command.

The empty string `""` or `None` syntax is conventionally used for this type of use case to mark end of line, such as for example in the Python `shlex` module (which is used by the `cmd` module), see https://docs.python.org/3/library/shlex.html#shlex.shlex.eof :

- _"This will be set to the empty string (`''`), in non-POSIX mode, and to `None` in POSIX mode."_

If we provide a way of determining what is the expected number of arguments (that might not be the same as the required number of arguments) so that they can be split appropriately (accounting for arguments that might either be a long message string or maybe just a single word), or so that where an argument is valid for a particular interface, but not for another.

Demarcating the End Of Line within the usage parameter does provide a means of doing this. Admittedly its implementation it not intuitive, so I am glad you have raised the issue.

An alternative might be to make a new parameter to specify `maximum_arguments` or `max_args` similar.

PR #2387, opened 2023-02-12
core_commands: ctcp command

slook:
+ Removed: `/ctcpversion` command
+ Added: `/ctcp` command with `query` argument
+ Changed: Allow sending of any CTCP-quoted message, let the receiving client decide what queries it accepts and responds to
+ Changed: Respond with the original query within the paired response as per the [CTCP protocol specification
](https://www.irchelp.org/protocol/ctcpspec.html)

This is backwards compatible, but obviously clients won't respond to anything other than the "VERSION" keyword.

Simpler implementation of #2381


mathiascode:
I'd like to look into CTCP after 3.3.0, but I've decided to move the /ctcpversion command to core_commands as-is for now.

mathiascode:
I think we can merge this as-is.

PR #2396, opened 2023-02-16
Result Filters: toggle recall last result filter item on icon press

slook:
Result filters bar combobox widget accessibility improvements:

+ Added: Click the entry icon to toggle between nothing (`""`) and the most recently used item in the filter history
+ Fixed: Relevant list item highlighted according to the currently active text entry on drop list menu shown
+ Fixed: Avoid unnecessary config write when updating the 1st item in the filter history with exactly the same value

This proposed feature makes it easy to do A/B comparisons on the results, and it is useful to transpose the most recently used filter by recalling it into another search tab without having to manually navigate to the desired item in the drop down list box.


pynicotine/gtkgui/search.py
```diff
@@ -546,9 +550,15 @@ def on_tooltip(widget, pos_x, pos_y, _keyboard_mode, tooltip):
         return show_file_type_tooltip(widget, pos_x, pos_y, tooltip, 6)
 
     def on_combobox_popup_shown(self, combobox, _param):
+
+        # Refilter
         entry = combobox.get_child()
         entry.emit("activate")
 
+        # Highlight current list item
```
mathiascode, line 558: Does this actually do anything? The active item is always at the top, and it was already highlighted.

pynicotine/gtkgui/search.py
```diff
@@ -1471,7 +1481,7 @@ def push_history(filter_id, value):
 
         history = config.sections["searches"].get(filter_id)
 
-        if history is None:
+        if history is None or (history and history[0] == value):
```
mathiascode, line 1484: Split this into a separate guard clause.

pynicotine/gtkgui/search.py
```diff
@@ -1596,6 +1606,26 @@ def on_filter_entry_changed(self, widget):
...
+            return
+
+        history = config.sections["searches"].get(entry.filter_id)
+        recall_text = history[0] if history else ""
+        text = entry.get_text()
+
+        # Recall last filter entry if box empty
+        if not text:
+            entry.grab_focus_without_selecting()
+            entry.set_text(recall_text)
```
mathiascode, line 1621: You could add `entry.set_position(-1)` here

pynicotine/gtkgui/ui/search.ui
```diff
@@ -178,6 +178,7 @@
                             <property name="visible">True</property>
                             <property name="has-entry">True</property>
                             <property name="hexpand">True</property>
+                            <property name="id-column">0</property>
```
mathiascode, line 181: Is this change necessary?

pynicotine/gtkgui/search.py
```diff
@@ -1596,6 +1606,26 @@ def on_filter_entry_changed(self, widget):
         if not widget.get_text():
             self.on_refilter()
 
+    def on_filter_entry_icon_press(self, entry, _icon_pos, press_event, *_args):
+
+        if press_event.triggers_context_menu():
```
mathiascode, line 1611: This is not compatible with GTK 4. https://docs.gtk.org/gtk4/signal.Entry.icon-press.html

pynicotine/gtkgui/search.py
slook: Yes, the presets are at the top, so without this the incorrect item was highlighted upon opening the combo popup list.

pynicotine/gtkgui/ui/search.ui
slook: It is required for `set_active_id()` to highlight current list item.

pynicotine/gtkgui/search.py
slook: The method is present in both GTK 3 and 4:

https://docs.gtk.org/gdk3/method.Event.triggers_context_menu.html
https://docs.gtk.org/gdk4/method.Event.triggers_context_menu.html

mathiascode: Yes, but the event is no longer exposed in GTK 4.

slook: I see. The intention was to allow provision for a secondary action, but it isn't really needed anyway.

Issue #2100, opened 2022-06-27
Banned and ignored users indicators


ghost:
A visual indicator like forbidden sign symbol and muted audio symbol next to the name. Something that makes it easier to see than having to go to preferences and looking at the list.

slook:
Adding icon columns into TreeView is something that going to be considered at the point where the TreeView widget will be overhauled as discussed in https://github.com/nicotine-plus/nicotine-plus/issues/1731 :

_"I think we should aim for a sortable lock icon is_private column. Perhaps in future this could also be used for other attributes, such as _is_buddy_, _is_ignored_ or _is_blocked_. Perhaps this would assist in the implementation of filtering based on those characteristics."_

> Something that makes it easier to see than having to go to preferences and looking at the list.

For now an easier way you can open the Right-click > User(s) menu to see the checked menu-items if a user is banned or ignored.

mathiascode:
This request is blocked until we've ported the chat room user list to the new TreeView object.

mathiascode:
To clarify, where would you like to see these indicators? Only chat rooms, or somewhere else?

JediMaster25:
To be honest I haven't found the need for this since then and I don't remember on what context I thought this would be helpful. I'll write the specifics if I recall it at some point.

slook:
In the chat room users list, it might be logical to apply the greyed-out style to ignored user rows.

mathiascode:
@slook That's what I had in mind as well.

PR #2402, opened 2023-02-18
networkfilter.py: skip redundant ip address checks

slook:
There is a logic error since https://github.com/nicotine-plus/nicotine-plus/commit/d422b7d5cddbd29a20e4bb3457848f20784855a5 in PR #2347 where iterating previous IP's from the list only to then later check if they were in the same list, was waste of time, because we don't need to know the value of ip_address to return a boolean from this method, it is only needed to know if the user is in the list or not.

We already update changed ip addresses in the get_peer_address() callback, so we already have multiple entries with the same username, which means that we don't need to check for previous addresses again during this check, which must be as efficient as possible because it gets called for every relevant incoming network event, and for every joining/leaving chatroom user.

+ Added: Do quick no-membership test to avoid iterating through every item in the IP list, faster in un-filtered cases
+ Added: Do quick membership test to avoid iterating through every item in the IP list, faster in filtered cases
- Removed: Redundant check to get previous IP's from the list to check if they were in the same list, was a waste of time
+ Added: Add actual IP address entry on wildcard pattern match, to appear in the list with associated username
- ~Removed: 2 unused methods that were never referenced anywhere~
+ ~Fixed: Avoid adding `None` as the username value on using /ban or /ignore commands with ip address as argument~

The aim is for the filter check to be fast enough for the state to be visible in live tree views such as proposed in #2100, but reliable caching of usernames is needed for that, such that only one condition has to be passed without any need to iterate.

It might be worth moving the wildcard filter code, to somewhere that is not executed as often. Ideally we want to be at a point where checking the IP filter can be reliably done simply with quick membership testing by username.


mathiascode:
I'll have to look closer at this. I revised that code so many times I probably forgot something...

slook:
> I probably forgot something...

Your implementation of checking all of the `_get_previous_user_ip_addresses` in `_check_user_ips_filtered()` may have been appropriate before, but you did not account for the fact that since `_update_saved_user_ip_addresses()` now stores all of the old addresses for each user, then the information is already known.

mathiascode:
Looks good, sorry for taking so long to merge the PR. After working on the network filter-related code the last time, I didn't feel like looking at it again for a while.

PR #2404, opened 2023-02-20
Result Filters: recall cleared filters undo toggle

### LINKED ISSUE #1400: Improvement ideas for the search result filters
Currently, it's possible to have a "Maximum results per search visible at a time" smaller than "Maximum total results per search". I haven't taken the time to prove it, but I like to keep the former smaller than the latter to help keep Nicotine+'s memory footprint reasonable and its performance from degrading.

Assuming: "Maximum results per search visible at a time": `250`, and "Maximum total results per search" `1000`.

I have some usability improvement ideas for when configured as above:
- Allow pressing the "Results (250)" (specifically the number part) that is shown at the top of each search (above the search result filters). Pressing the button would instantly show any hidden results (i.e. those >250), which could sometimes be more convenient than using the filters to see if any of the possibly 750 hidden results contain what you're looking for.
- When there are hidden results, an asterisk could be placed next to the number, "Results ( 250 * )", giving you a cue to click it. The asterisk could also appear after filtering via the search result filters: "Results ( 123 * ), and clicking it would clear the filters _and_ show all potential 1000 results. Alternatively to an asterisk (which is a bit obscure), a "Show All" button (or similar) could be placed next the the results count and perform the same actions I described.

Other ideas:
- I miss being able to _turn off_ the Result Filters, _not just hide them_, as used to be possible via the main UI. Maybe the "Results Filters" button could be given a Ctrl + click feature to toggle their on/off state, instead of the default plain click which currently just toggles their visibility.
  - Ideally the state would be per set of search results, just as is the state of whether the filters are currently displayed or not. The benefit of being able to toggle the filters on/off is to quickly see any differences _without_ losing the values set in your search filter fields (toggling the active state shouldn't also toggle the fields' visibility - only the plain single-clicking function should do so).
  - The obscurity of this function may not matter due to it being a less used feature I imagine; a tooltip could help reveal its presence: "1 active filter(s) (Ctrl-click to toggle active state)", when deactivated: "Deactivated (Ctrl-click to toggle active state)". The UI should probably reflect this active state by dimming/graying the fields, anything to show that they're disabled.

- A quick way to clear an individual search result filter field without clearing them all could be nice. E.g. the typical "X" to the right of a field. If not done well, these Xs to clear can sometimes be quite annoying to target and click intentionally/successfully.
  - After having thought about it, the icons inside of each search field could be used for this very purpose: on mouse-over, replace the icon with an "X" which then clears the field if pressed (I'd argue that it shouldn't also _set_ the blank field, only _clear_ it).
  - On closer inspection, the icons already double as a method to set the search result filters as a whole (i.e. Enter). This is somewhat convenient since there is no other mouse method to set the filters. If a "Set" button (i.e. to Enter via click) was introduced to the left of the "<x" Clear button, repurposing these icons as I propose wouldn't be a loss of functionality. A "Set" button should likely be present regardless since it's less obscure than having to press a field's icon (assuming wanting to set them via mouse interaction).

Hopefully those make sense and can at minimum serve as some inspiration.

Thanks!

```
Nicotine+ v3.1.0.dev1
```

slook:
Closes #1400

+ Added: Recall Filters after they have been cleared (stored as dict) to enable A/B comparison toggling

Refactored methods are suggested as to help the acquisition of result filter data into the searches module via the core.

Whilst it would perhaps be simpler to just store the compiled `self.filters`, an approach similar to what is offered in this PR is intended to enable more flexible future use of stored filter data, such as stepped undo history and filtered wishlist searches.

For demonstrative purposes the concept is tested in this PR by implementing an undo function for the Clear Active Filters toggle button by means of storing the data in a dict for later recall (just click it again to revert).


mathiascode:
I think this could close https://github.com/nicotine-plus/nicotine-plus/issues/1400. I prefer this method over the "Deactivate search result filters when Result Filters bar is hidden" option suggested in that issue.

Perhaps we could also change the clear icon to an undo icon when appropriate?

slook:
> "Deactivate search result filters when Result Filters bar is hidden" 

I did try this action on the collapse/show bar, but the behaviour felt a bit funky, and there is no way of presenting the option that isn't confusing.

> change the clear icon to an undo icon when appropriate

I did do similar for changing the tooltip text, and I discovered for that to be possible it is necessary to have a FILTERS_CLEAR constant to compare what the empty filters look like (because empty filters represented by a dict is not None).

Therefore, the changing of the button icon can wait until a later time, once we know what direction is best to go with the implementation of the get_filters() and set_filters() methods and events.

PR #2405, opened 2023-02-20
Result Filters: apply search default filters in core with "set-filters" event

slook:
+ Added: New "set-filters" event to be used for setting search result filters, either in a new or currently active search
+ Changed: Get the default search result filters ["defilter"] setting in the core instead of in the gui module

There are no functional changes in this refactor. The intention is to provide a pathway for storing filters as dicts instead of lists.

There is no urgent need to migrate the existing ["defilter"] config setting that is stored as a list, although we could do that. The default filter is moved to be better applied from the core, in order to propose how we might use the nearby mechanism in `do_search()` for wishlist filters and other related tasks without dependancy on the gui search module.

A function is provided to allow the filters to be applied to an active search even if its `token` is not known by the caller. However, the search is referenced by token if `do_search()` calls the event.

The proposal is:

- `get_filter()` reads the currently active _filters_ data, it returns a dict `{filter_id: value,}`
  - `filter_id` are string keys that belong to any number of widgets that are defined in the interface module.
  - `value` are either booleans (buttons) or strings (text entries) that are human readable (joined from active _self.filters_ data).

- `set_filter()` writes values which are provided by the given dict containing `stored_filters`
  - If a `filter_id` is missing then its filter will be cleared. If the entire dict is missing then all the filters will be cleared.
  - If the method is called externally, then filters would be modified in all the open search tabs that have a matching `search_term`, since the page is not referenced directly by token. If no matching pages are open, then nothing happens.
  - `refilter` may be used disable automatic apply, then text entries will only be populated. Default is `True`, but is set to `False` when called from do_search().

It works but I don't know if this is the right way to approach the implementation of the event emitter or not.

It is not decided if the search module should emit an event to signal a changed filter, or if the caller is always responsible for getting the filters.


mathiascode:
The code in master has diverged too much. We'll have to revisit this once https://github.com/nicotine-plus/nicotine-plus/pull/2464 is ready, perhaps as a separate module.

PR #2406, opened 2023-02-21
Now Playing plugin

slook:
Taken from https://github.com/mathiascode/nicotine-plus/tree/nowplaying-plugin merged to latest master, plus:

- [x] Use double quotes everywhere
- [x] Use urlopen
- [x] Get current MPRIS player method
- [x] Fix Last.fm login
- [x] Other command
- [ ] Fix `%` and `$` replacers
- [ ] Migrate old settings from `config` to plugin `settings` on first run


mathiascode:
I don't think I'll be able to review/work on this in a timely manner for 3.3.0, so it will most likely be delayed for a future Nicotine+ release.

PR #2407, opened 2023-02-21
Implemented support fot NAT-PMP

homedirectory:
Nicotine+ currently provides automatic port forwarding only for network devices supporting UPnP. There is another port forwarding protocol -- [NAT-PMP](https://en.wikipedia.org/wiki/NAT_Port_Mapping_Protocol) -- that some of the old AirPort Extreme routers rely on. This pull request brings support for port forwarding through NAT-PMP. 

The library used is a slightly modified version of <https://github.com/yimingliu/py-natpmp>. It is contained in `natpmp.py`. The main class `NatPMP` is designed similary to the `UPnP` class to facilitate a uniform API for port forwarding. However, both of them still remain standalone classes, so it might make sense to define a port forwardng interface for them to implement.

I also updated the Network page in Preferences to include a checkbox for enabling NAT-PMP, very similar to one for UPnP.

### DIFF
--- a/pynicotine/natpmp.py
+++ b/pynicotine/natpmp.py
@@ -1,0 +1,580 @@
+# COPYRIGHT (C) 2020-2022 Nicotine+ Contributors
+#
+# GNU GENERAL PUBLIC LICENSE
+#    Version 3, 29 June 2007
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from threading import Thread
+
+from pynicotine.config import config
+from pynicotine.events import events
+from pynicotine.logfacility import log
+
+RENEWAL_INTERVAL = 14400  # 4 hours
+
+import os
+import re
+import struct
+import socket
+import select
+import platform
+
+# Modified version of NAT-PMP client library from https://github.com/yimingliu/py-natpmp
+# 
+# Provides functions to interact with NAT-PMP gateways implementing version 0
+# of the NAT-PMP draft specification.
+# 
+# This version does not completely implement the draft standard.
+# * It does not provide functionality to listen for address change packets.
+# * It does not have a proper request queuing system, meaning that
+# multiple requests may be issued in parallel, against spec recommendations.
+# 
+# For more information on NAT-PMP, see the NAT-PMP draft specification:
+# http://files.dns-sd.org/draft-cheshire-nat-pmp.txt
+
+# Copyright (c) 2008-2016, Yiming Liu, All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification,
+# are permitted provided that the following conditions are met:
+# 
+# * Redistributions of source code must retain the above copyright notice,
+#   this list of conditions and the following disclaimer.
+# * Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
+#   and/or other materials provided with the distribution.
+# * The names of the author and contributors may not be used to endorse or promote
+#   products derived from this software without specific prior written permission.
+# 
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS 'AS IS'
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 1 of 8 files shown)


pynicotine/natpmp.py
```diff
@@ -0,0 +1,643 @@
...
+# 
+# This version does not completely implement the draft standard.
+# * It does not provide functionality to listen for address change packets.
+# * It does not have a proper request queuing system, meaning that
+# multiple requests may be issued in parallel, against spec recommendations.
+# 
+# For more information on NAT-PMP, see the NAT-PMP draft specification:
+# http://files.dns-sd.org/draft-cheshire-nat-pmp.txt
+
+# Copyright (c) 2008-2016, Yiming Liu, All rights reserved.
```
slook, line 47: ~`2016`~ -> `2023`

Please add to attributions sections in:

+ https://github.com/nicotine-plus/nicotine-plus/blob/master/AUTHORS.md#third-party-attributions

+ https://github.com/nicotine-plus/nicotine-plus/blob/07d5ca8ce878d508484b2942ce37ee37652d2e34/pynicotine/gtkgui/dialogs/about.py#L308

Also add yourself in somewhere if you want.

pynicotine/natpmp.py
```diff
@@ -0,0 +1,643 @@
...
+    NOT_AUTHORIZED = 2  # Not Authorized/Refused/NATPMP turned off
+    NETWORK_FAILURE = 3  # Network Failure
+    OUT_OF_RESOURCES = 4  # can not create more mappings
+    UNSUPPORTED_OP = 5  # not a supported opcode
+    # all remaining results are fatal errors
+    GATEWAY_NO_VALID_GATEWAY = -10
+    GATEWAY_NO_SUPPORT = -11
+    GATEWAY_CANNOT_FIND = -12
+
+    MSG_DICT = {
```
slook, line 99: Please remove all the full-stops we don't use `.` at the end of any short strings.

slook:
> The library used is a slightly modified version of https://github.com/yimingliu/py-natpmp

Out of interest, what are the slight modifications needed for it to work with Nicotine+ specifically?

My review suggestions above don't have to be implemented if that would make maintenance more difficult.

mathiascode:
Thanks for your PR! I appreciate your work, but in the future, I would suggest opening an issue for discussion before submitting a larger PR.

I've refactored upnp.py and renamed it to portmapper.py to allow for more protocol implementations (possibly PCP in the future). Since this PR requires a fair bit of work to comply with our linters and code style, I've implemented the bits of the NAT-PMP protocol we need in the style of our existing UPnP code: https://github.com/mathiascode/nicotine-plus/tree/natpmp

I've reused the "upnp" config option for both UPnP and NAT-PMP, since we fall back to UPnP if NAT-PMP fails.

The only remaining part we have to think about is the best way to receive the gateway IP address, since I'd like to avoid third-party dependencies when possible, and the netifaces library is archived: https://github.com/al45tair/netifaces

slook:
https://github.com/tsukumijima/netifaces-plus 0.12.0

homedirectory:
@slook 
> Out of interest, what are the slight modifications needed for it to work with Nicotine+ specifically?

As of now, there is one required modification. At first, there was a tiny flaw in the library that I fixed to properly implement the spec, but it's already merged into the master branch. Other modifications came from my own refactoring, just for fun. But now that I think about it, using the original version in its current state lends itself better to maintenance. Would it be fine with you if I reverted to the original version to get rid of redundant refactoring?

slook:
Sure, if we can pull what we need directly from upstream where it is maintained separately, then this might be better for you to continue with contributing on that repo as changes may become necessary over time then you can help us to push them here.

However, if many of the functions are superfluous or require varying modification patches then it could be better to roll our own module. Please review Mat's PR to help with the integration in whatever way that makes sense. Your contribution will be worthy of credit either way, thank you for all the work you have put into this.

mathiascode:
> Perhaps it would be useful to provide a configuration setting for it to be manually specified if binding is enabled, either instead of automatic discovery (i.e. on Windows), or to override an automatically discovered gateway address.

I've since added some code in https://github.com/nicotine-plus/nicotine-plus/pull/2408 that works out of the box on Linux, Windows and macOS at least.

mathiascode:
> However, if many of the functions are superfluous or require varying modification patches then it could be better to roll our own module. Perhaps even, our UPnP module might be refactored to achieve a uniform API between the two supported protocols.

I'll most likely roll our own implementation in https://github.com/nicotine-plus/nicotine-plus/pull/2408, since the spec is so simple, but I'll gladly submit potential improvements to py-natpmp so everyone can benefit from them. I used some of their regex for parsing the gateway address, and I could submit the newer ip command method used on Linux.

slook:
> I've since added some code in https://github.com/nicotine-plus/nicotine-plus/pull/2408 that works out of the box on Linux, Windows and macOS at least.

Does it account for non-normal situations where perhaps there could be more than one gateway or suchlike?

It strikes me that the standard output cannot be relied upon in every case, albeit this is not very different from how other projects seem to get this information. However it seems common to offer a manual override setting, so I assume this must be for a good reason.

slook:
Closing in favour of https://github.com/nicotine-plus/nicotine-plus/pull/2408

This is a very worthwhile contribution. Thank you for all your work on this! :)

mathiascode:
> Does it account for non-normal situations where perhaps there could be more than one gateway or suchlike?
>
> It strikes me that the standard output cannot be relied upon in every case, albeit this is not very different from how other projects seem to get this information. However it seems common to offer a manual override setting, so I assume this must be for a good reason.

I don't know, but we can make improvements if necessary.

Nobody is going to understand what a gateway address is, why they need to care about it, and what they should do with such an option, so I'd rather not provide it. We should always do better, instead of forcing the user to work around flaws.

PR #2408, opened 2023-02-21
portmapper.py: add NAT-PMP implementation

mathiascode:


pynicotine/portmapper.py
```diff
@@ -402,17 +483,64 @@ class PortMapper:
...
+
+        log.add_debug("Creating Port Mapping rule...")
+
+        try:
+            try:
+                self._active_implementation = self._natpmp
+                self._natpmp.add_port_mapping(self.LEASE_DURATION)
+
+            except RuntimeError as error:
+                log.add_debug("NAT-PMP not available, falling back to UPnP: %s", error)
```
slook, line 506: Are we not starting with UPnP and falling back to NAT-PMP, whereas this is the other way around here?

pynicotine/portmapper.py
```diff
@@ -41,7 +45,127 @@ def set_port(self, port, local_ip_address):
...
+            request = self.PortmapRequest(public_port, private_port, lease_duration)
+            request.sendto(sock, (self._gateway_address, self.REQUEST_PORT))
+
+            try:
+                response = self.PortmapResponse(message=sock.recv(16))
+                return response.result
+
+            except socket.timeout:
+                pass
+
```
homedirectory, line 139: What about repeating the request and increasing the timeout, as per the NAT-PMP spec?

homedirectory (changes requested):

pynicotine/portmapper.py
slook: Since the spec requires the timeout is doubled up to 9 times, the resulting 64 seconds is a long time for the thread to be potentially waiting, as this might might prevent shutdown from occurring as expected in the event of `want_abort` becoming true in the main networking thread, so this would have to be accounted for.

It might be necessary to use the `event.schedule()` method for this purpose, or some other similar solution.

homedirectory: @slook
Why not simply use `PortMapper.add_port_mapping(blocking=True)` so that it never blocks? I'm not very familiar with concurrent programming, so please have mercy.

slook: It could well be simpler to avoid spawning any new threads or scheduled events that have the potential to go into a zombie state or otherwise cause spurious random behaviour.

My instinct tells me that implementing a retry mechanism as defined in the spec isn't really going be that useful in our use case, because from my experience if a local connection does not succeed the first time then it will never work no matter how many successive attempts are made.

However, for me this is entirely theoretical because I don't know about NAT-PMP and cannot test this module.

Please let us know what happens when you run this branch, and push a PR with any modifications/corrections that are required for it to work properly (i.e. without unrecoverable errors/failures), whilst considering that new functionality can always be added at a later date, since it is better for the implementation to be incomplete rather than incorrect.

mathiascode: The reason we block when forwarding ports, is because we want to have an open port before we connect to the server and other peers.

I'm not sure how much of a difference it would make if we follow the spec to the teeth, or just increase the timeout of the first try to, say, 1 second, like we do with UPnP. With my router, the first try has never failed so far.

pynicotine/portmapper.py
mathiascode: I believe NAT-PMP is available on most routers in addition to UPnP, and it's simpler and faster than UPnP, which is why I opted to try it first.

pynicotine/portmapper.py
```diff
@@ -41,7 +45,127 @@ def set_port(self, port, local_ip_address):
...
+        def __bytes__(self):
+
+            return struct.pack(
+                '!BBHHHI',
+                self.VERSION,
+                self.TCP_OP_CODE,
+                self.RESERVED_VALUE,
+                self._public_port,
+                self._private_port,
```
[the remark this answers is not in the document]
mathiascode: Good catch!

pynicotine/portmapper.py
slook: It seems reasonable enough.

pynicotine/portmapper.py
slook: That does make some logical sense.

My only concern might be that if users have been used to seeing an entry in their UPnP configuration, then this would potentially change that behaviour, albeit it is inconsequential once the port has been established in most cases.

pynicotine/portmapper.py
homedirectory: I implemented a simple retry mechanism and set max attempts to 2, which in total gives 0.75 seconds of timeout. I then ran this branch with my modifications and ensured that everything works as expected, both when port mapping fails and when it succeeds. I also added debug log messages.

@slook You wrote:
> Please let us know what happens when you run this branch, and **push a PR** with any modifications/corrections that are required for it to work properly

Do you want me to push a new PR and close my original one? Or perhaps I can somehow contribute to this PR (#2408) to group the commits nicely.

mathiascode: I could merge this PR as-is now, so you can make further changes on top of master in a separate PR, if it sounds fine?

pynicotine/portmapper.py
mathiascode: We can test this behavior in master for a while, and see how it works.

pynicotine/portmapper.py
slook: @homedirectory I suggest you make a new PR to implement the retry mechanism to add yourself to contributors.

Agreed up to 3 retry attempts would be wise to implement, as this may help account for sleepy network devices and such, without creating undue delay whilst establishing a connection.

homedirectory: @mathiascode 
> I could merge this PR as-is now, so you can make further changes on top of master in a separate PR, if it sounds fine?

Makes sense, go ahead.

PR #2414, opened 2023-02-25
portmapper.py: repeat NAT-PMP requests with increasing timeout

homedirectory:
As per discussion in #2408, *send with retry* mechanism was implemented for NAT-PMP port-mapping requests. The maximum number of retry attempts was set to 2, which should be enough (the spec says 9).

### DIFF
--- a/pynicotine/portmapper.py
+++ b/pynicotine/portmapper.py
@@ -51,8 +51,9 @@ class NATPMP(BaseImplementation):
 
     NAME = "NAT-PMP"
     REQUEST_PORT = 5351
-    REQUEST_TIMEOUT = 1  # 1 second
     SUCCESS_RESULT = 0
+    REQUEST_INIT_TIMEOUT = 0.250 # seconds
+    REQUEST_ATTEMPTS = 2 # spec says 9, but 2 should be enough
 
     class PortmapResponse:
 
@@ -126,17 +127,25 @@ class NATPMP(BaseImplementation):
 
         with socket.socket(family=socket.AF_INET, type=socket.SOCK_DGRAM, proto=socket.IPPROTO_UDP) as sock:
             sock.bind((self.local_ip_address, 0))
-            sock.settimeout(self.REQUEST_TIMEOUT)
 
             request = self.PortmapRequest(public_port, private_port, lease_duration)
-            request.sendto(sock, (self._gateway_address, self.REQUEST_PORT))
 
-            try:
-                response = self.PortmapResponse(message=sock.recv(16))
-                return response.result
+            timeout = self.REQUEST_INIT_TIMEOUT
+            i = 1
+            while i <= self.REQUEST_ATTEMPTS:
+                sock.settimeout(timeout)
+                request.sendto(sock, (self._gateway_address, self.REQUEST_PORT))
 
-            except socket.timeout:
-                pass
+                try:
+                    response = self.PortmapResponse(message=sock.recv(16))
+                    return response.result
+
+                except socket.timeout:
+                    timeout *= 2
+                    log.add_debug(f"NAT-PMP: Attempt {i} timed out, increased timeout to {timeout}s")
+                    i += 1
+
+        log.add_debug(f"NAT-PMP: Fatal: all attempts timed out")
 
         return None
 
--- a/pynicotine/portmapper.py
+++ b/pynicotine/portmapper.py
@@ -129,9 +129,8 @@ class NATPMP(BaseImplementation):
             sock.bind((self.local_ip_address, 0))
             request = self.PortmapRequest(public_port, private_port, lease_duration)
             timeout = self.REQUEST_INIT_TIMEOUT
-            i = 1
 
-            while i <= self.REQUEST_ATTEMPTS:
+            for i in range(1, self.REQUEST_ATTEMPTS + 1):
                 sock.settimeout(timeout)
                 request.sendto(sock, self._gateway_address, self.REQUEST_PORT, i)
 
@@ -141,7 +140,6 @@ class NATPMP(BaseImplementation):
 
                 except socket.timeout:
                     timeout *= 2
-                    i += 1
 
         log.add_debug(f"NAT-PMP: Giving up, all {self.REQUEST_ATTEMPTS} portmap requests timed out")
 
--- a/pynicotine/portmapper.py
+++ b/pynicotine/portmapper.py
@@ -85,10 +85,10 @@ class NATPMP(BaseImplementation):
             self._private_port = private_port
             self._lease_duration = lease_duration
 
-        def sendto(self, sock, addr, port, num_attempt):
+        def sendto(self, sock, addr, num_attempt):
 
             msg = bytes(self)
-            sock.sendto(msg, (addr, port))
+            sock.sendto(msg, addr)
 
             log.add_debug(f"NAT-PMP: Portmap request attempt {num_attempt} of {NATPMP.REQUEST_ATTEMPTS}: {msg}")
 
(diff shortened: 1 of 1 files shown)


pynicotine/portmapper.py
```diff
@@ -126,17 +127,25 @@ def _request_port_mapping(self, public_port, private_port, lease_duration):
...
-            request.sendto(sock, (self._gateway_address, self.REQUEST_PORT))
 
-            try:
-                response = self.PortmapResponse(message=sock.recv(16))
-                return response.result
+            timeout = self.REQUEST_INIT_TIMEOUT
+            i = 1
+            while i <= self.REQUEST_ATTEMPTS:
+                sock.settimeout(timeout)
+                request.sendto(sock, (self._gateway_address, self.REQUEST_PORT))
```
slook, line 137: suggest adding an argument to log the number of the current retry attempt `i` in `sendto()` instead of here...

`request.sendto(sock, (self._gateway_address, self.REQUEST_PORT, i))`


```
        def sendto(self, sock, addr, num_attempt):

            msg = bytes(self)
            sock.sendto(msg, addr)

            log.add_debug(f"NAT-PMP: Portmap request attempt {num_attempt} of {self.REQUEST_ATTEMPTS}: {msg}")
```

Then additional log lines would not be needed, that saves on the amount of log lines.

pynicotine/portmapper.py
```diff
@@ -126,17 +127,25 @@ def _request_port_mapping(self, public_port, private_port, lease_duration):
...
 
             request = self.PortmapRequest(public_port, private_port, lease_duration)
-            request.sendto(sock, (self._gateway_address, self.REQUEST_PORT))
 
-            try:
-                response = self.PortmapResponse(message=sock.recv(16))
-                return response.result
+            timeout = self.REQUEST_INIT_TIMEOUT
+            i = 1
+            while i <= self.REQUEST_ATTEMPTS:
```
slook, line 135: suggest `for num_attempt in range(self.REQUEST_ATTEMPTS):` would eliminate the need to define `i`.

Also, an empty line is desired above the block.

pynicotine/portmapper.py
```diff
@@ -51,8 +51,9 @@ class NATPMP(BaseImplementation):
 
     NAME = "NAT-PMP"
     REQUEST_PORT = 5351
-    REQUEST_TIMEOUT = 1  # 1 second
     SUCCESS_RESULT = 0
+    REQUEST_INIT_TIMEOUT = 0.250 # seconds
+    REQUEST_ATTEMPTS = 2 # spec says 9, but 2 should be enough
```
slook, line 56: keep constants in alphabetical order.

2 spaces before code comments.

pynicotine/portmapper.py
```diff
@@ -126,17 +127,25 @@ def _request_port_mapping(self, public_port, private_port, lease_duration):
...
+                    return response.result
+
+                except socket.timeout:
+                    timeout *= 2
+                    log.add_debug(f"NAT-PMP: Attempt {i} timed out, increased timeout to {timeout}s")
+                    i += 1
 
-            except socket.timeout:
-                pass
+        log.add_debug(f"NAT-PMP: Fatal: all attempts timed out")
```
slook, line 148: `f"NAT-PMP: Giving up, all {self.REQUEST_ATTEMPTS} attempts to map the port timed out"`

This might be more friendly to help troubleshooting end users who may not know what NAT-PMP is.

pynicotine/portmapper.py
```diff
@@ -51,8 +51,9 @@ class NATPMP(BaseImplementation):
 
     NAME = "NAT-PMP"
     REQUEST_PORT = 5351
-    REQUEST_TIMEOUT = 1  # 1 second
     SUCCESS_RESULT = 0
+    REQUEST_ATTEMPTS = 2 # spec says 9, but 2 should be enough
+    REQUEST_INIT_TIMEOUT = 0.250 # seconds
```
slook, line 56: Still a couple of linting errors to resolve:
https://github.com/nicotine-plus/nicotine-plus/actions/runs/4269222146/jobs/7432291890#step:5:12

pynicotine/portmapper.py
```diff
@@ -126,17 +127,23 @@ def _request_port_mapping(self, public_port, private_port, lease_duration):
...
-            request.sendto(sock, (self._gateway_address, self.REQUEST_PORT))
+            timeout = self.REQUEST_INIT_TIMEOUT
+            i = 1
 
-            try:
-                response = self.PortmapResponse(message=sock.recv(16))
-                return response.result
+            while i <= self.REQUEST_ATTEMPTS:
+                sock.settimeout(timeout)
+                request.sendto(sock, (self._gateway_address, self.REQUEST_PORT), i)
```
slook, line 136: I don't think the internal parenthesis are necessary.

homedirectory: Do you mean `(self._gateway_address, self.REQUEST_PORT)`? How would you indicate that it's a tuple then? I tried removing them and the interpreter considered the tuple parts as separate function arguments.

slook: Our sendto() function does not require a tuple,
it takes 4 arguments `(self, sock, addr, num_attempt)`.

pynicotine/portmapper.py
```diff
@@ -126,17 +127,23 @@ def _request_port_mapping(self, public_port, private_port, lease_duration):
...
+
+                try:
+                    response = self.PortmapResponse(message=sock.recv(16))
+                    return response.result
+
+                except socket.timeout:
```
slook, line 142: Are exceptions other than `socket.timeout` possible?

If so, then there is a chance of an infinate loop here, because `i` will never increment in the case of some other error.

pynicotine/portmapper.py
```diff
@@ -126,17 +127,23 @@ def _request_port_mapping(self, public_port, private_port, lease_duration):
...
-
             request = self.PortmapRequest(public_port, private_port, lease_duration)
-            request.sendto(sock, (self._gateway_address, self.REQUEST_PORT))
+            timeout = self.REQUEST_INIT_TIMEOUT
+            i = 1
 
-            try:
-                response = self.PortmapResponse(message=sock.recv(16))
-                return response.result
+            while i <= self.REQUEST_ATTEMPTS:
```
slook, line 134: Suggest to avoid any risk of an infinite loop, do away with `i`:

`for num_attempt in range(1, self.REQUEST_ATTEMPTS):`

pynicotine/portmapper.py
mathiascode: In any case, this is pre-existing code that affects UPnP as well. It's not relevant to this PR.

pynicotine/portmapper.py
homedirectory: You're right. I took care of it.

pynicotine/portmapper.py
slook: It doesn't really make any difference either way.

My mistake I missed that `addr` is in fact a tuple of `(addr, port)`

pynicotine/portmapper.py
mathiascode: An unhandled exception would break the loop.

mathiascode: But I do agree that range() is the better option.

slook: How many total attempts does `self.REQUEST_ATTEMPTS + 1` result in?

mathiascode: 2

mathiascode (approved):

slook (approved):

PR #2416, opened 2023-02-26
gtkgui/search.py: use correct fullpath column index

### LINKED ISSUE #2415: Critical error upon opening File Properties of search results
### Describe the bug
Upon clicking on *File Properties* menu item in the *Search Files* window to inspect the properties of a search result a critical error is encountered:
```
Nicotine+ Version: 3.3.0.dev4
GTK Version: 4.8.3
Python Version: 3.10.9 (linux)

Type: <class 'AttributeError'>
Value: 'int' object has no attribute 'rsplit'
Traceback:   File "/opt/git-cloned/nicotine-plus/pynicotine/gtkgui/search.py", line 1280, in on_file_properties
    directory, filename = virtual_path.rsplit("\\", 1)
```

### Expected behavior
A window with file properties is opened.

Right-click on any search result, click *File Properties*.

### Additional context
The encounter of this critical error was also causing Nicotine+ to hang up upon quitting.

homedirectory:
Resolve #2415.
Commit [d14494ac](https://github.com/nicotine-plus/nicotine-plus/commit/d14494ac316f3813abdc835df6ac2e5187c306ce) introduced a new column -- file type icon, which shifted some columns, changing their index. Therefore, the `fullpath` column is now at `12`. This change was reflect in a number of places in the code, but the one causing this bug was missed.

### DIFF
--- a/pynicotine/gtkgui/search.py
+++ b/pynicotine/gtkgui/search.py
@@ -1276,7 +1276,7 @@ class Search:
         selected_length = 0
 
         for iterator in self.selected_results.values():
-            virtual_path = self.resultsmodel.get_value(iterator, 11)
+            virtual_path = self.resultsmodel.get_value(iterator, 12)
             directory, filename = virtual_path.rsplit("\\", 1)
             file_size = self.resultsmodel.get_value(iterator, 14)
             selected_size += file_size


slook:
For this very reason I think the intention is to change the column ID's to strings instead of integers, otherwise the task of adding new columns or improving existing ones is more difficult than it needs to be.

For now however, this fix will do nicely to solve the crash.

PR #2418, opened 2023-02-26
Add button to view personal profile

### LINKED ISSUE #2412: Visit my own user profile
Right now when I want to open my user profile, I search for myself on the chat. There should be a more straightforward way of doing this. I think a button next to the `Update Info` one in the `User Info` tab would be a good option.

There's probably another way I just haven't heard about. Is there documentation?

mathiascode:
Closes #2412

### DIFF
--- a/pynicotine/gtkgui/ui/mainwindow.ui
+++ b/pynicotine/gtkgui/ui/mainwindow.ui
@@ -979,7 +979,7 @@
                     <child>
                       <object class="GtkButton" id="_update_profile_button">
                         <property name="visible">True</property>
-                        <property name="action-name">app.update-user-profile</property>
+                        <property name="action-name">app.personal-profile</property>
                         <child>
                           <object class="GtkBox">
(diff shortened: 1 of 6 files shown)


pynicotine/gtkgui/ui/mainwindow.ui
```diff
@@ -993,7 +993,7 @@
                             <child>
```
slook, line 993: Suggest `document-edit-symbolic` -> `user-info-symbolic`

image

image

mathiascode: Unfortunately this icon is not available in Breeze, and is marked as legacy in the Adwaita icon theme: https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/tree/master/Adwaita/symbolic/legacy

How about `user-home-symbolic`?

slook: That works.

PR #2420, opened 2023-02-28
gtkgui/userbrowse.py: don't allow saving an empty shares list

slook:
+ Changed: "Save Shares List to Disk" button sensitivity is reactive to the folder list view having any items in it or not
+ Fixed: Don't allow the `Ctrl+S` keyboard shortcut whilst shares are loading or not present

### DIFF
--- a/pynicotine/gtkgui/userbrowse.py
+++ b/pynicotine/gtkgui/userbrowse.py
@@ -169,6 +169,7 @@ class UserBrowse:
             self.progress_bar,
             self.refresh_button,
             self.retry_button,
+            self.save_button,
             self.search_entry,
             self.share_size_label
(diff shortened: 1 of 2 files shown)


pynicotine/gtkgui/userbrowse.py
```diff
@@ -563,6 +565,7 @@ def set_finished(self):
         self.userbrowses.request_tab_changed(self.container)
         self.progress_bar.set_fraction(1.0)
         self.refresh_button.set_sensitive(True)
+        self.save_button.set_sensitive(bool(self.num_folders))
```
mathiascode, line 568:
```suggestion
        self.save_button.set_sensitive(not self.folder_tree_view.is_empty())
```

PR #2431, opened 2023-03-05
userbrowse.py: minor optimization folder tree view

slook:
+ Changed: Use path instead of iterator for expanding and collapsing, path is already provided by the tree view widget


mathiascode:
path, column should be renamed to iterator, column_id

slook:
Done. Thanks, I see that makes more sense, I thought it was odd that iterator seemed interchangeable with path.

PR #2434, opened 2023-03-06
treeview.py: new expand_root_rows() method implemented for Browse Shares

slook:
Resolves https://github.com/nicotine-plus/nicotine-plus/discussions/1881#discussioncomment-3263521

+ Added: Expand into first level upon initial loading of Browse Shares list, instead of fully collapsed.

The required functionality already exists for `"folder_grouping"` view mode, the new method can be used for that as well, as commented in the legacy function.

The Expand / Collapse All button function is unaffected, it is still useful to remember the expand preference if desired.

### DIFF
--- a/pynicotine/gtkgui/widgets/treeview.py
+++ b/pynicotine/gtkgui/widgets/treeview.py
@@ -487,6 +487,16 @@ class TreeView:
     def collapse_all_rows(self):
         self.widget.collapse_all()
 
+    def expand_root_rows(self):
+
+        model = self.model
+        iterator = model.get_iter_first()
+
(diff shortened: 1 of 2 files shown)


pynicotine/gtkgui/widgets/treeview.py
```diff
@@ -487,6 +487,16 @@ def expand_all_rows(self):
     def collapse_all_rows(self):
         self.widget.collapse_all()
 
+    def expand_root_rows(self):
+
+        model = self.widget.get_model()
```
mathiascode, line 492: self.model

PR #2435, opened 2023-03-06
gtkgui/userbrowse.py: remember selection after refresh

slook:
If the file and/or folder still exists in the refreshed shares, then it was frustrating that the selection is lost after doing a refresh operation, because finding the desired item again can be difficult if the shares are large, as there are no location history features.

+ Added: Return to the currently selected location (file and/or folder) after doing a Refresh
+ Changed: Add a trailing slash onto to the end of folder path copy to clipboard string
+ Fixed: Grab the focus onto the relevant list view widget when entering a URL even if the tab is already open

The clipboard copy code is slightly repurposed for this. The new getter functions are designed to be useful for other purposes too, such as for storing location history, status text, tooltips, etc.

If the selected file doesn't exist anymore, then the folder is selected at least. Only a single file selection is recalled (upper-most).

### DIFF
--- a/pynicotine/gtkgui/userbrowse.py
+++ b/pynicotine/gtkgui/userbrowse.py
@@ -489,8 +489,8 @@ class UserBrowse:
         if not iterator:
             return
 
-        # Scroll to the requested file
-        self.file_list_view.select_row(iterator)
+        # Scroll to the requested file and grab focus to it
+        self.file_list_view.select_row(iterator, should_expand=True)
 
     def shared_file_list(self, msg):
 
@@ -610,6 +610,14 @@ class UserBrowse:
             filesize = self.file_list_view.get_row_value(iterator, "size_data")
 
             self.selected_files[rawfilename] = filesize
+
+    def get_selected_folder_path(self):
+        return f'{self.selected_folder or ""}\\'
+
+    def get_selected_file_path(self):
+        selected_folder = self.get_selected_folder_path()
+        selected_file = next(iter(self.selected_files)) if self.selected_files else ""
+        return f"{selected_folder}{selected_file}"
 
     """ Search """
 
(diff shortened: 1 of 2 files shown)


pynicotine/gtkgui/userbrowse.py
```diff
@@ -489,8 +489,8 @@ def browse_queued_path(self):
         if not iterator:
             return
 
-        # Scroll to the requested file
-        self.file_list_view.select_row(iterator)
+        # Scroll to the requested file and grab focus to it
+        self.file_list_view.select_row(iterator, should_expand=True)
```
mathiascode, line 493: This feels strange, since we don't have expanders in the file list.

slook: I know, this is my least liked element of this PR.

Functionally it works perfectly, but it feels like a hack.

Suggest perhaps renaming/reworking the method in treeview.py to seperate:
- `should_scroll`
- `should_expand`
- `should_focus`

I cannot think of a scenario where we want to expand but not focus, but I could be mistaken.

mathiascode: I think you could just call grab_focus() when should_focus is True. There shouldn't be any situations where we don't want to focus the widget when scrolling to a row.

slook: No, we can't do that because it breaks the search next match behaviour.

mathiascode: That's a shame. I'll have to take a closer look, since the select_row() method is becoming dangerously large.

slook: I reverted the changes to treeview.py in favour of grabbing the focus in browse_queued_path()

pynicotine/gtkgui/userbrowse.py
```diff
@@ -1044,25 +1047,17 @@ def on_file_properties(self, _action, _state, all_files=False):
...
 
     def on_copy_file_path(self, *_args):
-
-        if self.selected_folder is None or not self.selected_files:
-            return
-
-        text = "\\".join([self.selected_folder, next(iter(self.selected_files))])
-        copy_text(text)
+        file_path = self.get_selected_file_path()
+        copy_text(f"{self.user}\\{file_path}")
```
mathiascode, line 1051: Changes to paths should not be included in this PR

slook: reverted.

PR #2436, opened 2023-03-07
gtkgui/userbrowse.py: fix ValueError in browse_queued_path()

slook:
+ Fixed: Crash on entering an invalid Soulseek URL that only contains a root path, for example `slsk://user/folder`

```
Nicotine+ Version: 3.3.0.dev4
GTK Version: 3.24.24
Python Version: 3.9.2 (linux)

Type: <class 'ValueError'>
Value: not enough values to unpack (expected 2, got 1)
Traceback:   File "/home/user/Git/nicotine-plus/pynicotine/gtkgui/application.py", line 783, in on_process_thread_events
    events.process_thread_events()
  File "/home/user/Git/nicotine-plus/pynicotine/events.py", line 256, in process_thread_events
    self.emit(event_name, *args, **kwargs)
  File "/home/user/Git/nicotine-plus/pynicotine/events.py", line 226, in emit
    function(*args, **kwargs)
  File "/home/user/Git/nicotine-plus/pynicotine/gtkgui/userbrowse.py", line 149, in shared_file_list
    page.shared_file_list(msg)
  File "/home/user/Git/nicotine-plus/pynicotine/gtkgui/userbrowse.py", line 501, in shared_file_list
    self.browse_queued_path()
  File "/home/user/Git/nicotine-plus/pynicotine/gtkgui/userbrowse.py", line 476, in browse_queued_path
    folder, filename = self.queued_path.rsplit("\\", 1)
```

```
Nicotine+ Version: 3.2.9
GTK Version: 3.22.11
Python Version: 3.5.3 (default, Nov  4 2021, 15:29:10) 
[GCC 6.3.0 20170516]

Type: <class 'ValueError'>
Value: not enough values to unpack (expected 2, got 1)
Traceback:   File "/home/user/Git/nicotine-plus/3.2.x/pynicotine/gtkgui/userbrowse.py", line 147, in _shared_file_list
    self.pages[user].shared_file_list(msg)
  File "/home/user/Git/nicotine-plus/3.2.x/pynicotine/gtkgui/userbrowse.py", line 523, in shared_file_list
    self.browse_queued_path()
  File "/home/user/Git/nicotine-plus/3.2.x/pynicotine/gtkgui/userbrowse.py", line 493, in browse_queued_path
    folder, filename = self.queued_path.rsplit("\\", 1)
```

or `slsk://user/` will also do it, either text entry or link click handler. -- Edit: Fixed.

@mathiascode Critical Error (master; 3.2.x)

PR #2437, opened 2023-03-07
userbrowse.py: open standard Soulseek path (not URL)

slook:
Soulseek usernames don't seem to allow backslashes in them, so we can use `"\"` as a path separator to navigate to a folder\file within a users shares...

+ Added: Browse to a standard Soulseek path if the entry contains text separated with backslashes `\`
+ ~Fixed: Critical ValueError in browse_queued_path() when browsing to a root share folder, also affects URL browsing in master~

A Soulseek path is assumed to have the syntax `username\share name\folder name\file name`.

Any optional leading `\` before the username are stripped.

If the path ends with a `\` then it is a folder, otherwise it is a file.

If the path only contains a single element, then it is a root folder, regardless if it ends with a `\` or not.

Forward slashes are assumed to be valid characters and are not replaced.

This PR also contains an Alternative fix to #2436


pynicotine/gtkgui/userbrowse.py
```diff
@@ -473,7 +473,11 @@ def browse_queued_path(self):
         if self.queued_path is None:
             return
 
-        folder, filename = self.queued_path.rsplit("\\", 1)
+        if "\\" in self.queued_path:
+            folder, filename = self.queued_path.rsplit("\\", 1)
+        else:
+            folder, filename = self.queued_path, ""
```
slook, line 479: The crash in #2436 can be avoided here like this

pynicotine/plugins/core_commands/__init__.py
```diff
@@ -152,8 +152,8 @@ def __init__(self, *args, **kwargs):
                 "description": _("Browse files of user"),
                 "disable": ["cli"],
                 "group": _CommandGroup.USERS,
-                "usage": ["<user>"],
-                "usage_private_chat": ["[user]"]
+                "usage": ["<user\\folder or url>"],
+                "usage_private_chat": ["[user\\folder]", "[url]"]
```
mathiascode, line 156: Is this syntax based on some existing convention? Not sure I like it, to me it reads like two positional arguments.

mathiascode:
Why have this in addition to URLs? Feels like two solutions competing against each other.

pynicotine/plugins/core_commands/__init__.py
slook: Agreed, I shouldn't have put it in the Network Filter commands either. It does look confusing.

slook:
> Why have this in addition to URLs?

Because we should use the text entry as an address bar to display the current location, as would be expected of a file manager like interface, and perhaps even implement back/forward history or bookmark favorite location functions, etc.

> two solutions competing against each other

Not really. Both syntaxes are present, so both should be supported. The Windows style paths are more native to the network than the Nicotine+ specific URLs.

A Qt user might need to provide a Windows-like Soulseek path to an N+ user.

mathiascode:
> > Why have this in addition to URLs?
> 
> Both syntaxes are present, so both should be supported. The Windows style paths are more native to the network than the Nicotine+ specific URLs.
> 
> The 'Copy Path' function does not otherwise serve a useful purpose unless it can be pasted here.
> 
> A Qt user might need to provide a Windows-like Soulseek path to an N+ user.

I need to check the history of the 'copy path' menu items, but I'm pretty sure the main purpose was to easily copy the path for use in other programs. Perhaps the file properties dialog would be enough for this nowadays.

There are no paths that start with a username, so you're basically inventing an alternative to slsk URLs. I'd rather not add this.

slook:
All good points. However, I feel like the current current interface is well below par, even if it is not intended to be a fully fledged file manager.

> Is the main issue that you can't see parent folders?

Yes, determining the containing parent folder is difficult. Maybe it would help if the tool tip could say the whole path.

mathiascode:
Closing, as this isn't something I want to add right now.

mathiascode:
> Yes, determining the containing parent folder is difficult. Maybe it would help if the tool tip could say the whole path.

Could be interesting to experiment with a path bar with clickable parts similar to Nautilus, but clicking them would scroll to the desired folder. The path bar could be placed above the folder tree view.

Tooltips showing the full path should be possible now, since we started storing the full path in the treeview model.

PR #2440, opened 2023-03-11
CI: test Python '3.7'

slook:
+ Fixed: Mismatched quote mark in syntax of `'3.7'`
+ Changed: Use oldest available revision of Python `'3.6.7'`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -74,7 +74,7 @@ jobs:
 
       - name: Install dependencies
         run: |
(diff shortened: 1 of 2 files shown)


mathiascode:
> * Changed: Use oldest available revision of Python `'3.6.7'`

It should pick the latest version already.

slook:
Surely the _oldest_ revision (3.6.7) is wanted, rather than the newest revision (3.6.15)?

mathiascode:
3.6.8 was the last bugfix release, anything past this is just a security fix release. Afaik, Python hasn't broken API in patch releases, so I'd rather not cling on to old releases.

PR #2452, opened 2023-03-21
treeview.py: simplify select_row() method

mathiascode:


mathiascode:
@slook Is this fine?

slook:
Yes, it makes sense.

PR #2453, opened 2023-03-22
userbrowse.py: cleanup browse_local_shares()

slook:
Code cleanup. No functional changes.

userbrowse.py: A single method can handle any local share group without hard-coding the share group name.

shares.py: If the share group name does not exist (for example "public"), then fall back to "normal".

### DIFF
--- a/pynicotine/userbrowse.py
+++ b/pynicotine/userbrowse.py
@@ -72,7 +72,7 @@ class UserBrowse:
         del self.user_shares[user]
         events.emit("user-browse-remove-user", user)
 
-    def parse_local_shares(self, username, msg):
+    def _parse_local_shares(self, username, msg):
         """ Parse a local shares list and show it in the UI """
 
         built = msg.make_network_message()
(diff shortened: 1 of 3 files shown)


pynicotine/userbrowse.py
```diff
@@ -81,31 +81,22 @@ def parse_local_shares(self, username, msg):
 
         events.emit_main_thread("shared-file-list-response", msg)
 
-    def browse_local_public_shares(self, path=None, new_request=None):
-        """ Browse your own public shares """
+    def browse_local_shares(self, path=None, local_shares_type=None, new_request=False):
```
mathiascode, line 84: local_shares_type -> shares_type

pynicotine/userbrowse.py
```diff
@@ -81,31 +81,22 @@ def parse_local_shares(self, username, msg):
...
-
-        username = config.sections["server"]["login"] or "Default"
-
-        if username not in self.user_shares or new_request:
-            msg = core.shares.get_compressed_shares_message("buddy")
             Thread(
-                target=self.parse_local_shares, args=(username, msg), name="LocalBuddyShareParser", daemon=True
+                target=self._parse_local_shares,
+                args=(username, msg),
+                name=f"Local{local_shares_type.title()}ShareParser",
```
mathiascode, line 95: Let's just go with LocalShareParser.

pynicotine/userbrowse.py
```diff
@@ -119,11 +110,7 @@ def browse_user(self, username, path=None, local_shares_type=None, new_request=F
             user_share.clear()
 
         if username == (config.sections["server"]["login"] or "Default"):
-            if local_shares_type == "normal":
-                self.browse_local_public_shares(path, new_request)
-                return
-
-            self.browse_local_buddy_shares(path, new_request)
+            self.browse_local_shares(path, local_shares_type or "buddy", new_request)
```
mathiascode, line 113:
```suggestion
            self.browse_local_shares(path, local_shares_type, new_request)
```

mathiascode: and local_shares_type="buddy" instead of None

pynicotine/userbrowse.py
mathiascode: We can improve this later, but `local_` is redundant in this case.

pynicotine/userbrowse.py
mathiascode: Assign a default value "buddy" to the local_shares_type parameter instead of None.

PR #2457, opened 2023-03-24
Result Filters: recall cleared filters undo toggle

mathiascode:
Continuation of https://github.com/nicotine-plus/nicotine-plus/pull/2404
Closes https://github.com/nicotine-plus/nicotine-plus/issues/1400


mathiascode:
I realized that this approach doesn't work with the file type groups...

pynicotine/gtkgui/search.py
```diff
@@ -1587,20 +1650,21 @@ def on_refilter(self, *_args):
         # Replace generic file type filters with real file extensions
         file_type_filters = filters["filtertype"]
 
-        for filter_name, file_extensions in self.FILTER_GENERIC_FILE_TYPES:
-            excluded_filter_name = f"!{filter_name}"
```
slook, line 1591: It might be best for this code to exist in its own separate function(s).

pynicotine/gtkgui/search.py
```diff
@@ -1626,23 +1690,15 @@ def on_filter_entry_icon_press(self, entry, *_args):
...
-        self.clearing_filters = True
-
-        for widget in self.filter_comboboxes.values():
-            widget.get_child().set_text("")
-
-        self.filter_free_slot_button.set_active(False)
+        self.set_filters(self.filters_undo)
 
         if self.filters_button.get_active():
             self.filter_include_combobox.get_child().grab_focus()
```
slook, line 1698: It is not wanted to grab the focus into the text entry when restoring the filters.

pynicotine/gtkgui/search.py
```diff
@@ -303,14 +303,14 @@ class Search:
         "filterlength": (">15:00", ">8:00 <=15:00", ">5:00 <=8:00", ">2:00 <=5:00", "<=2:00")
     }
     EMPTY_FILTERS = {
```
slook, line 305: constants alphabetical order, move to top.

pynicotine/gtkgui/search.py
slook: Well guess so yes, albeit that does not really provide much value either as the current selection is not remembered when refiltering, but it would be less disconcerting.

It

pynicotine/gtkgui/search.py
mathiascode: Not addressing this right now.

pynicotine/gtkgui/search.py
mathiascode: Leaving this as-is for now, can be improved in the future if necessary.

slook: Sure, maybe now the button does have a function, it would be no harm to allow it to be focused, such that the space bar can be used to toggle clear/restore.

PR #2458, opened 2023-03-24
gtkgui/search.py: cleanup populate_default_filters()

slook:
+ Fixed: Crash on search with default length or type filter set
+ Changed: Loop iteration instead of elif ladder

correction to #2457

### DIFF
--- a/pynicotine/gtkgui/search.py
+++ b/pynicotine/gtkgui/search.py
@@ -309,10 +309,10 @@ class Search:
         "filterout": (None, ""),
         "filtersize": (None, ""),
         "filterbr": (None, ""),
+        "filterslot": (False, False),
         "filtercc": (None, ""),
         "filtertype": (None, ""),
-        "filterlength": (None, ""),
-        "filterslot": (False, False)
+        "filterlength": (None, "")
     }
 
     def __init__(self, searches, text, token, mode, mode_label, show_page):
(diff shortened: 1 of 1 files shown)


pynicotine/gtkgui/search.py
```diff
@@ -612,39 +612,23 @@ def update_filter_widgets(self):
...
         sfilter = config.sections["searches"]["defilter"]
         num_filters = len(sfilter)
-        stored_filters = {}
+        stored_filters = self.FILTERS_EMPTY
 
-        if num_filters > 0:
-            stored_filters["filterin"] = (None, str(sfilter[0]))
+        # Convert from list to dict
+        for i, filter_id in enumerate(stored_filters):
+            if num_filters > i == 4:
```
mathiascode, line 626: I would break when i <= num_filters, and check filter_id == "filterslot" here.

pynicotine/gtkgui/search.py
```diff
@@ -612,39 +612,23 @@ def update_filter_widgets(self):
...
-    def populate_filters(self):
+    def populate_default_filters(self):
 
         if not config.sections["searches"]["enablefilters"]:
             return
 
         sfilter = config.sections["searches"]["defilter"]
         num_filters = len(sfilter)
-        stored_filters = {}
+        stored_filters = self.FILTERS_EMPTY
```
mathiascode, line 622:
```suggestion
        stored_filters = self.FILTERS_EMPTY.copy()
```

otherwise you modify the constant

pynicotine/gtkgui/search.py
slook: You mean break when `i >= num_filters` right?

PR #2459, opened 2023-03-25
userinfo.py: automatically resize user picture

mathiascode:
Also remove the option to manually zoom. We're many things, but not an image viewer...


mathiascode:
GTK 3 drawing should now be as smooth as GTK 4.

PR #2460, opened 2023-03-25
GUI: add clipboard module

mathiascode:


pynicotine/gtkgui/widgets/clipboard.py
```diff
@@ -17,6 +17,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from gi.repository import Gdk
+from gi.repository import GObject
 from gi.repository import Gtk
```
slook, line 21: this import is only required if GTK_API_VERSION == 3

pynicotine/gtkgui/widgets/clipboard.py
```diff
@@ -25,12 +26,25 @@
 """ Clipboard """
 
 
+if GTK_API_VERSION >= 4:
+    _clipboard = Gdk.Display.get_default().get_clipboard()
+else:
+    _clipboard = Gtk.Clipboard.get(Gdk.SELECTION_CLIPBOARD)
```
slook, line 32: It might be better to free the Clipboard object assignment after each use?

pynicotine/gtkgui/widgets/clipboard.py
mathiascode: It's already imported elsewhere, so it doesn't matter.

pynicotine/gtkgui/widgets/clipboard.py
mathiascode: GTK keeps a reference to it too, so it doesn't matter.

PR #2466, opened 2023-03-28
popupmenu selected files human count and size

slook:
+ Added: "1 File Selected" or "_n_ Files Selected" translatable strings for the file popup menus
+ Removed: "_n_ File(s) Selected" translatable string

The idea of including the numeric digit within singular version of the string is to help the translators. It also allows flexibility if a number is not desired, such as "No Files" or just the word "Folder" without a number, for example.

In the long term, the proposed new pluralize() utility function should help to reduce the overall number of strings in the project.

image

image


mathiascode:
Apparently gettext.install() can also make ngettext() available, so you can `gettext.install(names=["ngettext"])` in i18n.py, then do something like:

```py
        self.items["selected_files"].set_label(
            ngettext("{num} File Selected", "{num} Files Selected", num_files).format(num=num_files)
        )
```

slook:
It would be nicer to wrap the suggested ngettext() implementation into a function which only requires the key "selected_files" to look-up the assigned strings. Otherwise it will be too cumbersome to use when defining menu items and labels etc.

mathiascode:
It depends. Ideally we would only use a string once in the codebase anyway.

mathiascode:
Closing in favor of #3397

PR #2502, opened 2023-05-21
Replace UTF8 hyphens with dashes

hboetes:
For some mysterious reason these two files contained some hyphens where dashes should be. Hard to spot of course.


mathiascode:
Thanks, but you're using a hyphen as the replacement character. I've made the proper changes in 6ca91d2f880eac061e3e4af62c7e04e25b43d51e.

PR #2508, opened 2023-05-27
textentry.py: fix chat entry dropdown without tab completion

slook:
+ **Fixed: Possible KeyError crash if a user leaves a joined room after chat completion preferences are altered:**
The KeyError was caused by attempting to remove a non-existant dropdown list item, if the option was previously disabled.

+ **Fixed: Allow chat entry dropdown-completion even if tab-completion is not enabled:**
The dropdown list was prevented from being populated, and thereby was not shown if the 'Enable tab-key completion' config setting was off, yet those two features are supposed to be independent of each-other.

### DIFF
--- a/pynicotine/gtkgui/widgets/textentry.py
+++ b/pynicotine/gtkgui/widgets/textentry.py
@@ -127,7 +127,7 @@ class ChatCompletion:
 
     def create_entry_completion(self):
 
-        self.entry_completion = Gtk.EntryCompletion(model=self.model)
+        self.entry_completion = Gtk.EntryCompletion(popup_single_match=False, model=self.model)
         self.entry_completion.set_text_column(0)
         self.entry_completion.set_match_func(self.entry_completion_find_match)
         self.entry_completion.connect("match-selected", self.entry_completion_found_match)
@@ -147,13 +147,8 @@ class ChatCompletion:
 
     def add_completion(self, item):
 
-        if not config.sections["words"]["tab"]:
-            return
-
-        if item in self.completion_list:
-            return
-
-        self.completion_list.append(item)
+        if config.sections["words"]["tab"] and item not in self.completion_list:
+            self.completion_list.append(item)
 
         if config.sections["words"]["dropdown"]:
             self.completion_iters[item] = self.model.insert_with_valuesv(-1, self.column_numbers, [item])
@@ -179,18 +174,17 @@ class ChatCompletion:
 
     def remove_completion(self, item):
 
-        if not config.sections["words"]["tab"]:
-            return
-
-        if item not in self.completion_list:
-            return
-
-        self.completion_list.remove(item)
+        if config.sections["words"]["tab"] and item in self.completion_list:
+            self.completion_list.remove(item)
 
         if not config.sections["words"]["dropdown"]:
             return
 
-        iterator = self.completion_iters[item]
+        iterator = self.completion_iters.get(item)
+
+        if iterator is None:
+            return
+
         self.model.remove(iterator)
         del self.completion_iters[item]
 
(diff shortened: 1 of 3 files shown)


slook:
Sorry I don't follow what you mean by your explanation, please propose a change in context of the code.

slook:
> Did the crash occur when the tab completion option was disabled[?]

No, I think it happened when it was enabled, after it had previously been disabled before.

pynicotine/gtkgui/widgets/textentry.py
```diff
@@ -147,13 +147,8 @@ def set_entry(self, entry):
...
     def add_completion(self, item):
 
-        if not config.sections["words"]["tab"]:
-            return
-
-        if item in self.completion_list:
-            return
-
-        self.completion_list.append(item)
+        if config.sections["words"]["tab"] and item not in self.completion_list:
```
mathiascode, line 150: The completions are always supposed to be updated when tab/popup completion is enabled/disabled (see https://github.com/nicotine-plus/nicotine-plus/blob/4d402687ecb7802668ea1eca522035540b74388c/pynicotine/gtkgui/dialogs/preferences.py#L2760-L2762), so `item not in self.completion_list` should ideally not be required.

pynicotine/gtkgui/widgets/textentry.py
```diff
@@ -179,18 +174,17 @@ def get_completions(part, completion_list):
...
 
-        if item not in self.completion_list:
+        if not config.sections["words"]["dropdown"]:
             return
 
-        self.completion_list.remove(item)
+        iterator = self.completion_iters.get(item)
 
-        if not config.sections["words"]["dropdown"]:
+        if iterator is None:
```
mathiascode, line 185: Ditto

pynicotine/gtkgui/widgets/textentry.py
slook: Is that not also needed to avoid duplicates? Such as if a user is both in a chat room and is a buddy, for example.

mathiascode: Duplicates are removed earlier: https://github.com/nicotine-plus/nicotine-plus/blob/4d402687ecb7802668ea1eca522035540b74388c/pynicotine/gtkgui/chatrooms.py#L1228-L1232

mathiascode: Sorry, I forgot that `add_completion()` has a check as well. That one should indeed remain in place.

slook: But there is another instance in user_joined_room() where items get added without checking for duplicates: 

https://github.com/nicotine-plus/nicotine-plus/blob/4d402687ecb7802668ea1eca522035540b74388c/pynicotine/gtkgui/chatrooms.py#L950-L951

mathiascode:
All rooms use the same `ChatCompletion` object, but only the focused room is supposed to make changes to it. I think the issue is that other rooms can currently manipulate the completion user list as well. If a user leaves multiple rooms that you've joined, the crash occurs.

slook:
Right, such as if a user logs off then.

The ideal solution is beyond the scope of this PR, it seems dumb to maintain two seperate models list+dict where we could just use ListStore for both modes of completion. That would require a refactor of the entire class.

Currently, the crash is avoided because the feature is prevented from working properly in all cases.

pynicotine/gtkgui/widgets/textentry.py
slook: Without the check it results in `TypeError: Argument 1 does not allow None as a value`

pynicotine/gtkgui/widgets/textentry.py
slook: The duplicate checking could be avoided if `completion_list = []` was instead `completion_words = set()`

mathiascode:
I've pushed some commits to master that should fix this issue.

mathiascode:
> The ideal solution is beyond the scope of this PR, it seems dumb to maintain two seperate models list+dict where we could just use ListStore for both modes of completion. That would require a refactor of the entire class.

Gtk.EntryCompletion and Gtk.ListStore are deprecated, so we should move to something else at some point.

Issue #1948, opened 2022-03-08
I need file sizes to be displayed in bytes, not interested in abbreviated MiB sizing.


rolex905:
Hello there, I just discovered the app and was looking forward to using it, but I can instantly see a major problem that is totally unacceptable for me:  results and file lists are being displayed in MiB format instead of exact BYTES the way Soulseek client does.

I have not been able to use Nicotine because of this.  When I search for files I am looking for very specific file sizes that match up with torrents I would like to also seed, so abbreviated format does not, and will never work for me.

Let me know if there's any way to toggle this back to the original functionality of displaying full file sizes.  Furthermore, clicking on file details within the results ALSO is displaying an abbreviated file size in MiB.  Why?

slook:
> I just discovered the app

Welcome :)

> file sizes that match up

This is a good point, I also like to refer the last 3 or so digits of a file to verify if it is identical to the original version of the file, or if it has been modified in some way. With audio files, slight difference in size usually indicates it has been re-tagged or the cover art is missing etc. Since the epoch, this has been a standard and expected feature of any program that manages files.

> I have not been able to use Nicotine because of this.

You are probably not the only one.

> clicking on file details [File Properties] within the results ALSO is displaying an abbreviated file size in MiB

I have this on my development fork so I shall request it be added into master.

> Why?

afiak, nobody else has ever raised this issue or desired to implement Bytes display in the GUI before. Thanks for raising it.

> Let me know if there's any way to toggle this back to the original functionality of displaying full file sizes

If you [clone and run from a local Git folder](https://github.com/nicotine-plus/nicotine-plus/blob/master/doc/TESTING.md#git), the File Search size column display is converted into KiB/MiB/GiB here:
https://github.com/nicotine-plus/nicotine-plus/blob/4a3c58e46bd96b32e15fea314e8b4087d464040d/pynicotine/gtkgui/search.py#L568

If you want thousands separators, change this line to:
`            h_size = humanize(size)`

Or if you just want a raw integer, just set h_size to size (which is stored for sorting purposes):
`            h_size = size`

Then the results look like this:
![image](https://user-images.githubusercontent.com/88614182/157340770-883f22c4-d7e7-4a10-bd22-fabf8801becd.png)

`            h_size = humanize(size) + " B"` to append a B on the end.

For Browse Shares it's here:
[...]

PR #2511, opened 2023-05-28
config: exact_file_sizes -> file_size_units

slook:
The proposed config name in https://github.com/nicotine-plus/nicotine-plus/commit/7bb5a123a89b6a522e448165688e244a20fdfd99 #1948 is not future proof.  This PR suggests changing the name and datatype of it to allow for the future implementation of more options other than `True` (Bytes), such as "B", "KiB", "MiB", etc.

+ Changed: `exact_file_sizes` -> `file_size_units` config setting, and change type from bool to str.
+ Changed: Toggle -> ComboBox in Preferences.
+ Added:  temporary migration code `True` -> `"B"` (only needed for 3.3.0 testers).
+ Added: New label and _("Automatic") and _("B (Bytes)") new translatable strings for displaying in UI.

This PR only implements two `""` or `"B"` values in config. Thus, the user functionality is not changed.

image

If we change this name now then it won't be needed to carry any migration code into stable, because the feature was only recently implemented in 3.3.0.dev. The migration code is temporarily added here for the convenience of testers.

After this PR it would be relatively** simple to implement other file_size_units (and speed units) if and when desired.

** although, suggestions welcome as to how to best do fixed units, see the TODO comment in human_speed_or_size().

### DIFF
--- a/pynicotine/config.py
+++ b/pynicotine/config.py
@@ -404,7 +404,7 @@ class Config:
                 "urgencyhint": True,
                 "file_path_tooltips": True,
                 "reverse_file_paths": True,
-                "exact_file_sizes": False
+                "file_size_units": ""
             },
             "private_rooms": {
                 "rooms": {}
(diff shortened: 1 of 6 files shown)


mathiascode:
I agree that a string option is probably a better idea, but I'll need convincing reasons for supporting units other than bytes in the future (SoulseekQt showing speeds in kB is not a valid reason on its own). Bytes makes sense because some users want to compare file sizes.

Until then, I want to keep the existing toggle for exact file sizes (https://github.com/nicotine-plus/nicotine-plus/pull/2512).

PR #2513, opened 2023-05-28
properly unset chat completion when disabled

slook:
+ Fixed: Allow the chat completion to be disabled if both modes are set then unset during the session.
+ Changed: Don't add items to the completions list when user joins room if completions are disabled.
+ Added: Make pop() return None upon removing a non-existent item, because that doesn't happen by default.

If completion was already enabled, then disabling was ineffectual if both modes are toggled off, as in the dropdown still appears and tab completion still happens, because the code in textentry.py to disable it was never reached in this edge case.

This PR allows the clear() code to be reached once after both of the options are unset during the session, or otherwise either need a dedicated clear completions function or to mark the preference label with "(requires a restart)".

Also, there is a performance improvement by not compiling the completion_list if it isn't needed.

----

Marking as draft, too many changes.

### DIFF
--- a/pynicotine/chatrooms.py
+++ b/pynicotine/chatrooms.py
@@ -385,16 +385,19 @@ class ChatRooms:
 
     def update_completions(self):
 
-        self.completion_list = [config.sections["server"]["login"]]
+        if not config.sections["words"]["tab"] and not config.sections["words"]["dropdown"]:
+            self.completion_list = []
+        else:
+            self.completion_list = [config.sections["server"]["login"]]
 
-        if config.sections["words"]["roomnames"]:
-            self.completion_list += self.server_rooms
+            if config.sections["words"]["roomnames"]:
+                self.completion_list += self.server_rooms
 
-        if config.sections["words"]["buddies"]:
-            self.completion_list += list(core.userlist.buddies)
+            if config.sections["words"]["buddies"]:
+                self.completion_list += list(core.userlist.buddies)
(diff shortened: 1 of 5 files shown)


pynicotine/chatrooms.py
```diff
@@ -385,16 +385,19 @@ def _user_left_room(self, msg):
 
     def update_completions(self):
 
-        self.completion_list = [config.sections["server"]["login"]]
+        if not config.sections["words"]["tab"] and not config.sections["words"]["dropdown"]:
```
mathiascode, line 388: I don't think these options should be referenced in the core.

How is performance for you since 91331231b2bb14418dbfa6aa16857e6a80ba804b?

slook: It's okay, the most expensive operation is always https://github.com/nicotine-plus/nicotine-plus/blob/6736dec3cc75b367f79ac97ecbbda8b1370f3c65/pynicotine/gtkgui/widgets/textentry.py#L195

it seems excessive to clear and compile and rebuild the entire set and model each time a change occurs. I can't help thinking that in the case of a single item changing (such as a user leaving or joining a room) it would be better to get the intersection of the sets (make a copy beforehand) and then just do an add/remove.

mathiascode: We could look closer into this at some point. Since completions are now properly unset in master, I'll close the PR.

Issue #2482, opened 2023-04-19
Public chat room feed: Don't notify on new messages + clickable room names


redactedscribe:
### Describe the solution you'd like
"Show feed of public chat room messages" is a neat chat rooms feature, but two improvements come to mind:

- Clickable room names to switch to the chat room of a given message (even if you are not already joined).
- Ability to exclude the Public chat room from triggering the blue notification dots on the Chat Rooms tab (perhaps any individual chat room should also be able to be muted in this way?). The benefit here would be that we'd only see new notification dots for chat rooms that we've actually joined.

### Additional context
Nicotine+ v3.2.9

slook:
The clickable username tags should also be removed in the Feed, since they are dysfunctional for users who are not watched.

PR #2514, opened 2023-05-28
chatrooms.py: global room feed restore on startup

slook:
+ Fixed: Open the "Public " global room feed at startup if it's in the autojoin list.

As mentioned in issue https://github.com/nicotine-plus/nicotine-plus/issues/2482#issuecomment-1516251512 it was remembered in the config but failed to open the Public chat tab at startup.


mathiascode:
There's a minor issue; the tab position of the Public room is not remembered between sessions.

slook:
No, it does always appear in the first position.

mathiascode:
> Since PR #2520 was merged, this PR no longer works at all.

Works on my end.

PR #2517, opened 2023-05-28
iconnotebook.py: tab highlight is_important fix priority

slook:
+ Fixed: logic error where `self.is_important` set itself back to False without being manually cleared by the user.

Chat room mentions could be overridden by a normal message, because remove_changed() was called too early, such that only the top-level tab retained the proper highlight, thereby after some time making it impossible to see which room the mention was in.

Regression possibly since a5d0e334bf although it was already even more confusing before that... the saga continues.

### DIFF
--- a/pynicotine/gtkgui/widgets/iconnotebook.py
+++ b/pynicotine/gtkgui/widgets/iconnotebook.py
@@ -169,15 +169,14 @@ class TabLabel:
 
     def request_changed(self, is_important=False):
 
-        self.remove_changed()
-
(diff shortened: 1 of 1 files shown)


mathiascode:
I went a step further in 632df38824bdc90ae1f1890cc691d029c5134ed6. Let me know if any issues are remaining.

slook:
It works a lot better. However, the top-level tab mention highlight does not dismiss when all the important tabs are viewed, the important mentioned highlight hangs around until after none remaining normal unread tabs are totally dismissed.

PR #2520, opened 2023-05-29
Always restore previously joined chatrooms on startup

mathiascode:
This is a simpler alternative to the current "Auto-join room" toggle.

https://github.com/nicotine-plus/nicotine-plus/discussions/2519#discussioncomment-6026386


slook:
It's a good idea.

The global Public room feed is not remembered, even though it does save into the conifg autojoin list.

hboetes:
I like the idea.

PR #2524, opened 2023-05-30
settings/chats.ui: completion dropdown spinner sensitive

slook:
+ Added: Grey out spinner widget if completion dropdown not enabled

image


pynicotine/gtkgui/ui/settings/chats.ui
```diff
@@ -164,6 +164,7 @@
                         <property name="hexpand">True</property>
                         <property name="label" translatable="yes">Minimum characters required to display drop-down:</property>
                         <property name="mnemonic-widget">min_chars_dropdown_spinner</property>
+                        <property name="sensitive" bind-source="enable_completion_dropdown_toggle" bind-property="active" bind-flags="bidirectional|sync-create"/>
```
mathiascode, line 167: Add this property to the parent box instead.

slook: It is another issue, there should be many multiple matches displayed. The problem is long-standing, I assume it must be something to do with X11 or suchlike, because it the vertical positioning is jumpy in a visually glitchy way.

PR #2527, opened 2023-05-30
iconnotebook.py: prepend_page for global room feed

slook:
+ Added: `prepend_page()` It makes sense to open the global room feed chat tab in the first position.
+ Added: `insert_page()` Sometimes it may be useful to open a new tab at a specific position, such as adjacent to current tab.
+ Fixed: Fail to activate new foreground tab if the Public room feed was enabled and is disabled then enabled again.

Since a custom position for the Public chat feed tab cannot be remembered across restarts, this avoids the risk of it getting buried in amongst other open chat tabs. If it is found to be annoying, then the opportunity close it is more easily presented this way, because it will be visible in the same position as it is on startup.

This approach might become more relevant if clickable room name tags as suggested in #2482 are implemented, therin it will be desirable to open a new tab adjacent to the current one.

### DIFF
--- a/pynicotine/gtkgui/chatrooms.py
+++ b/pynicotine/gtkgui/chatrooms.py
@@ -256,8 +256,13 @@ class ChatRooms(IconNotebook):
             return
 
         self.pages[msg.room] = tab = ChatRoom(self, msg.room, msg.users)
+        is_global = (msg.room == core.chatrooms.GLOBAL_ROOM_NAME)
 
-        self.append_page(tab.container, msg.room, focus_callback=tab.on_focus, close_callback=tab.on_leave_room)
+        if is_global:
+            self.prepend_page(tab.container, msg.room, focus_callback=tab.on_focus, close_callback=tab.on_leave_room)
+        else:
+            self.append_page(tab.container, msg.room, focus_callback=tab.on_focus, close_callback=tab.on_leave_room)
+
         tab.set_label(self.get_tab_label_inner(tab.container))
 
         if msg.room in self.autojoin_rooms:
@@ -266,7 +271,7 @@ class ChatRooms(IconNotebook):
             # Did not auto-join room, switch to tab
             core.chatrooms.show_room(msg.room)
 
-        if msg.room == core.chatrooms.GLOBAL_ROOM_NAME:
+        if is_global:
             self.roomlist.toggle_public_feed(True)
             return
 
@@ -367,7 +372,7 @@ class ChatRooms(IconNotebook):
             return
 
         for room in config.sections["server"]["autojoin"]:
-            if isinstance(room, str):
+            if isinstance(room, str) and room not in core.chatrooms.GLOBAL_ROOM_NAME:
                 self.autojoin_rooms.add(room)
 
     def server_disconnect(self, *_args):
--- a/pynicotine/gtkgui/chatrooms.py
+++ b/pynicotine/gtkgui/chatrooms.py
@@ -372,7 +372,7 @@ class ChatRooms(IconNotebook):
             return
 
         for room in config.sections["server"]["autojoin"]:
-            if isinstance(room, str) and room not in core.chatrooms.GLOBAL_ROOM_NAME:
+            if isinstance(room, str) and room != core.chatrooms.GLOBAL_ROOM_NAME:
                 self.autojoin_rooms.add(room)
 
     def server_disconnect(self, *_args):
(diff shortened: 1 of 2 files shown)


pynicotine/gtkgui/chatrooms.py
```diff
@@ -257,7 +257,13 @@ def join_room(self, msg):
 
         self.pages[msg.room] = tab = ChatRoom(self, msg.room, msg.users)
 
-        self.append_page(tab.container, msg.room, focus_callback=tab.on_focus, close_callback=tab.on_leave_room)
+        if msg.room == core.chatrooms.GLOBAL_ROOM_NAME:
+            is_global = True
```
mathiascode, line 261: Assign `is_global = (msg.room == core.chatrooms.GLOBAL_ROOM_NAME)` before the if clause, then check `if is_global`

pynicotine/gtkgui/chatrooms.py
```diff
@@ -266,7 +271,7 @@ def join_room(self, msg):
             # Did not auto-join room, switch to tab
             core.chatrooms.show_room(msg.room)
 
-        if msg.room == core.chatrooms.GLOBAL_ROOM_NAME:
+        if is_global:
             self.roomlist.toggle_public_feed(True)
```
slook, line 275: The activation behavior is inconsistent.

The first time it appears in the background (without changing focus). Thereafter when it is disabled/re-enabled for a second time then it grabs the tab focus.

pynicotine/gtkgui/chatrooms.py
```diff
@@ -266,7 +271,7 @@ def join_room(self, msg):
             # Did not auto-join room, switch to tab
             core.chatrooms.show_room(msg.room)
```
slook, line 272: The activation behavior is inconsistent.

The first time it appears in the background (without changing focus). Thereafter when it is disabled/re-enabled for a second time then it grabs the tab focus.

I suspect it depends upon if the room is in the autojoin_rooms list or not.

However, that problem is not related to this PR.

slook:
The global room feed Public tab is now pinned into position, it cannot be reordered.

pynicotine/gtkgui/chatrooms.py
```diff
@@ -256,17 +256,22 @@ def join_room(self, msg):
...
+        if is_global:
+            self.prepend_page(tab.container, msg.room, focus_callback=tab.on_focus, close_callback=tab.on_leave_room)
+        else:
+            self.append_page(tab.container, msg.room, focus_callback=tab.on_focus, close_callback=tab.on_leave_room)
 
-        self.append_page(tab.container, msg.room, focus_callback=tab.on_focus, close_callback=tab.on_leave_room)
         tab.set_label(self.get_tab_label_inner(tab.container))
 
-        if msg.room in self.autojoin_rooms:
+        if msg.room in self.autojoin_rooms and not is_global:
```
mathiascode, line 268: This doesn't seem right. What does this do?

slook: It is a workaround for the issue described above in https://github.com/nicotine-plus/nicotine-plus/pull/2527#discussion_r1210739489

mathiascode: There's a timing issue where the global room is joined before https://github.com/nicotine-plus/nicotine-plus/blob/2f79b8b7ac1bc41871b5d297893acfbb7776b0a5/pynicotine/gtkgui/chatrooms.py#L364-L371 runs. You can probably get around this by not adding the global room name to self.autojoin_rooms.

mathiascode: I really wish the server would send a confirmation message when joining the global room, would make things much easier.

slook: > Edit: annoying that GitHub doesn't display new review comments until you refresh the page...

No it's my fault yesterday I Resolved the comment before the review was properly solved, today I unresolved it so you'd see it.

pynicotine/gtkgui/chatrooms.py
```diff
@@ -367,7 +372,7 @@ def server_login(self, msg):
             return
 
         for room in config.sections["server"]["autojoin"]:
-            if isinstance(room, str):
+            if isinstance(room, str) and room not in core.chatrooms.GLOBAL_ROOM_NAME:
```
mathiascode, line 375:
```suggestion
            if isinstance(room, str) and room != core.chatrooms.GLOBAL_ROOM_NAME:
```

mathiascode:
54e7feb4efe3a5e07cba1ea8c24d8afd3c5ed438 should do it

slook:
It does work, nice one.

PR #2531, opened 2023-06-01
iconnotebook.py: don't highlight current tab on global feed

slook:
Regression since https://github.com/nicotine-plus/nicotine-plus/commit/20f28f862ff0b9a0cb224c9ef412ca98626a020d caused by #2516. Aims to resolve point 2 in issue #2482.

+ Fixed: wrongly highlighted the global room feed tab on new message when it's the current page and active parent window tab.

### DIFF
--- a/pynicotine/gtkgui/widgets/iconnotebook.py
+++ b/pynicotine/gtkgui/widgets/iconnotebook.py
@@ -468,20 +468,16 @@ class IconNotebook:
     def request_tab_changed(self, page, is_important=False, is_quiet=False):
 
         if self.parent_page is not None:
+            is_current_parent = (self.window.current_page_id == self.parent_page.id)
             is_current_page = (self.get_current_page() == page)
 
-            if is_quiet and not is_important:
-                # Don't highlight top-level tab on global feed message unless mentioned
-                pass
-
-            elif self.window.current_page_id != self.parent_page.id or not is_current_page:
-                # Highlight top-level tab
-                self.window.notebook.request_tab_changed(self.parent_page, is_important)
-
-            elif is_current_page:
+            if is_current_parent and is_current_page:
                 return
 
-            self.append_unread_page(page, is_important)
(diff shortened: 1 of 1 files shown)


mathiascode:
What are we at, like 10 attempts to get these darn tab highlights to work? :)

mathiascode:
This change reintroduces the issue https://github.com/nicotine-plus/nicotine-plus/commit/20f28f862ff0b9a0cb224c9ef412ca98626a020d tried to fix. Select a secondary tab that will be highlighted, switch to another main tab, wait for the highlight, switch back, and the highlight will remain on the main tab.

mathiascode:
This seems to work:

```py
    def request_tab_changed(self, page, is_important=False, is_quiet=False):

        if self.parent_page is not None:
            is_current_parent = (self.window.current_page_id == self.parent_page.id)
            is_current_page = (self.get_current_page() == page)

            if is_current_parent and is_current_page:
                return

            if not is_quiet or is_important:
                # Highlight top-level tab
                self.window.notebook.request_tab_changed(self.parent_page, is_important)
                self.append_unread_page(page, is_important)

        tab_label = self.get_tab_label(page)
        tab_label.request_changed(is_important)
```

slook:
> What are we at, like 10 attempts to get these darn tab highlights to work? :)

More than that, this has been years of work on it's own. There must be over a hundred commits that attempt address this problem. People must think we're idiots or something!

However, I fail to accept there won't be a way that covers all the basis, even if by that point we won't still be alive to witness it working properly. As I said at the outset of this undertaking in pr #2517, there is no doubt that this saga will continue...

Even though this turns out to be an infinitely iterative process, the recent iterations have actually made a big improvement over the way it was before.

Who would have thought that in a fully-fledged network client complete with a reverse-engineered protocol specification, that in fact it is these poxy notifier icons that are the most complex part of the program!

PR #2532, opened 2023-06-01
gtkgui/privatechat.py: implement buddy tab is_important

slook:
+ Added: Consider messages that are explicitly addressed to us by mentioned name, or are from buddies, to be important.

### DIFF
--- a/pynicotine/gtkgui/privatechat.py
+++ b/pynicotine/gtkgui/privatechat.py
@@ -385,9 +385,10 @@ class PrivateChat:
             callback=self.on_delete_chat_log_response
         ).show()
 
-    def show_notification(self, text):
+    def _show_notification(self, text, tag, is_buddy):
 
-        self.chats.request_tab_changed(self.container)
+        mentioned = (tag == self.chat_view.tag_highlight)
+        self.chats.request_tab_changed(self.container, is_important=is_buddy or mentioned)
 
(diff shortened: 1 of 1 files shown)


slook:
> I like to get popup notifications whenever anyone sends me anything.

That is my preference also. Perhaps this is more relevant for Chat Room messages, wherein it might be desirable to be alerted only about buddy chat activity, if that option is enabled, for that might warrant another toggle in preferences.

I have reverted the use of `notify_status`.

pynicotine/gtkgui/privatechat.py
```diff
@@ -398,20 +399,23 @@ def show_notification(self, text):
         self.chats.highlight_user(self.user)
 
         if config.sections["notifications"]["notification_popup_private_message"]:
+            user_type = _("Buddy") if is_buddy else _("User")
```
mathiascode, line 402: I would prefer two strings, `Private Message from %(user)s` and `Private Message from Buddy %(user)s`

slook: It doesn't seem worth making a new string just for this.

mathiascode: Separate strings would be much clearer for a translator. English is very forgiving when hacking together strings like this, but inserting standalone words into another sentence like this wouldn't work in e.g. one of the other languages I know due to declension.

PR #2536, opened 2023-06-01
iconnotebook.py: count unread tabs in tooltip text

slook:
+ Added: Count unread tabs in tooltip text

image

Ideally the parent top-level tab could acquire the same tooltip text, which would help to determine what the notification is about without the need to activate it.

### DIFF
--- a/pynicotine/gtkgui/widgets/iconnotebook.py
+++ b/pynicotine/gtkgui/widgets/iconnotebook.py
@@ -408,7 +408,7 @@ class IconNotebook:
 
         if self.unread_pages:
             icon_name = "emblem-important-symbolic"
-            tooltip_text = _("Unread Tabs")
+            tooltip_text = f'{len(self.unread_pages)} {_("Unread Tabs")}'
         else:
             icon_name = "pan-down-symbolic"
             tooltip_text = _("All Tabs")


pynicotine/gtkgui/widgets/iconnotebook.py
```diff
@@ -408,7 +408,7 @@ def update_pages_menu_button(self):
 
         if self.unread_pages:
             icon_name = "emblem-important-symbolic"
-            tooltip_text = _("Unread Tabs")
+            tooltip_text = f'{len(self.unread_pages)} {_("Unread Tabs")}'
```
mathiascode, line 411:
```suggestion
            tooltip_text = _("%(num)i Unread Tabs") % {"num": len(self.unread_pages)}'
```

slook: Are you sure it's worth creating a new translatable string for this?

mathiascode: Might also be worth looking into using format() for translatable strings in the future. You could simply do `_("{num} Unread Tabs").format(num=len(self.unread_pages))`, and the strings would look cleaner.

slook: > creating a new translatable string, it's safer to postpone this change

In that case, I will mark this PR as draft.

mathiascode: Since we're still doing final tweaks to strings, I decided to add a new string: 963370ddf3b22f16825eb5dcaf2df5e6c2d734ac

I'll of course give more time for translators to complete their translations.

PR #2540, opened 2023-06-02
ChatView: tag insert_new_line methods

slook:
Draft PR. Not intended for merging. The intention is to cherry-pick 

Some ideas for avoiding the need to deconstruct a chat line, instead add a rich-text chat line using raw data by taggging it straight into textbuffer, because it is faster and less prone to mistakes.

Please run this branch to evaluate the UX and benchmark performance.

Objectives:
- Optimize display of chat messages by avoiding the use of find()
- Read chat log file backwards in preparation for future implementation of on-demand readback
- Add room name tags as suggested in #2482 
- Experiment with a way to signify [buddy] in chat
- Tidy-up the timestamps

String manipulation is only required for log file readback because there is no enclosing demarkation character to end a logged `* ` (action) line (cannot know the username since we can't be certain of username containing spaces).

Fortunately, there is no need to do string dissemination when adding a chat line in real-time, because we have the individual data parts. As a result, sent or received `/me ` action lines can be processed raw retaining the username that we know to be correct.


slook:
Closing since divergence with master.

mathiascode:
Since chat logging was moved to core, I think we're in a better position to revisit this eventually.

Issue #2093, opened 2022-06-24
Granular control over shares with trusted buddies (+ custom buddy shares?)


redactedscribe:
### Describe the solution you'd like
Currently, shares can be individually limited to buddies only or not, with an additional option _Limit buddy-only shares to trusted buddies_ which is applied globally to all shares. A nice feature, however it's an all-or-nothing situation. I'd like to see this feature refined so that the augmenting trusted buddies option is no longer global, but instead settable on a per-share basis (the same way you can limit a particular share to be buddy-only or not).

The limit to trusted buddies option was original introduced as a way to let Nicotine+ user's have buddies on their buddy list without them also having access their private shares as this isn't always desired. With the suggested improvement, we could not only have private shares for the average buddy, but also a subset of those shares limited to buddies which are marked as trusted.

The idea here is that each type of share in the table assumes the same level of trust as the type above it:

Share type | Examples
-|-
Public share | A user's Downloads folder; lossy music; unsorted music
Buddy-only share | Lossless music; sorted music
Trusted buddies share | Hi-res lossless music; personally ripped music
Custom buddies share | Music library folder; misc. personal files

Taking it a step further (as seen in the last row of the table), perhaps a share could also be limited to a list of specific buddy usernames and/or IPs? E.g. one adds a buddy the share's custom list, then chooses whether the limitation is username-based, or IP-based, or both.

Should these ideas ever be implemented, the _Limit buddy-only shares to trusted buddies_ option could either be retired, or kept as a global switch, overriding any per-share trusted settings.

However for now, being able to set individual shares to public, buddy-only, or trusted buddies would be ideal (if this would be able to make it into the upcoming release, that would be perfect).

Leave a comment if further control of who has access to what shares is something you see any benefit in.

Thanks.

mathiascode:
> However for now, being able to set individual shares to public, buddy-only, or trusted buddies would be ideal (if this would be able to make it into the upcoming release, that would be perfect).

This, including the other share-related requests like private shares and excluding files/folders, will likely require a complete overhaul of the way shares are stored in Nicotine+. Shares are currently divided into two databases, public and buddy shares, which provide very little flexibility when it comes to individual folders.

Unfortunately, I have too many other things on my Nicotine+ TODO list right now, so I won't be able to work on this anytime soon.

redactedscribe:
[...]

PR #2557, opened 2023-06-28
Allow showing buddy-only shares to everyone

### LINKED ISSUE #1870: show files to all but locked
im confused i cant seem to work out how to make all my shares public but not downloadable unless the user is added to my Buddie List, people who are not on my list cannot view anything im sharing only my buddie list can

mathiascode:
Implements the plan described in https://github.com/nicotine-plus/nicotine-plus/issues/1870#issuecomment-1604290920

Closes #1870 

Seems to be working, but I wrote this yesterday while very tired, so variable names and code quality could perhaps be improved.

### DIFF
--- a/pynicotine/config.py
+++ b/pynicotine/config.py
@@ -210,6 +210,7 @@ class Config:
                 "queuelimit": 10000,
                 "filelimit": 100,
                 "buddysharestrustedonly": False,
+                "buddy_shares_visible_everyone": False,
                 "friendsnolimits": False,
                 "groupdownloads": "folder_grouping",
                 "groupuploads": "folder_grouping",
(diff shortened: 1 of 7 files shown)


slook:
I don't think it will be satisfactory enough for this to be a global setting. It is inevitably going to be required for this to be defined for each configured shared folder.

mathiascode:
I have no desire to work on groups and the performance and UI challenges it would bring in the near future. We either add this solution now, or wait another few years for something more granular like SoulseekQt has.

This PR addresses what was asked for in the first post in the issue report. For anything else, we have https://github.com/nicotine-plus/nicotine-plus/issues/2093.

slook:
It seems non-sensical to introduce an option that has planned obsolescence.

However, I can appreciate that it is difficult to see a more linear development path at this time.

mathiascode:
I received more requests about adding this as-is, so I've decided to finish this PR. The changes to the db files provide some nice reductions in memory and disk space usage anyway, due to no longer duplicating public shares in the buddy*.db files.

PR #2559, opened 2023-06-29
Modernize message dialogs

mathiascode:
- Use Adw.MessageDialog instead of deprecated Gtk.MessageDialog when libadwaita is enabled
- Use clearer labels for affirmative buttons according to GNOME HIG
- Add support for suggested/destructive buttons
- Make affirmative buttons activate by default according to GNOME HIG
- Stop using deprecated Gtk.ResponseType when possible
- Various API cleanups

### DIFF
--- a/pynicotine/gtkgui/__init__.py
+++ b/pynicotine/gtkgui/__init__.py
@@ -32,11 +32,6 @@ def check_gtk_version(gtk_api_version):
         gtk_api_version = "3"
         pygobject_version = (3, 26, 1)
 
-    if os.getenv("NICOTINE_LIBADWAITA") is None:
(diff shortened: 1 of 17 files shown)


slook:
Cannot cancel Shares Not Available dialog after attempting Retry.

mathiascode:
Are these new issues, or did they exist before?

mathiascode:
https://github.com/nicotine-plus/nicotine-plus/commit/4a13a081c5154121f35904164aaf99fed37b0d49 should fix the first issue you mentioned.

slook:
> - Cannot Unminimize main window after Minimizing while a dialog is open.
>> Are these new issues, or did they exist before?

Actually the problem with stuck minimized window existed before.

PR #2570, opened 2023-07-05
ChatEntry: try init_spell_checker only once

slook:
Code cleanup/optimization...

+ If initializing the Gspell package fails the first time when loading chats, then give up instead of trying for every open chat tab.
+ Retain the possibility for the Gspell package to be installed during the session (to make Preferences checkbox go visible).
+ Allow spell checking to be enabled/disabled without restarting the application (only switching the chat tab is needed).
+ Remove dependency on ChatEntry and centralize the spell checker object into `gtkgui.application`, because this will make is easier in future to enable spell checking on other kinds of widgets, if required.


pynicotine/gtkgui/dialogs/preferences.py
```diff
@@ -2947,6 +2942,9 @@ def update_settings(self, settings_closed=False):
         self.application.window.chatrooms.toggle_chat_buttons()
         self.application.window.privatechat.toggle_chat_buttons()
 
+        if not config.sections["ui"]["spellcheck"]:
+            self.application.spell_checker = None
```
mathiascode, line 2946: This doesn't seem like a good idea if we still keep a spell checker around in the existing ChatEntry widgets.

mathiascode: Especially not in combination with the call to init_spell_checker() above. Initializing a spell checker uses a fair bit of resources (especially memory).

slook: The spell checker that is initialized within `application` does not seem to be used anywhere, the changes in this PR would only re-initialize it (which I agree would be a good idea to avoid doing anyway).

> Initializing a spell checker uses a fair bit of resources

Would it perhaps be better to re-use the `application.spell_checker` object and re-assign the `spell_buffer` / `spell_view` in `on_switch_chat`?

slook: We only need to check spelling in one entry widget at a time, and with only a single buffer being active, since only 1 widget can have the input focus. Therefore multiple buffers are not needed.

slook:
Closing as superseeded by #2581, creating a class will make it easier as and when it's nessassry to migrate to a different spelling package.

PR #2571, opened 2023-07-05
ChatView: keyboard scrolling accessibility improvements

slook:
Accessibility improvements for chat view scrolling:

+ Added: Key handler to detect when scrolled to bottom (using the keyboard) and give the focus to the chat entry.
+ Added: Key handler for Down to give the focus to the chat entry, even if the scroll adjustment was already at the bottom.
+ Added: Key handler for Down to prevent the status bar or log pane from inadvertently stealing focus from the chat entry.
+ Added: Key handlers for Page_Up/Page_Down to allow chat view scrolling to be accessed from the chat entry.
+ Fixed: Public global room feed keyboard scrolling inoperative at startup if Chat Rooms is the default tab.

Note: No key handler for Up is required because the Entry widget naturally gives the focus away to the widget above.

Keyboard widget navigation behaviours are improved in this PR, especially when using the auto completion dropdown, which was very difficult to operate due to the focus being given away to above/below widgets and then getting trapped.


pynicotine/gtkgui/widgets/textview.py
```diff
@@ -331,8 +332,10 @@ def on_adjustment_upper_changed(self, *_args):
         self.adjustment_bottom = new_adjustment_bottom
 
     def on_adjustment_value_changed(self, *_args):
-
         new_value = self.adjustment.get_value()
+        self.update_adjustment_value(new_value)
+
+    def update_adjustment_value(self, new_value):
```
mathiascode, line 338: revert

pynicotine/gtkgui/widgets/textview.py
```diff
@@ -470,3 +477,18 @@ def update_user_tag(self, username):
...
+        if (self.widget.is_focus() and new_value >= self.adjustment_bottom
+                and not self.textbuffer.get_has_selection()):
+            # Give focus to text entry upon scrolling down to the bottom
+            self.chat_entry.grab_focus_without_selecting()
+
+        self.update_adjustment_value(new_value)
+
+    def on_page_down_accelerator(self, *_args):
+        """ Page_Down, Down: Give focus to text entry if already scrolled at the bottom """
+        self.on_adjustment_value_changed()
```
mathiascode, line 494:
```suggestion

    def on_page_down_accelerator(self, *_args):
        """ Page_Down, Down: Give focus to text entry if already scrolled at the bottom """

        if self.adjustment_value >= self.adjustment_bottom:
            # Give focus to text entry upon scrolling down to the bottom
            self.chat_entry.grab_focus_without_selecting()
```

pynicotine/gtkgui/widgets/textview.py
```diff
@@ -343,12 +346,16 @@ def on_adjustment_value_changed(self, *_args):
 
 class ChatView(TextView):
 
-    def __init__(self, *args, username_event=None, **kwargs):
+    def __init__(self, parent, chat, username_event=None, **kwargs):
```
mathiascode, line 349:
```suggestion
    def __init__(self, *args, chat_entry=None, username_event=None, **kwargs):
```

pynicotine/gtkgui/widgets/textentry.py
```diff
@@ -37,10 +37,11 @@
 class ChatEntry:
     """ Custom text entry with support for chat commands and completions """
 
-    def __init__(self, application, widget, completion, entity, send_message, is_chatroom=False):
+    def __init__(self, application, chat, completion, entity, send_message, is_chatroom=False):
 
         self.application = application
-        self.widget = widget
+        self.chat_view = chat.chat_view
+        self.widget = widget = chat.chat_entry
```
mathiascode, line 44: Please pass both as arguments explicitly.

pynicotine/gtkgui/widgets/textview.py
slook: What reason did you not like about focusing automatically upon mouse scrolling?

PR #2575, opened 2023-07-06
ChatEntry: remove unused message_class argument

slook:
Code cleanups:

- Removed: Unused `message_class` argument and variable from textentry.py.
- Removed: Don't import entire `slskmessages` module in 15 cases where only `UserStatus` is required.
- Removed: Don't import entire `slskmessages` module in 2 cases where only `FileListMessage` is required.

### DIFF
--- a/pynicotine/gtkgui/application.py
+++ b/pynicotine/gtkgui/application.py
@@ -26,11 +26,11 @@ from gi.repository import Gio
 from gi.repository import GLib
 from gi.repository import Gtk
 
-from pynicotine import slskmessages
 from pynicotine.config import config
 from pynicotine.core import core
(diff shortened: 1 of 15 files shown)


mathiascode:
> * Removed: Don't import entire `slskmessages` module in 15 cases where only `UserStatus` is required.
> * Removed: Don't import entire `slskmessages` module in 2 cases where only `FileListMessage` is required.

Your change is fine, but note that both methods still import the whole module due to how Python works.

slook:
It still seems like a good idea to restrict the exposure of core modules as much as possible. At the very least, it discourages developers from implementing networking functions within the gtkgui modules, that should instead be done in the core.

PR #2577, opened 2023-07-09
transferlist.py: don't get human speed unless value is changed

slook:
The call to get the human speed or time values can and should be avoided in most iterations of the row update, because it's not required if it's the same value.

### DIFF
--- a/pynicotine/gtkgui/transferlist.py
+++ b/pynicotine/gtkgui/transferlist.py
@@ -466,9 +466,7 @@ class TransferList:
 
         size = transfer.size or 0
         speed = transfer.speed or 0
-        hspeed = self.get_hspeed(speed)
         elapsed = transfer.time_elapsed or 0
-        helapsed = self.get_helapsed(elapsed)
         left = transfer.time_left or 0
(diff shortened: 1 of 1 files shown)


mathiascode:
I have a vague memory this was done because the exact values were too precise with frequent updates, but I think each transfer is only updated once per second nowadays?

slook:
> the exact values were too precise with frequent updates

In that case, then it would be better and more efficient to write some sort of math expression to smooth out the displayed speed, because that would enable slow string operations in human_speed to be avoided.

Perhaps this could be done by factoring down to the nearest ~1KiB/s before doing the comparison (in fact for all speed calculations, it seems pointless to have the accuracy of B/s, so that could be done in the core maybe?)

Ideally it would use an average or median value taken from the last few values, that would help improve readability if the values appear changing in the same direction, or at least remained still for a few seconds.

> I think each transfer is only updated once per second nowadays?

 Irrc the updates are only triggered after a certain chunk of `size`. What ever we did, that doesn't seem to be a concern anymore.

PR #2578, opened 2023-07-09
transfers.py: avoid unnecessary membership check while Transferring

slook:
Code cleanup/optimization is relevant if there is a very large number of queued users to check if `user` every time there is an update is waste of time in any case, because that data isn't needed for this condition.

### DIFF
--- a/pynicotine/transfers.py
+++ b/pynicotine/transfers.py
@@ -2158,13 +2158,13 @@ class Transfers:
 
         events.emit("update-upload", transfer, update_parent)
 
+        if status == "Transferring":
+            # Avoid unnecessary user counter updates while transferring
+            return
+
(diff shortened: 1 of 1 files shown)


mathiascode:
I don't see how this change makes any difference. If the status is different than Queued, the user check isn't reached anyway.

slook:
Well then it's one less status string match operation to do during `_file_upload_progress()`.

The circumstance is reached far more commonly than Queued transfers, because the function is persistently called for Transferring transfers.

mathiascode:
I suppose it's a question of what you want to prioritize, a potentially large number of files enqueued at once, or transferring files once per second. Unless you've actually profiled these changes and seen a difference, I'm inclined not to merge this PR.

slook:
Agreed, it makes no difference.

PR #2579, opened 2023-07-09
search.py: port to new TreeView class

mathiascode:
We can eventually look into creating a FileTreeView class or similar, but this will do until I've added support for Gtk.ColumnView.

Please keep the "Blocked" tag on treeview-related issues until we're using Gtk.ColumnView and know what's possible and what isn't.

### DIFF
--- a/pynicotine/gtkgui/search.py
+++ b/pynicotine/gtkgui/search.py
@@ -27,7 +27,6 @@ import re
 from collections import defaultdict
 
 from gi.repository import GObject
-from gi.repository import Gtk
 
 from pynicotine.config import config
 from pynicotine.core import core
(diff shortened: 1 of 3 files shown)


pynicotine/gtkgui/search.py
```diff
@@ -436,57 +428,101 @@ def __init__(self, searches, text, token, mode, mode_label, show_page):
...
+                "country_data": {"data_type": str},
+                "speed_data": {"data_type": GObject.TYPE_UINT},
+                "in_queue_data": {"data_type": GObject.TYPE_UINT},
+                "size_data": {"data_type": GObject.TYPE_UINT64},
+                "bitrate_data": {"data_type": GObject.TYPE_UINT},
+                "length_data": {"data_type": GObject.TYPE_UINT},
+                "file_path_data": {"data_type": str},
+                "free_slot_data": {"data_type": bool},
+                "file_attributes_data": {"data_type": GObject.TYPE_PYOBJECT},
+                "id_data": {
```
slook, line 518: I suppose the optional "ID" column no longer desired, hence why it is moved out of visible columns?

mathiascode: Yes, you can now click the column headers three times to revert to this sorting.

slook: It works flawless for most of the columns, apart from the Speed column, on the third press it still appears to be in ascending order for some reason.

slook: I must have been mistaken. It seems fine now.

mathiascode: There might be a subtle bug somewhere, I think I had a similar issue a while ago.

pynicotine/gtkgui/search.py
```diff
@@ -1315,25 +1316,15 @@ def update_result_counter(self):
 
         self.results_label.set_text(humanize(self.num_results_visible) + str_plus)
 
-    def on_column_position_changed(self, column, _param):
-        """ Save column position and width to config """
+    def on_file_path_tooltip(self, treeview, iterator):
```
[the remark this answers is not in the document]
mathiascode: Seems like this is not the only place where this change is needed. I'll do a treeview -> tree_view across the codebase after merging.

pynicotine/gtkgui/search.py
mathiascode: > It works flawless for most of the columns, apart from the "Speed" column, on the third press it still appears to be in ascending order for some reason.
> 
> Also the "In Queue" column is affected by the same problem - both ascending and unsorted are the same - so that would be all columns of type "number".

Possibly fixed in https://github.com/nicotine-plus/nicotine-plus/commit/707e187c5434fce0aa4761359cedf9904bdbdbf1

Issue #2555, opened 2023-06-27
Auto resume/retry


m3thm4th:
When downloading a lot of files and the queue is filled, the remaining files get the "Too many files" status; if the uploader terminates the program the status changes to "User logged off". When the queue is empty again, or when the uploader logs back in, the download doesn't start automatically, the user must press the "Resume" button.

It would be very useful to have the program automatically resume all the files in the download list every X minutes, so the downloads can proceed unattended. The user should be able to set the time interval between every retry. A button named "Auto-Resume" or "Auto-Retry" could be placed beside the "Resume" button, so the user would be able to activate and disactivate this behavior without going in the settings.

slook:
It should already be the behaviour, unless you are experiencing some bug. What version of Nicotine+ are you running?

cproensa:
I dont know about the specific situation of the remote user disconnecting, and not resuming when he's back online.

But, At first i also suspected something wrong about the auto retry behavior, becasue after finishing the current downloads, those files in "too many files" took a noticeable time to be restarted. However, they will eventually will be restarted.
So yes, the auto-resume works as intended (eventhough the interval is not congurable?)

What i also noticed:
* The auto resume is not windowed. I mean, ideally, when one file gets to completion, the first (or a number of) "locally queued" files should be probed. Currently, when the allowed downloads have finished, you still have to wait X time for the auto-resume to kick in.
* Manual "resume" over the list to force immediate retry, seems to try to probe all files. For a long list this has a big performance hit (ui is unresponsive for some time)
* A better approach may be a lazier probing. So that once we receive a "too many files" response, keep all succesive files locally queued without probing. This helps improving perfomance, and would work nicely with the previous suggestions for a windowed retry

slook:
> the auto-resume works as intended (eventhough the interval is not congurable?)

As far as I know, the intervals are based on the behaviour of the official client, and they are non-configurable for reasons of overall network stability and predictability.

> Manual "resume" over the list to force immediate retry, seems to try to probe all files. For a long list this has a big performance hit (ui is unresponsive for some time)

Any actions which cause the UI to hang are considered to be a bug, and should be investigated as such.

> A better approach may be a lazier probing.

That might be a strategy to consider which could help to avoid unnecessarily invoking a large iteration.

mathiascode:
> When the queue is empty again, or when the uploader logs back in, the download doesn't start automatically, the user must press the "Resume" button.

[...]

PR #2580, opened 2023-07-09
transfers.py: avoid redundant looping for users in manual retry uploads

slook:
Possible fix for one of the UI hang bugs mentioned by @cproensa in https://github.com/nicotine-plus/nicotine-plus/issues/2555#issuecomment-1610367072:

> Manual "resume" over the list to force immediate retry, seems to try to probe all files. For a long list this has a big performance hit (ui is unresponsive for some time)

Assuming that it is likely that a multi-selection will include mainly a lot of files with the same user, optimize to avoid looping through all the uploads for each individual transfer, instead just do it once for each user.

### DIFF
--- a/pynicotine/transfers.py
+++ b/pynicotine/transfers.py
@@ -1728,11 +1728,6 @@ class Transfers:
             transfer.status = "Queued"
             transfer.token = None
 
-        log.add_transfer("Initializing upload request for file %(file)s to user %(user)s", {
-            "file": filename,
-            "user": user
-        })
-
         core.watch_user(user)
 
         if slskmessages.UserStatus.OFFLINE in (core.user_status, core.user_statuses.get(user)):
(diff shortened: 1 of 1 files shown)


mathiascode:
> Possible fix for one of the UI hang bugs mentioned by @cproensa in [#2555 (comment)](https://github.com/nicotine-plus/nicotine-plus/issues/2555#issuecomment-1610367072):

What causes Nicotine+ to hang is the number of UploadDenied messages, which update both the download row and parent rows.

slook:
Possibly it is avoided by implementing the `locally_queued` flag in `push_file()`, which was not previously implemented for `retry_uploads()`.

Obviously there are several places where a similar one-pass mechanism is needed, but this PR only concerns Retry on a large multi-selection.

PR #2581, opened 2023-07-10
textentry.py: implement SpellChecker class

mathiascode:
Alternative to #2570


mathiascode:
> Having said that, it's much better than this happening for every open chat during startup, and it is advantageous that spelling starts working straight away as the package may be installed while the application is running.

There's probably a minor performance penalty for invoking import again, but it's much smaller than initially importing the module. I'm leaning towards keeping it, since it's nice to have things work instantly.

mathiascode:
Should be ready now.

PR #2584, opened 2023-07-15
trayicon.py: add Win32 implementation

mathiascode:
Restores tray icon and notification support on Windows.


mathiascode:
Ready on the Windows side, but my main computer with Linux is waiting for repairs, so I can't check why the tests fail there.

slook:
No errors are encountered at runtime on Linux here.

PR #2594, opened 2023-07-28
core_commands: share and unshare commands

mathiascode:
Replaces #2313

Should also contain most fixes from the previous PR, but it's possible I forgot to implement something. I tried to make the code as straightforward as possible, let me know if something is unclear or missing. @slook

### DIFF
--- a/pynicotine/gtkgui/dialogs/fastconfigure.py
+++ b/pynicotine/gtkgui/dialogs/fastconfigure.py
@@ -160,17 +160,18 @@ class FastConfigure(Dialog):
     def on_edit_shared_folder_response(self, dialog, _response_id, iterator):
 
         new_virtual_name = dialog.get_entry_value()
+        old_virtual_name = self.shares_list_view.get_row_value(iterator, "virtual_name")
 
-        if not new_virtual_name:
+        if new_virtual_name == old_virtual_name:
             return
 
         self.rescan_required = True
-
         folder_path = self.shares_list_view.get_row_value(iterator, "folder")
-        old_virtual_name = self.shares_list_view.get_row_value(iterator, "virtual_name")
 
         core.shares.remove_share(old_virtual_name)
-        core.shares.add_share(folder_path, virtual_name=new_virtual_name, validate_path=False)
+        new_virtual_name = core.shares.add_share(
+            folder_path, virtual_name=new_virtual_name, validate_path=False
+        )
 
         self.shares_list_view.set_row_value(iterator, "virtual_name", new_virtual_name)
 
--- a/pynicotine/gtkgui/dialogs/fastconfigure.py
+++ b/pynicotine/gtkgui/dialogs/fastconfigure.py
@@ -84,7 +84,7 @@ class FastConfigure(Dialog):
             activate_row_callback=self.on_edit_shared_folder,
             delete_accelerator_callback=self.on_remove_shared_folder,
             columns={
-                "virtual_folder": {
+                "virtual_name": {
                     "column_type": "text",
                     "title": _("Virtual Folder"),
                     "width": 1,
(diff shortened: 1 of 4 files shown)


pynicotine/shares.py
```diff
@@ -744,6 +725,69 @@ def get_compressed_shares_message(self, share_type):
 
         return self.compressed_shares.get(share_type)
 
+    def get_shared_folders(self):
+
+        shared_folders = config.sections["transfers"]["shared"]
+        shared_folders_buddy = config.sections["transfers"]["buddyshared"]
```
mathiascode, line 731: Need to verify if these have to be copied before passing them to the scanner process.

pynicotine/shares.py
```diff
@@ -744,6 +725,69 @@ def get_compressed_shares_message(self, share_type):
...
+
+    def add_share(self, folder_path, group_name="public", virtual_name=None, share_groups=None, validate_path=True):
+
+        if share_groups is None:
+            share_groups = self.get_shared_folders()
+
+        public_shares, buddy_shares = share_groups
+
+        if folder_path in (x[1] for x in public_shares + buddy_shares):
+            return None
```
slook, line 765: Suggest the wise thing to do here is to `remove_share()`. This would allow the share group to be changed (obviously after checking that the folder_path is accessible).

Without this suggested functionality, the "Cannot share inaccessible folder" error is incorrectly printed, yet the folder can be accessed but it is already shared.

Alternatively, `return False` to disambiguate the printed error message to something like "Already shared folder".

pynicotine/plugins/core_commands/__init__.py
```diff
@@ -537,6 +549,35 @@ def list_shares_command(self, args, **_unused):
...
+    def share_command(self, args, **_unused):
+
+        args_split = args.split(maxsplit=1)
+        group_name, folder_path = args_split[0], args_split[1].strip(' "')
+
+        self.core.shares.remove_share(folder_path)
+        virtual_name = self.core.shares.add_share(group_name, folder_path)
+
+        if not virtual_name:
+            self.output(_("Cannot share inaccessible folder \"%s\"") % folder_path)
```
slook, line 561: Error is incorrectly printed, yet the folder can be accessed but it is already shared.

slook (changes requested):

pynicotine/shares.py
slook: A copy is required unless you intend to make immediate changes to the shares that cannot be Cancelled by destoying the Preferences dialog.

pynicotine/shares.py
```diff
@@ -744,6 +725,69 @@ def get_compressed_shares_message(self, share_type):
...
+        if share_groups is None:
+            share_groups = self.get_shared_folders()
+
+        for shares in share_groups:
+            for virtual_name, folder_path in shares:
+                if virtual_name_or_folder_path in (virtual_name, folder_path):
+                    shares.remove((virtual_name, folder_path))
+                    return True
+
+        return False
```
slook, line 789: Consider returning the `folder_path` or `None` to print a more helpful CLI output.

slook:
>  I tried to make the code as straightforward as possible

The code in this PR is considerably nicer than my attempt, however my CLI works much better than yours.

pynicotine/gtkgui/dialogs/fastconfigure.py
```diff
@@ -172,34 +159,25 @@ def on_add_shared_folder(self, *_args):
 
     def on_edit_shared_folder_response(self, dialog, _response_id, iterator):
 
-        virtual_name = dialog.get_entry_value()
+        new_virtual_name = dialog.get_entry_value()
 
-        if not virtual_name:
+        if not new_virtual_name:
             return
```
slook, line 165: A noop is undesirable here.

Consider returning the virtual_name to the default instead (i.e. based on the folder name) in this case whereupon the text entry is cleared to `""` then pass `None`.

mathiascode:
Sorry, forgot to move an argument after refactoring. Can you try the latest changes?

pynicotine/shares.py
mathiascode: > After removal using the `/unshare` command, it doesn't seem possible to `/share` it again.

This happened due to group_name and folder_path arguments being incorrectly swapped in the share command.

slook:
Much better it works quite well now.

pynicotine/gtkgui/dialogs/fastconfigure.py
mathiascode: Is it better now?

slook: That is the correct dialog behaviour yes.

slook (approved):

pynicotine/shares.py
mathiascode: We definitely want copies, but I'm curious if multiprocessing does some magic here already. I'll verify later.

slook: I don't understand why we still need to pass `shared_folders` around between functions within the shares module.

mathiascode: > I suppose it is need while the dialog list box is in the process of being edited.

Yes.

The lists are indeed copied when passed to the scanner process already.

slook: The other thing I don't understand is why the database is not enough on it's own for storing the list of shared folders and virtual names. What is the benefit of having config settings for this?

mathiascode: None of the current .db files are designed for storing a list of top-level shares, and the files are not human-readable/editable anyway. They are also more "temporary" than say the downloads/uploads.json files, in the sense that they're essentially caching folder/file data in a way that's convenient for Nicotine+'s needs, and can be rebuilt if they're lost.

PR #2602, opened 2023-08-01
Support reopening closed tabs

### LINKED ISSUE #1424: Reopen closed tabs
Would be nice to have an reopen feature, which brings back a recently closed search tab. On most current browsers this is achieved with pressing `CRTL + SHIFT + T`.

When pressed again, it could bring back tabs in chronical order last in first out.

mathiascode:
Adds Ctrl+Shift+T shortcut for reopening recently closed tabs.

Closes #1424 

I don't use this feature in other applications, so I'm not entirely sure if the behavior is correct, but feedback is welcome.

### DIFF
--- a/pynicotine/gtkgui/widgets/iconnotebook.py
+++ b/pynicotine/gtkgui/widgets/iconnotebook.py
@@ -22,6 +22,8 @@
 
 import sys
 
+from collections import deque
+
 from gi.repository import Gdk
 from gi.repository import GLib
 from gi.repository import Gtk
@@ -242,6 +244,7 @@ class IconNotebook:
         self.pages = {}
         self.tab_labels = {}
         self.unread_pages = {}
+        self.recently_removed_pages = deque(maxlen=5)  # Low limit to prevent excessive server traffic
 
         self.widget = Gtk.Notebook(scrollable=True, show_border=False, visible=True)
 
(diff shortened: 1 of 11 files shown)


pynicotine/gtkgui/widgets/iconnotebook.py
```diff
@@ -634,10 +655,12 @@ def on_pages_button_pressed(self, *args):
 
         self.popup_menu_pages.add_items(
             ("", None),
+            ("#" + _("_Reopen Closed Tab"), self.restore_removed_page),
```
slook, line 658: `Reopen Closed _Tab` or `Re_open Closed Tab`

pynicotine/search.py
```diff
@@ -132,9 +132,7 @@ def remove_all_searches(self):
     def show_search(self, token):
         events.emit("show-search", token)
 
-    def process_search_term(self, search_term, mode, room=None, user=None):
-
-        users = []
+    def process_search_term(self, search_term, mode, room=None, users=None):
```
slook, line 135: What is the reason for `user` -> `users`?

mathiascode: Was easier to deal with the internal support for multiple users this way, but on second thought, I don't know why we still support this when it's never been used. It should probably be removed.

mathiascode: Saving this for a separate PR in any case.

PR #2625, opened 2023-08-25
reword the "invalid password" dialogue

hboetes:


pynicotine/gtkgui/application.py
```diff
@@ -344,8 +344,8 @@ def on_invalid_password(self):
         from pynicotine.gtkgui.widgets.dialogs import OptionDialog
 
         title = _("Invalid Password")
-        msg = _("User %s already exists, and the password you entered is invalid. Please choose another username "
-                "if this is your first time logging in.") % config.sections["server"]["login"]
+        msg = _("Either user %s already exists, or the password you entered is invalid."
+                " If this is the first time you log in, please choose a different username.") % config.sections["server"]["login"]
```
slook, line 348: "...time you log in" doesn't make sense.

slook (changes requested):

pynicotine/gtkgui/application.py
hboetes: How would you put it?

mathiascode: What was wrong with `your first time logging in`?

pynicotine/gtkgui/application.py
```diff
@@ -344,8 +344,8 @@ def on_invalid_password(self):
         from pynicotine.gtkgui.widgets.dialogs import OptionDialog
 
         title = _("Invalid Password")
-        msg = _("User %s already exists, and the password you entered is invalid. Please choose another username "
-                "if this is your first time logging in.") % config.sections["server"]["login"]
+        msg = _("Either user %s already exists, or the password you entered is invalid."
```
mathiascode, line 347: A password can only be invalid if the user exists.

pynicotine/gtkgui/application.py
hboetes: Strictly speaking there isn't, but the proposed construction is how a native speaker would formulate it.

slook: An alternative way of putting it is `If this is the first time you have logged in`.

slook:
I don't think we should bother updating the invalid password string at the moment. The entire mechanism requires deeper review.

pynicotine/gtkgui/application.py
slook: "right" -> "correct"

PR #2630, opened 2023-08-30
fastconfigure.ui: add notice about encryption

mathiascode:


README.md
```diff
@@ -51,6 +51,8 @@ I stumbled across [this quote](http://www.bash.org/?10159):
 
 Nicotine+ does not collect any user data. Different policies may apply to the default Soulseek server, which is not operated by the Nicotine+ Team.
 
+Soulseek is unencrypted, and should not be used for secure communication. Use a VPN if you are concerned about your IP address being public.
```
slook, line 54: "should not be used for" -> "is not intended for"

It is not warranted to raise any concerns about the IP address, nor is it our remit to suggest using VPN or other such.

slook (changes requested):

README.md
hboetes: Soulseek is an unencrypted protocol,...

mathiascode: > "should not be used for" -> "is not intended for"
> 
> It is not warranted to raise any concerns about the IP address, nor is it our remit to suggest using VPN or other such.

Right, I suppose you can already draw that conclusion from the statement about encryption.

mathiascode:
Updated

PR #2631, opened 2023-09-05
search.py: add context menu item to view user profiles

mathiascode:
User profiles are an integral part of the Soulseek network. It's where you learn more about a user's collection, requirements and how to get access to private files.

Currently, the menu item for viewing a user's profile is buried in a submenu, making it awkward to access frequently. Add a primary menu item to make user profiles more obvious and easier to access, similar to SoulseekQt.

### DIFF
--- a/pynicotine/gtkgui/search.py
+++ b/pynicotine/gtkgui/search.py
@@ -566,7 +566,8 @@ class Search:
             ("#" + _("Download _Folder(s)"), self.on_download_folders),
             ("#" + _("Download F_older(s) To…"), self.on_download_folders_to),
             ("", None),
-            ("#" + _("_Browse Folder(s)"), self.on_browse_folder),
+            ("#" + _("View User _Profile"), self.on_user_profile),
+            ("#" + _("_Browse Folder"), self.on_browse_folder),
             ("#" + _("F_ile Properties"), self.on_file_properties),
             ("", None),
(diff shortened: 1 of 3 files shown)


pynicotine/gtkgui/search.py
```diff
@@ -566,6 +566,7 @@ def __init__(self, searches, text, token, mode, mode_label, room, users, show_pa
             ("#" + _("Download _Folder(s)"), self.on_download_folders),
             ("#" + _("Download F_older(s) To…"), self.on_download_folders_to),
             ("", None),
+            ("#" + _("_View User Profile(s)"), self.on_user_profile),
```
slook, line 569: `"_View User Profile(s)"` -> `"_View User Profile"`

Suggest hide or disable the menu item if multiple users are selected, because it is not useful to open multiple User Profile tabs.

The menu item visibility/sensitivity could be set on opening of the context menu, at the same time as the User Actions menu(s) are built..

slook (changes requested):

pynicotine/gtkgui/search.py
mathiascode: > Suggest hide or disable the menu item if multiple users are selected

Not necessary in my opinion, we can just open the user profile of the first selected user.

slook: Fair enough, that is a lot more straight-forward than my suggestion.

PR #2634, opened 2023-09-06
Move buddy list position option to Preferences dialog

mathiascode:
This is easier to maintain than radio menu items, and seems like an option that wouldn't be used frequently. We already include other position-related options in the Preferences dialog, might as well be consistent here.

### DIFF
--- a/pynicotine/gtkgui/dialogs/preferences.py
+++ b/pynicotine/gtkgui/dialogs/preferences.py
@@ -1487,6 +1487,7 @@ class UserInterfacePage:
     def __init__(self, application):
 
         (
+            self.buddy_list_position_label,
             self.chat_colored_usernames_toggle,
             self.chat_username_appearance_label,
             self.close_action_label,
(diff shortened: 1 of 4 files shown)


pynicotine/gtkgui/dialogs/preferences.py
```diff
@@ -1607,6 +1608,16 @@ def __init__(self, application):
             )
         )
 
+        self.buddy_list_position_combobox = ComboBox(
+            container=self.buddy_list_position_label.get_parent(), label=self.buddy_list_position_label,
+            item_selected_callback=self.on_select_buddy_list_position,
+            items=(
+                (_("Separate tab"), "tab"),
+                (_("Chat rooms tab"), "chatrooms"),
```
slook, line 1616: Capitalization `"Chat rooms tab"` -> `"Chat Rooms tab"`

More descriptive choices might be desirable...

"Separate Buddies tab"
"Sidebar in Chat Rooms tab"
"Sidebar always visible in all tabs"

mathiascode: Would `Always visible sidebar` work for the last choice?

slook: Yes, but for the want of something better.

PR #2637, opened 2023-09-08
Move version checker to About dialog

mathiascode:
Up to date
New version available
Error checking version


mathiascode:
@hboetes @slook I'm a bit conflicted on the `New version available` string. What sounds better?
- New version available
- New release available
- Something else?

hboetes:
Both are fine, I slightly prefer the second one.

PR #2641, opened 2023-09-09
Update DOWNLOADS.md - add apt-key command

mbledkowski:
Hello,
I have added `apt-key` command to DOWNLOADS.md documentation. Adding key for ppa repository is necessary, in order for it to work.
Best regards,
Maciej Błędkowski


mathiascode:
We shouldn't do this anymore, apt-key is deprecated.

add-apt-repository also added the key when I tested the command on Debian. Is this not the case on your system?

mathiascode:
Thank you for your PR, but I'll close it for the reason above. Let me know if there's a Debian-based system where the commands don't work.

PR #2642, opened 2023-09-09
Remove View menu

mathiascode:
All options in this menu have either been moved to Preferences -> User Interface, or can be toggled by other means. Nicotine+ also respects the system dark mode preference on Windows and macOS now, so a quick-access dark mode menu item shouldn't be necessary in most cases.


mathiascode:
Merging for further feedback.

PR #2643, opened 2023-09-09
settings/uploads.ui: improve a few strings

mathiascode:
Replaces #1936

Upload prefs

### DIFF
--- a/pynicotine/gtkgui/dialogs/preferences.py
+++ b/pynicotine/gtkgui/dialogs/preferences.py
@@ -736,8 +736,8 @@ class UploadsPage:
             self.use_alt_speed_limit_radio,
             self.use_speed_limit_radio,
             self.use_unlimited_speed_radio,
-            self.use_upload_bandwidth_radio,
-            self.use_upload_slots_radio
+            self.use_upload_slots_bandwidth_radio,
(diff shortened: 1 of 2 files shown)


mathiascode:
@hboetes Thoughts?

hboetes:
I'd remove the word 'use' so:

- upload speed limit
- alternative upload speed limit

And I think putting the buddy checkboxes as the last item of the paragraph looks better, since they are the exception:

- Upload queue type
- Allocate upload slots until total speed reaches (KiB/s):
- Fixed number of upload slots:
- Prioritize Buddies

- Each user may queue a maximum of either:
- Limits do not apply to buddies.

And I left a few words out on purpose.

mathiascode:
> * Prioritize Buddies

Needs to mention that the option prioritizes every buddy, i.e. overrides the preference in the buddy list.

hboetes:
> > * Prioritize Buddies
> 
> Needs to mention that the option prioritizes every buddy, i.e. overrides the preference in the buddy list.

OK, then it should be "prioritize all buddies" indeed.

mathiascode:
Made some changes:
Upload preferences

hboetes:
Looking good!  😎

PR #2651, opened 2023-09-17
Leech Detector: remember previous leechers

### LINKED ISSUE #2105: Save leechers persistently
Save a list of leechers messaged by the Leech Detector plugin to avoid messaging them more than once.

I would actually use this list to check users at regular intervals, like once a day, or at startup, to check the number of files and remove the ban from them, since I've set the plugin to ban instead of messaging.

mathiascode:
Closes #2105 

I opted for a list string option, since it exposes a convenient list of leechers that can be modified in the GUI.

### DIFF
--- a/pynicotine/plugins/leech_detector/__init__.py
+++ b/pynicotine/plugins/leech_detector/__init__.py
@@ -76,6 +76,7 @@ class Plugin(BasePlugin):
         if self.settings["num_folders"] < min_num_folders:
             self.settings["num_folders"] = min_num_folders
 
+        # Separate messaged_users set for faster membership checks
         self.messaged_users = set(self.settings["messaged_users"])
 
         self.log(
@@ -84,10 +85,6 @@ class Plugin(BasePlugin):
         )
 
     def upload_queued_notification(self, user, virtual_path, real_path):
-
-        if user in self.messaged_users:
-            # Already messaged user previously
-            return
 
         if user in self.probed_users:
             # We already have stats for this user.
@@ -110,10 +107,20 @@ class Plugin(BasePlugin):
 
         num_files = stats["files"]
         num_folders = stats["dirs"]
+        was_messaged = (user in self.messaged_users)
 
         if num_files >= self.settings["num_files"] and num_folders >= self.settings["num_folders"]:
+            if was_messaged:
+                self.messaged_users.remove(user)
+                self.settings["messaged_users"].remove(user)
+
             self.probed_users[user] = "okay"
             self.log("User %s is okay, sharing %s files in %s folders.", (user, num_files, num_folders))
+            return
+
+        if was_messaged:
+            # Still leeching, but we already messaged user previously
+            self.probed_users[user] = "processed"
             return
 
         if user in self.core.userlist.buddies:
@@ -165,7 +172,7 @@ class Plugin(BasePlugin):
 
             self.send_private(user, line, show_ui=self.settings["open_private_chat"], switch_page=False)
 
-        self.settings["messaged_users"].append(user)
         self.messaged_users.add(user)
+        self.settings["messaged_users"].append(user)
 
         self.log("Leecher %s doesn't share enough files. Message sent.", user)
--- a/pynicotine/plugins/leech_detector/__init__.py
+++ b/pynicotine/plugins/leech_detector/__init__.py
@@ -101,26 +101,21 @@ class Plugin(BasePlugin):
             # We did not trigger this notification
             return
 
+        num_files = stats["files"]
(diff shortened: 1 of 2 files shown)


pynicotine/plugins/leech_detector/__init__.py
```diff
@@ -71,72 +76,83 @@ def loaded_notification(self):
...
-            (self.str_action, self.settings["num_files"], self.settings["num_folders"])
+            "Require users have a minimum of %d files in %d shared public folders.",
+            (self.settings["num_files"], self.settings["num_folders"])
         )
 
     def upload_queued_notification(self, user, virtual_path, real_path):
 
-        if user in self.probed:
+        if user in self.messaged_users:
+            # Already messaged user previously
```
slook, line 89: What about if the user has since gained more files, then should we remove them from the list?

pynicotine/plugins/leech_detector/__init__.py
```diff
@@ -54,11 +55,15 @@ def __init__(self, *args, **kwargs):
...
+            "messaged_users": {
+                "description": "Previously messaged users",
+                "type": "list string"
             }
         }
 
-        self.probed = {}
-        self.str_action = ""
+        self.probed_users = {}
+        self.messaged_users = set()
```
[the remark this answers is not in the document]
mathiascode: I'm concerned about O(n) membership checks if the list grows large, the set ensures O(1).

pynicotine/plugins/leech_detector/__init__.py
mathiascode: I suppose we could do that.

pynicotine/plugins/leech_detector/__init__.py
```diff
@@ -71,72 +76,90 @@ def loaded_notification(self):
...
 
-        if stats["files"] >= self.settings["num_files"] and stats["dirs"] >= self.settings["num_folders"]:
-            self.log("User %s is okay, sharing %s files in %s folders.", (user, stats["files"], stats["dirs"]))
-            self.probed[user] = "okay"
+        num_files = stats["files"]
+        num_folders = stats["dirs"]
+        is_previous_leecher = (user in self.detected_leechers)
+
+        if num_files >= self.settings["num_files"] and num_folders >= self.settings["num_folders"]:
+            if is_previous_leecher:
```
slook, line 113: This code is unreachable, it is not above the condition on Line 104.

mathiascode: It's reachable in later sessions. Not sure if it's worth making it reachable in the same session, it would require adding some checks to ensure we don't spam the server with user stats requests.

pynicotine/plugins/leech_detector/__init__.py
```diff
@@ -71,72 +76,85 @@ def loaded_notification(self):
...
             return
 
-        if self.probed[user] != "requesting":
+        num_files = stats["files"]
+        num_folders = stats["dirs"]
+        is_user_accepted = (num_files >= self.settings["num_files"] and num_folders >= self.settings["num_folders"])
+
+        if is_user_accepted and user in self.detected_leechers:
+            self.detected_leechers.remove(user)
+            self.settings["detected_leechers"].remove(user)
```
slook, line 110: `remove` -> `discard`

pynicotine/plugins/leech_detector/__init__.py
slook: It is still unreachable, due to the condition on Line 89, consider in `upload_queued_notification()`:

```
    if user in self.probed_users and user not in self.detected_leechers:
        # We already have stats and accepted this user
```

pynicotine/plugins/leech_detector/__init__.py
mathiascode: Not while Nicotine+ is running, but I'll change to discard regardless.

mathiascode: Maybe I need more sleep, list doesn't have discard()...

pynicotine/plugins/leech_detector/__init__.py
mathiascode: I'll revisit later, this fucking plugin is getting on my nerves.

mathiascode: I'm not going to spam the server with user stat requests on every queued file request, we need something better.

mathiascode: Alright, pushed a change that should make my idea work.

mathiascode:
Hopefully I got it right this time...

PR #2654, opened 2023-09-18
userbrowse.py: implement path bar

mathiascode:
There are some issues left to sort out related to horizontal scrolling, but otherwise the path bar works fine.

The new design also allows for adding a filter bar later, similar to searches.

share

### DIFF
--- a/pynicotine/gtkgui/userbrowse.py
+++ b/pynicotine/gtkgui/userbrowse.py
@@ -28,6 +28,7 @@ from locale import strxfrm
 from gi.repository import GLib
 from gi.repository import GObject
 from gi.repository import Gtk
+from gi.repository import Pango
 
 from pynicotine.config import config
 from pynicotine.core import core
@@ -45,7 +46,9 @@ from pynicotine.gtkgui.widgets.popupmenu import PopupMenu
 from pynicotine.gtkgui.widgets.popupmenu import FilePopupMenu
 from pynicotine.gtkgui.widgets.popupmenu import UserPopupMenu
 from pynicotine.gtkgui.widgets.textentry import ComboBox
+from pynicotine.gtkgui.widgets.theme import add_css_class
 from pynicotine.gtkgui.widgets.theme import get_file_type_icon_name
+from pynicotine.gtkgui.widgets.theme import remove_css_class
 from pynicotine.gtkgui.widgets.treeview import TreeView
 from pynicotine.slskmessages import UserStatus
 from pynicotine.slskmessages import FileListMessage
@@ -194,11 +197,14 @@ class UserBrowse:
             self.folder_tree_container,
             self.info_bar_container,
             self.num_folders_label,
+            self.path_bar,
             self.progress_bar,
             self.refresh_button,
             self.retry_button,
             self.save_button,
+            self.search_button,
             self.search_entry,
+            self.search_entry_revealer,
             self.share_size_label
         ) = ui.load(scope=self, path="userbrowse.ui")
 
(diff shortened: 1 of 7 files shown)


pynicotine/gtkgui/userbrowse.py
```diff
@@ -447,8 +454,7 @@ def make_new_model(self, shares, private_shares=None):
         self.share_size = size + private_size
         self.num_folders = num_folders + num_private_folders
 
-        self.share_size_label.set_text(human_size(self.share_size))
-        self.num_folders_label.set_text(humanize(self.num_folders))
+        self.num_folders_label.set_text(f"{humanize(self.num_folders)}   /   {human_size(self.share_size)}")
```
slook, line 457: `f"{humanize(self.num_folders)} ({human_size(self.share_size)})"`

pynicotine/gtkgui/userbrowse.py
```diff
@@ -603,11 +609,42 @@ def set_finished(self):
...
+    def populate_path_bar(self, folder_path=""):
+
+        for widget in list(self.path_bar):
+            self.path_bar.remove(widget)
+
+        for index, folder in enumerate(folder_path.split("\\")):
+            i_folder_path = "\\".join(folder_path.split("\\")[:index + 1])
+
+            if index:
+                label = Gtk.Label(label="/", visible=True)
```
slook, line 621: `"/"` -> `"\"`
Soulseek paths use backslashes.

Also, the horizontal padding seems a bit too spacious.

slook:
The positioning of the "Expand / Collapse All" button looks like it should be a button which goes up a folder level.

Maybe consider combining it with the "Folder" label?

pynicotine/gtkgui/userbrowse.py
mathiascode: I went with a backslash first, but @hboetes preferred a forward slash.

I can try reducing the padding slightly more, but I don't want to deviate from Nautilus' current design too much.

slook: In all other parts of the application, a backslash `\` is used in remote file paths, and on *nix a forward slash is used for local paths.

A forward slash here is technically incorrect on all systems, especially on Windows, where a forward slash is never seen in any path string.

slook:
It's nice, but not really intuitive at first glance. It still looks like either a Download button or a Up a folder level button.

I'm beginning to wonder if the Expand/Collapse All button actually has any real value, or if it would be better to bury it in the Folder context menu or something like where the right click within empty space in the folder tree view.

mathiascode:
Made some more changes. I think the only remaining issue is that the path bar doesn't always scroll all the way to the right in GTK 4.

slook:
A horizontal border line is required at the bottom of the toolbar, or at the top of the tree views.

mathiascode:
> A horizontal border line is required at the bottom of the toolbar, or at the top of the tree views.

I tried, but it felt a bit heavy with all the borders already present. I wonder if making the progress bar take up the full width of the tab would help correct the visual imbalance between the ends of the pane border.

slook:
> making the progress bar take up the full width of the tab

The progress bar could ideally be moved into the toolbar to replace the path bar during loading operations, and then be invisible once loading is complete, because the path bar is not required during loading, and the progress bar widget doesn't serve any purpose after loading, other than just taking up space.

mathiascode:
Alright, made some more changes.

slook:
The progress bar doesn't appear at all on 3.3.0.dev5   •   Python 3.9.2   •   GTK 3.24.24

mathiascode:
> The progress bar doesn't appear at all on GTK 3.24.24

How about https://github.com/nicotine-plus/nicotine-plus/pull/2661?

slook:
Yes it works. Beautiful.

PR #2660, opened 2023-09-21
Fix minor warning whilst testing the albanian po file

hboetes:
nicotine@bran ~/src/nicotine-plus % for i in po/*.po; msgfmt -c -o /dev/null $i po/sq.po:7: warning: header field 'Project-Id-Version' still has the initial default value

Perhaps this can also be added to the tests?

### DIFF
--- a/po/sq.po
+++ b/po/sq.po
@@ -5,7 +5,7 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: PACKAGE VERSION\n"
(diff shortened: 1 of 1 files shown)


mathiascode:
https://github.com/nicotine-plus/nicotine-plus/blob/master/doc/DEVELOPING.md#adding-a-language

Adding it to the tests would probably help encourage me to do this a bit faster.

mathiascode:
Added a test.

PR #2653, opened 2023-09-18
shares.py: implement custom database format for shares

mathiascode:
gdbm is a hassle to install on some systems, and semidbm has been unmaintained for several years. Implement a custom, straightforward key-value database format that's optimized for Nicotine+ shares.

Should be merged at the same time as https://github.com/nicotine-plus/nicotine-plus/pull/2557

PR #2675, opened 2023-10-07
Shares: new database format and trusted shares

### LINKED ISSUE #1870: show files to all but locked

### LINKED ISSUE #2093: Granular control over shares with trusted buddies (+ custom buddy shares?)

### LINKED ISSUE #2447: Uploads erroneously getting listed as complete, actually "aborted" at 99%
Nicotine+ version:  3.2.8/3.2.9
Operating System/Distribution: Windows 10 (22H2)

### Describe the bug

I'm getting reports from numerous people that their downloads from me get aborted at 99% completion, which I surely did not initiate myself. Worse, on my end, the uploads are listed as 100% complete. Since updating to 3.2.8 (_maybe_ 3.2.7 but I believe not) I _have_ been seeing strange activity in my uploads tab on about two-thirds of all my uploads, which I think could be related in some way; I don't know if it is, as I've never happened to see this behaviour and then gotten a message about the same upload, but I'll explain in the case that it's related.

For about two-thirds of the uploads that get queued, on my end I'll instantly see the first queued upload get set to 100%, yet it'll still take a while to move on (doing the actual upload, one would assume). Occasionally, these same uploads get set back to 0% and get requeued.

I will add that, though I've seen this behaviour as of updating to version 3.2.8, I've only been receiving messages for the failing uploads as of yesterday, a week after updating to version 3.2.9. On my end, nothing appears to have changed.

### Expected behavior

Uploads that are listed as 100% complete to be listed as 100% complete rather than 99% aborted on the downloader's end.

Haven't the foggiest, sorry. Could try downloading from me but I'm not online 24/7, and it would only provide the downloader perspective.

### Additional context

N/A

mathiascode:
gdbm is a hassle to install on some systems, and semidbm has been unmaintained for several years. Implement a custom, straightforward key-value database format that's optimized for Nicotine+ shares.

Also add support for trusted shares, as well as showing buddy and trusted shares to everyone.

Alternative to #2653 that includes #2557 

Fixes #2447 by using file modification times instead of folder
Closes #1870 
Closes #2093


pynicotine/shares.py
```diff
@@ -578,24 +643,26 @@ def __init__(self):
...
-        self.should_compress_shares = False
-        self.compressed_shares = {}
-
-        for share_type in ("public", "buddy"):
-            self.compressed_shares[share_type] = slskmessages.SharedFileListResponse()
+        self.compressed_shares = {
+            "public": slskmessages.SharedFileListResponse(),
+            "buddy": slskmessages.SharedFileListResponse(),
+            "trusted": slskmessages.SharedFileListResponse(),
+            "banned": slskmessages.SharedFileListResponse()
```
slook, line 650: Unclear of "banned" purpose in here?

mathiascode: If a banned user requests your shares, they get an empty list (this used to be in the _shared_file_list_request method before, but I needed it for local shares as well).

PR #2677, opened 2023-10-09
gtkgui/transfers.py: select parent rows of new transfers

mathiascode:
Makes it easier to find transfers that were recently added.

### DIFF
--- a/pynicotine/gtkgui/transfers.py
+++ b/pynicotine/gtkgui/transfers.py
@@ -256,7 +256,7 @@ class Transfers:
         self.transfer_list = transfer_list
 
         self.container.get_parent().set_visible(bool(transfer_list))
-        self.update_model()
+        self.update_model(select_parent=False)
 
     def select_transfers(self):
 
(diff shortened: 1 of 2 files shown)


pynicotine/gtkgui/transfers.py
```diff
@@ -625,11 +626,25 @@ def update_specific(self, transfer):
...
             else:
                 parent_iterator = user_iterator
                 user_child_transfers.append(transfer)
+
+                if select_parent:
+                    select_iterator = user_iterator
+
+            if select_iterator and (not self.tree_view.is_row_selected(select_iterator)
+                                    or self.tree_view.get_num_selected_rows() != 1):
+                # Select parent row of newly added transfer
```
slook, line 644: Please clarify or explain the condition.

slook (changes requested):

pynicotine/gtkgui/transfers.py
mathiascode: Done

PR #2679, opened 2023-10-10
Use locale's timestamp format by default

mathiascode:
So we don't force 24-hour time on everyone.


pynicotine/gtkgui/ui/settings/chats.ui
```diff
@@ -309,7 +309,7 @@
             <property name="visible">True</property>
             <child>
               <object class="GtkLabel">
-                <property name="label" translatable="yes">Private chat format:</property>
+                <property name="label" translatable="yes">Private chat format (strftime):</property>
```
slook, line 312: These appendages don't really provide much value.

slook (changes requested):

pynicotine/gtkgui/ui/settings/chats.ui
mathiascode: Ideas on how to present the format? Empty text boxes are useless too, but I'd rather not provide links to random websites. The format doesn't seem to have a friendly name unfortunately.

mathiascode: Links are fine, but I need stable ones. I put a link to the formatting codes in the Python documentation, but let me know if you find a better one.

slook: - If no value is set, then populate the text entry with the default locale value.
- Add a label beneath the entry, which presents on_entry_changed() the result of the entry format using the current time or an exemplar time such as the time the dialog was opened.

slook: https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes

mathiascode: I wonder if it would be easier to just change the default timestamp formats in the config. You can currently hide the timestamps by leaving the timestamp format blank, and I'm not sure if I want to change this behavior.

I think I'll just add a link for now to keep things simple. Maybe we could revisit some kind of preview in the future.

mathiascode:
Superseded by 181195bcbd88d18f492e37d086a9edc3bf741fdd

PR #2681, opened 2023-10-11
mainwindow.py: attempt to fix Away/Online status button

slook:
~WIP.~ I give up... it doesn't work.

Doesn't fix https://github.com/nicotine-plus/nicotine-plus/discussions/2145#discussioncomment-7219966


mathiascode:
https://github.com/nicotine-plus/nicotine-plus/commit/d3456354edb9d98b40709cdb26ded9751d6f2e99 should fix the button not working at all.

There's still the issue where the status initially reverts if you press the button while being auto-away, but it isn't as easy to solve.

PR #2692, opened 2023-10-18
Show chat room tabs before connecting to server

mathiascode:
Instead of waiting for a response from the server before showing a chat room tab, show it immediately.

Benefits:
- All tabs in Nicotine+ now behave the same
- Can read chat room history before connecting to the server
- Makes Nicotine+ feel faster, since slow connections no longer determine when tabs will be added
- If you're no longer able to join a private room that you've joined in the past, the tab still appears, allowing you to remove the room without manually editing the config. Redirect private messages from the server about room join failures to the appropriate chat room tab.

### DIFF
--- a/pynicotine/gtkgui/chatrooms.py
+++ b/pynicotine/gtkgui/chatrooms.py
@@ -86,6 +86,7 @@ class ChatRooms(IconNotebook):
             ("echo-room-message", self.echo_room_message),
             ("global-room-message", self.global_room_message),
             ("join-room", self.join_room),
+            ("leave-room", self.leave_room),
             ("remove-room", self.remove_room),
             ("room-completions", self.update_completions),
             ("say-chat-room", self.say_chat_room),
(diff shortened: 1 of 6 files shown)


pynicotine/gtkgui/chatrooms.py
```diff
@@ -952,9 +948,14 @@ def server_disconnect(self):
...
 
         self.chat_view.update_user_tags()
+        self.chat_entry.set_sensitive(False)
 
     def join_room(self, msg):
         self.populate_room_users(msg.users)
+        self.chat_entry.set_sensitive(True)
+
+    def leave_room(self, _msg):
+        self.server_disconnect()
```
slook, line 958: Put the code from `server_disconnect()` in here, and then call this function from there instead.

slook (changes requested):

pynicotine/privatechat.py
```diff
@@ -213,16 +213,24 @@ def _message_user(self, msg, queued_message=False):
...
             })
 
             core.send_message_to_server(slskmessages.MessageAcked(msg.msgid))
 
-        if username != "server":
+        if username == "server":
+            if message.startswith("The room you are trying to enter (") and ") " in message:
+                # Show fake chat room message if join wasn't successful
+                msg.user = None
+                room = message[message.find(" (") + 2:message.rfind(") ")]
```
slook, line 230: The position index of `" ("` is known within this explicit string, so there is no need to `find` it.

pynicotine/gtkgui/chatrooms.py
```diff
@@ -205,12 +206,30 @@ def clear_notifications(self):
...
+            is_global = (room == core.chatrooms.GLOBAL_ROOM_NAME)
 
-        if page is not None:
-            self.set_current_page(page.container)
+            if is_global:
+                tab_position = 0
+            elif remembered:
+                tab_position = -1
+            else:
+                tab_position = 1
```
slook, line 220: This is not the correct tab_position unless the Public global room tab is open.

Although, the bug existed before this PR.

pynicotine/gtkgui/chatrooms.py
```diff
@@ -205,12 +206,30 @@ def clear_notifications(self):
...
+
+            self.insert_page(
+                tab.container, room, focus_callback=tab.on_focus, close_callback=tab.on_leave_room,
+                position=tab_position
```
slook, line 224: `position=0 if is_global else -1`

pynicotine/privatechat.py
```diff
@@ -213,16 +213,26 @@ def _message_user(self, msg, queued_message=False):
...
             })
 
             core.send_message_to_server(slskmessages.MessageAcked(msg.msgid))
 
-        if username != "server":
+        if username == "server":
+            start_str = "The room you are trying to enter ("
+
+            if message.startswith(start_str) and ") " in message:
+                # Show fake chat room message if join wasn't successful
```
slook, line 230: `# Show fake chat room message `... -> `# Refer message to chat room tab `...

There is nothing "fake" about this, it's a really sensible idea.

slook (changes requested):

pynicotine/privatechat.py
mathiascode: It's fake in the sense that it's not an actual chat room message, but I suppose it can be interpreted as the original message itself being fake. I'll change the comment.

slook: > Not sure this specific change alone is worth mentioning

It is, because the user experience of this event is substantially altered.

PR #2698, opened 2023-10-20
CLI: prompt for username and password when necessary

mathiascode:
These changes are enough to get you up and running in headless mode, but avoids adding commands for logging in as another user or changing your password.

We should add these eventually, but it will require more discussion and work to get them right. For now, I'm simply adding the bare minimum to get Nicotine+ 3.3.0 out the door.

### DIFF
--- a/pynicotine/cli.py
+++ b/pynicotine/cli.py
@@ -21,6 +21,7 @@ import sys
 import time
 
 from collections import deque
+from getpass import getpass
 from threading import Thread
 
 from pynicotine.events import events
@@ -44,19 +45,20 @@ class CLIInputProcessor(Thread):
         self.has_custom_prompt = False
         self.prompt_message = ""
         self.prompt_callback = None
+        self.prompt_silent = False
 
     def run(self):
 
         while True:
+            # Small time window to set custom prompt
+            time.sleep(0.25)
+
             try:
                 self._handle_prompt()
 
             except Exception as error:
                 log.add_debug("CLI input prompt is no longer available: %s", error)
                 return
-
-            # Small time window to set custom prompt
-            time.sleep(0.25)
 
     def _handle_prompt_callback(self, user_input, callback):
 
@@ -85,13 +87,15 @@ class CLIInputProcessor(Thread):
     def _handle_prompt(self):
 
         callback = self.prompt_callback
+        input_func = getpass if self.prompt_silent else input
         self.has_custom_prompt = (callback is not None)
 
-        user_input = input(self.prompt_message)
+        user_input = input_func(self.prompt_message)
 
         self.has_custom_prompt = False
         self.prompt_message = ""
         self.prompt_callback = None
+        self.prompt_silent = False
 
         events.emit("cli-prompt-finished")
 
@@ -133,9 +137,11 @@ class CLI:
         ):
             events.connect(event_name, callback)
 
-    def prompt(self, message, callback):
+    def prompt(self, message, callback, is_silent=False):
+
         self._input_processor.prompt_message = message
         self._input_processor.prompt_callback = callback
+        self._input_processor.prompt_silent = is_silent
 
     def _print_log_message(self, log_message):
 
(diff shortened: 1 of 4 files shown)


mathiascode:
I forgot to look into https://docs.python.org/3/library/getpass.html, will do that before merging.

pynicotine/core.py
```diff
@@ -271,8 +271,10 @@ def quit(self, signal_type=None, _frame=None, should_finish_uploads=False):
 
     def connect(self):
 
+        if self.user_status == slskmessages.UserStatus.ONLINE:
+            return
+
         if config.need_config():
-            log.add(_("You need to specify a username and password before connecting…"))
```
slook, line 275: Restore this string.

pynicotine/headless/application.py
```diff
@@ -73,8 +74,30 @@ def on_confirm_quit(self, _only_on_active_uploads):
         cli.prompt(_("Do you really want to exit? %s") % responses, callback=self.on_confirm_quit_response)
 
     def on_invalid_password(self):
-        log.add(_("User %s already exists, and the password you entered is invalid. Please choose another username "
-                  "if this is your first time logging in."), config.sections["server"]["login"])
+
+        log.add(_("User %(user)s already exists, and the password you entered is invalid. Type %(command)s to "
+                  "log in with another username or password."), {
```
slook, line 79: Split this string, because console word wrap obscures the hint, and the first sentence is reusable in the gui.

`_("User %(user)s already exists, and the password you entered is invalid.")`
`+ "\n" + _("Type %(command)s to log in with another username or password.")`

pynicotine/headless/application.py
```diff
@@ -73,8 +74,30 @@ def on_confirm_quit(self, _only_on_active_uploads):
...
+
+    def on_setup_password_response(self, user_input):
+
+        config.sections["server"]["passw"] = user_input
+        config.write_configuration()
+
+        core.connect()
+
+    def on_setup_username_response(self, user_input):
+        config.sections["server"]["login"] = user_input
```
slook, line 94: If Username is blank, then keep the old `login` value, don't wipe out the existing.

```
if user_input:
    config.sections["server"]["login"] = user_input
```

pynicotine/headless/application.py
```diff
@@ -73,8 +74,30 @@ def on_confirm_quit(self, _only_on_active_uploads):
...
+            "user": config.sections["server"]["login"],
+            "command": "/connect"
+        })
+
+        config.sections["server"]["passw"] = ""
+
+    def on_setup_password_response(self, user_input):
+
+        config.sections["server"]["passw"] = user_input
+        config.write_configuration()
```
slook, line 89: Don't wipe the config if nothing was entered.

```
if not config.need_config():
    config.write_configuration()
```

pynicotine/core.py
```diff
@@ -271,8 +271,10 @@ def quit(self, signal_type=None, _frame=None, should_finish_uploads=False):
 
     def connect(self):
 
+        if self.user_status == slskmessages.UserStatus.ONLINE:
```
slook, line 274: Away is the same as online in this context.

`== slskmessages.UserStatus.ONLINE` -> `=! slskmessages.UserStatus.OFFLINE`

slook (changes requested):

pynicotine/cli.py
```diff
@@ -44,9 +45,13 @@ def __init__(self):
         self.has_custom_prompt = False
         self.prompt_message = ""
         self.prompt_callback = None
+        self.prompt_silent = False
 
     def run(self):
 
+        # Small time window for setup prompts on startup
+        time.sleep(1)
```
slook, line 53: Is it enough to use the existing `sleep` on Line 64 by moving it to the start of the `while` loop.

pynicotine/headless/application.py
mathiascode: I agree with using the previous username when pressing enter, but I don't think old passwords should remain.

PR #2706, opened 2023-10-24
downloads.py: add per-user download limit retries

### LINKED ISSUE #2555: Auto resume/retry

mathiascode:
Closes #2555

### DIFF
--- a/pynicotine/downloads.py
+++ b/pynicotine/downloads.py
@@ -46,6 +46,7 @@ class Downloads(Transfers):
         super().__init__(transfers_file_path=os.path.join(config.data_folder_path, "downloads.json"))
 
         self.transfers = {}
+        self.limited_user_timers = {}
         self.requested_folders = defaultdict(dict)
         self.requested_folder_token = 0
 
@@ -91,16 +92,14 @@ class Downloads(Transfers):
         # Request queue position of queued downloads and retry failed downloads every 3 minutes
         self._download_queue_timer_id = events.schedule(delay=180, callback=self.check_download_queue, repeat=True)
 
-        # Re-queue limited downloads every 12 minutes
-        self._retry_download_limits_timer_id = events.schedule(
-            delay=720, callback=self.retry_download_limits, repeat=True)
-
     def _server_disconnect(self, msg):
 
         super()._server_disconnect(msg)
 
-        for timer_id in (self._download_queue_timer_id, self._retry_download_limits_timer_id):
+        for timer_id in self.limited_user_timers.values():
             events.cancel_scheduled(timer_id)
+
+        events.cancel_scheduled(self._download_queue_timer_id)
 
         need_update = False
         ignored_statuses = {"Finished", "Filtered", "Paused"}
@@ -114,6 +113,7 @@ class Downloads(Transfers):
         if need_update:
             events.emit("update-downloads")
 
+        self.limited_user_timers.clear()
         self.requested_folders.clear()
 
     # Load Transfers #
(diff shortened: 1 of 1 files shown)


pynicotine/downloads.py
```diff
@@ -1042,20 +1048,39 @@ def retry_downloads(self, downloads):
...
+        """Re-queue limited downloads every 5 minutes."""
+
+        if not self.limited_user_times:
+            return
 
+        current_time = time.monotonic()
+        ready_users = set()
         limited_statuses = {"Too many files", "Too many megabytes"}
 
+        for username, time_limited in self.limited_user_times:
```
slook, line 1060: If we are checking this every second, then this loop is not required.

It would seem to be satisfactory enough to check if just the oldest item is 12 minutes old or not.

pynicotine/downloads.py
```diff
@@ -543,6 +545,10 @@ def _upload_denied(self, msg):
                 self.get_file(username, virtual_path, transfer=download)
                 break
 
+            if (reason in {"Too many files", "Too many megabytes"} or reason.startswith("User limit of")
+                    and username not in self.limited_user_times):
+                self.limited_user_times[username] = time.monotonic()
```
slook, line 550: Consider creating `events.schedule()` for each username instead of time for the value.

pynicotine/downloads.py
mathiascode: Looks like I forgot the timer values are in seconds. The timer is supposed to run every 60 seconds.

pynicotine/downloads.py
mathiascode: Since we'd have to iterate through the list of downloads for every limited user, I prefer keeping the current implementation.

mathiascode: Then again, we likely won't have many limited users. I'll investigate.

slook: Ideally it would be best to avoid any batch operations that have the potential to trigger a large iteration.

There is no rush to retry these failed downloads, since we have a window of 12 or 5 minutes to do this, so each pending username auto-retry might as well be spread out across that time.

mathiascode: Changed it to your initial suggestion.

pynicotine/downloads.py
```diff
@@ -1046,21 +1052,26 @@ def retry_downloads(self, downloads):
...
-                self.get_file(download.username, download.virtual_path, transfer=download)
+            log.add_transfer("Re-queuing file %(filename)s from user %(user)s in download queue", {
+                "filename": download.virtual_path,
+                "user": download.username
+            })
+
+            self.abort_download(download, abort_reason=None)
+            self.get_file(download.username, download.virtual_path, transfer=download)
+
+        del self.limited_user_timers[username]
```
slook, line 1074: Consider `break` out of the loop to only re-queue a single file, and come back in another 5 minutes to do the next download. After the last set a flag for if not any download found in transfers then delete username timer.

It might help to avoid the peer banning us due to hitting the limit so hard.

mathiascode: Only re-queuing a single file is a bit ridiculous when transfer speeds are fast, but we could limit the number of re-queued downloads to something reasonable, e.g. 100 (same as the default file limit for upload queues in Nicotine+).

slook: > we could limit the number of re-queued downloads to something reasonable, e.g. 100

If the limit set by the uploader could be determined with certainty, such as from knowing the number or total size of downloads that we had queued before encountering the limit, then it might be useful intelligence for the module to cache this to set the request rate deterministically, albeit this would be a lot more complex.

mathiascode:
Superseded by 002127e1d186d0c26655a2da2b62a94c8aea0eae

PR #2709, opened 2023-10-26
Performance improvements for transfers

mathiascode:
- Use a dictionary for download list for fast transfer object retrievals
- Update parent rows of individual transfers once per second, to prevent redundant updates in case of many upload rejection messages

### DIFF
--- a/pynicotine/downloads.py
+++ b/pynicotine/downloads.py
@@ -45,6 +45,7 @@ class Downloads(Transfers):
 
         super().__init__(transfers_file_path=os.path.join(config.data_folder_path, "downloads.json"))
 
+        self.transfers = {}
         self.requested_folders = defaultdict(dict)
         self.requested_folder_token = 0
 
@@ -71,7 +72,10 @@ class Downloads(Transfers):
         self.update_download_filters()
 
     def _quit(self):
+
         super()._quit()
+
+        self.transfers.clear()
         self.requested_folder_token = 0
 
     def _server_login(self, msg):
@@ -101,7 +105,7 @@ class Downloads(Transfers):
         need_update = False
         ignored_statuses = {"Finished", "Filtered", "Paused"}
 
-        for download in self.transfers:
+        for download in self.transfers.values():
             if download.status not in ignored_statuses:
                 download.status = "User logged off"
                 self.abort_download(download, abort_reason=None)
@@ -142,7 +146,8 @@ class Downloads(Transfers):
         if transfers_file_path != self.transfers_file_path:
             load_func = self.load_legacy_transfers_file
 
-        self.add_stored_transfers(transfers_file_path, load_func)
+        for transfer in self.get_stored_transfers(transfers_file_path, load_func):
+            self.transfers[transfer.username + transfer.virtual_path] = transfer
 
     def watch_stored_downloads(self):
         """When logging in, we request to watch the status of our downloads."""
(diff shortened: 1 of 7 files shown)


mathiascode:
Further improvements can probably be made (the uploads module is untouched for now, since it's more complex and needs to be profiled for bottlenecks).

In any case, it's a fairly straightforward change, and with 60k queued downloads, Nicotine+ recovers after a few seconds instead of hanging for several minutes.

@DeathStalker77

DeathStalker77:
> Further improvements can probably be made (the uploads module is untouched for now, since it's more complex and needs to be profiled for bottlenecks).
> 
> In any case, it's a fairly straightforward change, and with 60k queued downloads, Nicotine+ recovers after a few seconds instead of hanging for several minutes.
> 
> @DeathStalker77

Can I presume you'll also be testing with an ever increasing number of downloads?  (For example, sometimes you don't know how deeply nested or how many subfolders/files are in a parent folder; you may want to DL a "Rush Bootlegs" folder, but you have no way of knowing just how large that folder is, and it becomes .... challenging ... to try and DL parts/sections of folders at a time)  :-)  I look forward to trying this build though!

slook:
Can I presume you'll also be testing with an ever increasing number of downloads?

No, please feel free to test this scenario on your machine @DeathStalker77 

> I look forward to trying this build though!

Please do https://github.com/nicotine-plus/nicotine-plus/suites/17642508744/artifacts/1008665527

pynicotine/gtkgui/userbrowse.py
```diff
@@ -265,7 +265,7 @@ def __init__(self, userbrowses, user):
         self.folder_popup_menu = PopupMenu(self.window.application, self.folder_tree_view.widget,
                                            self.on_folder_popup_menu)
 
-        if user == config.sections["server"]["login"]:
+        if user != config.sections["server"]["login"]:
```
slook, line 268: for testing debug!

slook (changes requested):

pynicotine/downloads.py
```diff
@@ -360,7 +362,7 @@ def _transfer_request_downloads(self, msg):
 
                 transfer = Transfer(username=username, virtual_path=virtual_path, folder_path=folder_path,
                                     status="Queued", size=size, token=token)
-                self.transfers.appendleft(transfer)
+                self.transfers[username + virtual_path] = transfer
```
slook, line 365: Consider using `token` as key.

mathiascode: Tokens are only used for active downloads. We could add a separate active_transfers dict for them eventually, but there were no obvious bottlenecks, and I didn't want to complicate things too much yet.

mathiascode:
Note that this change will limit us to a single `username + virtual_path` list item at a time, instead of the previous `username + virtual_path + download_folder`. Seems like a fair trade-off considering the constant O(1) access times we gain.

If you download a user's file to some location, and decide to download it to another location later without clearing the old list item, the new download will not be added. Would it be better to automatically replace the old item in this case?

mathiascode:
Latest build for testing (scroll down for download links): https://github.com/mathiascode/nicotine-plus/actions/runs/6653722463

DeathStalker77:
> Latest build for testing (scroll down for download links): https://github.com/mathiascode/nicotine-plus/actions/runs/6653722463

This latest build, from the 26th, is DRASTICALLY improved!!!!  Currently handling over 130k files queued with only the occasional lag - no freezing (so far) - and currently at ~125k files.  It WILL still "pause" (running circle, but not white screen "freeze") when you do certain actions, like a manual resume.  Also unsure of the current status/functionality if a user's shares go offline (network issues, they move them, etc) and they then show up as FILE NOT SHARED - I know there is a separate thread for that issue though :-)   VERY impressed with such fast & efficient work!!!

PR #2717, opened 2023-10-29
test_startup.py: remove dependency on Xvfb

mathiascode:
Note that Ubuntu 22.04 doesn't ship gtk4-broadwayd, so we still need to use Xvfb there.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt update
-          sudo apt install gettext gir1.2-gtk-3.0 libgirepository1.0-dev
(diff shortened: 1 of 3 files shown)


mathiascode:
@yochananmarqos FYI, since you maintain the nicotine-plus-git AUR package.

yochananmarqos:
Arch does ship gtk4-broadwayd, by the way.

```
❯ pacman -Fl gtk4 | grep broadwayd
gtk4 usr/bin/gtk4-broadwayd
gtk4 usr/include/gtk-4.0/gdk/broadway/gdkbroadwaydisplay.h
gtk4 usr/share/man/man1/gtk4-broadwayd.1.gz
```

mathiascode:
Good, I'll merge this then. Let me know if something doesn't work.

PR #2722, opened 2023-10-30
Performance improvements for uploads

mathiascode:
Replaces the uploads deque with dictionaries for fast transfer object retrievals.

### DIFF
--- a/pynicotine/uploads.py
+++ b/pynicotine/uploads.py
@@ -20,7 +20,7 @@ import os
 import os.path
 import time
 
-from collections import deque
+from collections import defaultdict
 
 from pynicotine import slskmessages
 from pynicotine.config import config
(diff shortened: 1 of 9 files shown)


mathiascode:
12 years later, finally done. :)

https://github.com/nicotine-plus/nicotine-plus/commit/80654eae797285bb9758d3e75076bdeb4563b0e6

pynicotine/uploads.py
```diff
@@ -970,82 +975,72 @@ def get_upload_queue_size(self, username=None):
...
+                if self.is_privileged(upload.username):
                     queue_size += 1
 
             return queue_size
 
-        return sum(1 for upload in self.transfers if upload.status == "Queued")
+        return len(self.queued_uploads)
 
     def get_downloading_users(self):
+        return set(self.active_users) + set(self.queued_users)
```
slook, line 987: A user may exist in both sets, but only one instance is required.

`return set(self.active_users + self.queued_users)`

pynicotine/uploads.py
```diff
@@ -958,7 +963,7 @@ def get_total_uploads_allowed(self):
 
             return maxupslots
 
-        lstlen = sum(1 for upload in self.transfers if upload.sock is not None)
+        lstlen = sum(len(user_uploads) for user_uploads in self.active_users.values())
```
slook, line 966: Can the list length not be comprehended without iteration?

slook (changes requested):

Issue #2710, opened 2023-10-26
make banning user stop them from viewing profile


TheKfcBandit:
### Describe the solution you'd like
click ban user --> Dont reply to PM_USER_INFO_REQUEST

### Additional context
maybe block file search too if it isn't already

DeathStalker77:
If I might add a nuance to this.

I think banning PMs should be a separate action.  Banning, imho, should be the last resort for a "bad" user, not the first (Sadly, I've seen it used as the first option without even a message as to why).  I believe we should (somehow) enforce communication between users BEFORE a ban can occur.  There could be a miscommunication, or a share improperly read/detected, etc.  If you then, automatically cut off all communication because of a ban, then the issue can NEVER be rectified.  That, imho, would be a disservice to all, and I also feel it could be abused.

mathiascode:
Is the main concern that banned users can read your personal description? We currently hide the picture from banned users, and add a ban message above the description, but we could also hide the description if that helps.

mathiascode:
The only things that can be hidden are the description, picture and the upload slot-related stats (set them to some other value). Everything else is server-side.

mathiascode:
Not sure what difference it makes, since a user would presumably be banned while downloading something, and see the "Banned" status in the download list.

mathiascode:
Although I suppose we could go the same route as SoulseekQt, and change the "Banned" status to "File not shared": https://www.slsknet.org/news/node/156

If you set a custom ban reason, it would show up in the description, otherwise the description would be empty.

DeathStalker77:
I think that if a user is banned, the user should KNOW they are banned - and why.  People make mistakes - such as Leech Detector not showing any files because they accidentally changed drive letters, moved stuff and didn't refresh, etc.  Messaging them with a notification allows them to respond and correct any "issue".  I would not want to see "Ban Abuse" - for that matter, I REALLY don't like the "File Not Shared" option - that IS abused FAR too often.  Either something should be available & shared, or not available at all - and therefore not show up in searches.  Allowing the "File Not Shared" permits abuse by users with unreasonable "demands".

slook:
> an extra option

Please be specific, what kind of option (checkbox, radio buttons, dropdown), what string of text should the label(s) say, propose what the internal config name reference, etc.

PR #2737, opened 2023-11-05
Change 'Banned' status to 'File not shared.'

mathiascode:
Matches the behavior of the official client (SoulseekQt).

Related to #2710


mathiascode:
Note that this PR removes custom ban reasons from the status message, since they are non-standard and useless in most clients without resizing the status column. Custom ban reasons only appear on the user profile now.

PR #2738, opened 2023-11-05
Help VPN users

hboetes:
1. Add (For VPN) so VPN users know they should look into this setting.
2. Make tooltip easier to understand and translate


hboetes:
For after the next release.

slook (changes requested):

PR #2740, opened 2023-11-05
Statistics: remove started transfers from GUI

mathiascode:
Stats for both started and finished transfers were originally added when Nicotine+ had several bugs in its transfers module, and we wanted to know what percentage of transfers failed.

The statistics for started transfers have a flaw though: pausing and resuming a transfer counts as a new one. I don't think they are worth keeping, since the number of completed transfers seems more useful. It would also give us more space to potentially add more statistics to the GUI.

I don't want to remove any stored statistics yet, but I'd like to remove started transfers from the GUI, and see if we get any feedback/complaints about it. If not, we can remove them completely in a future release.

### DIFF
--- a/pynicotine/gtkgui/dialogs/statistics.py
+++ b/pynicotine/gtkgui/dialogs/statistics.py
@@ -42,10 +42,6 @@ class Statistics(Dialog):
             self.downloaded_size_total_label,
             self.reset_button,
             self.since_timestamp_total_label,
-            self.started_downloads_session_label,
-            self.started_downloads_total_label,
-            self.started_uploads_session_label,
-            self.started_uploads_total_label,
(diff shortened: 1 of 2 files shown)


slook:
On a loosely related topic, we should review the performance impact of updating the total number of Bytes Transferred after each in-progress chunk, as to whether or not it might be better to update that statistic after each completed/failed download has been completed/aborted, instead of after each chunk during the file transfer.

mathiascode:
> _Updtate:_ It doesn't really matter much after all, since the minor performance impact is only present whilst the Statistics dialog is visible.

Is the bottleneck getattr() or updating the label text? We could use a dictionary mapping stat IDs to the widgets, and check get_text() != new_text before updating the label.

PR #2743, opened 2023-11-06
Statistics: update ui values only if dialog is visible

slook:
Minor performance optimization

+ Fixed: After the Statistics dialog is shown once or more then closed then it's unnecessary to make the strings thereafter

### DIFF
--- a/pynicotine/transfers.py
+++ b/pynicotine/transfers.py
@@ -506,9 +506,6 @@ class Statistics:
         for stat_id in config.defaults["statistics"]:
             self.session_stats[stat_id] = 0 if stat_id != "since_timestamp" else int(time.time())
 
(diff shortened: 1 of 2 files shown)


pynicotine/transfers.py
```diff
@@ -525,10 +522,14 @@ def update_ui(self, stat_id):
 
         events.emit("update-stat-value", stat_id, session_stat_value, total_stat_value)
 
+    def refresh_stats(self):
```
mathiascode, line 525: It would be better if we don't mention "ui". Might as well rename this method to `update_stats`, and `update_ui` to `update_stat`.

slook: I concur.

Issue #2673, opened 2023-10-06
UPnP forwarding error: [Errno 126] Required key not available


pabateman:
Nicotine+ version:  3.2.9
Operating System/Distribution: Linux/Fedora 38 KDE

Nicotine+ comes from flatpak

### Describe the bug

Got an error right after start Nicotine+:

```
[2023-10-06 09:49:31] Loading GTK 3.24.38
[2023-10-06 09:49:31] Loading Python 3.11.5 (main, Nov 10 2011, 15:00:00) [GCC 13.2.0]
[2023-10-06 09:49:31] Loading Nicotine+ 3.2.9
[2023-10-06 09:49:31] Listening on port: 2236
[2023-10-06 09:49:31] UPnP: Failed to forward external port 2236: [Errno 126] Required key not available
[2023-10-06 09:49:31] Loading plugin system
[2023-10-06 09:49:31] Rescanning shares…
[2023-10-06 09:49:31] 0 folders found before rescan, rebuilding…
[2023-10-06 09:49:31] Rescan complete: 0 folders found
[2023-10-06 09:49:31] Connecting to server.slsknet.org:2242
[2023-10-06 09:49:31] Connected to server server.slsknet.org:2242, logging in…
[2023-10-06 09:49:32] 1901 privileged users
[2023-10-06 09:49:32] You have no privileges. Privileges are not required, but allow your downloads to be queued ahead of non-privileged users.
```

And have no forwarded ports from Nicotine+ on my home router. Other software like Transmission forwards ports well.

Manual port forwarding works great, but it would be better to use dynamic port opening.

### Expected behavior

No `UPnP: Failed to forward external port 2236: [Errno 126] Required key not available` error and forwarded ports on my router.

### Steps to reproduce the bug

Just run Nicotine+ with flatpak:

```
flatpak run org.nicotine_plus.Nicotine
```

slook:
Please can you test if forwarding works in the latest unstable build of Nicotine+ 3.3.0.dev5, link:

https://github.com/nicotine-plus/nicotine-plus/blob/master/doc/TESTING.md#flatpak

If it does, then no further action shall be required.

mathiascode:
Please also enable Miscellaneous debug logging in 3.2.9 first, since the unstable build added NAT-PMP support, which might hide issues with UPnP.

pabateman:
Hi. Done with both versions:

[...]

PR #2773, opened 2023-11-30
fix UPnP

stefantalpalaru:
Turns out you cannot `connect()` to a multicast address, so there was a good reason why Nicotine+-3.2.9 used `sendto()` instead of `send()`.

Further info: https://www.cs.unc.edu/~jeffay/dirt/FAQ/comp249-001-F99/mcast-socket.html

Fixes: https://github.com/nicotine-plus/nicotine-plus/issues/2673

### DIFF
--- a/pynicotine/portmapper.py
+++ b/pynicotine/portmapper.py
@@ -228,10 +228,10 @@ class UPnP(BaseImplementation):
                 "MX": str(UPnP.MX_RESPONSE_DELAY)
             }
 
-        def send(self, sock):
+        def sendto(self, sock, addr):
 
             msg = bytes(self)
-            sock.send(msg)
+            sock.sendto(msg, addr)
 
             log.add_debug("UPnP: SSDP request: %s", msg)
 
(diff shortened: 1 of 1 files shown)


mathiascode:
Not sure how this fixes #2673, since the code is the same as 3.2.9.

In any case, I've reverted the connect() change in master.

stefantalpalaru:
> I've reverted the connect() change in master

So first you refactor without testing, and then you shit on my fix by not merging my PR? What do you think that does to my willingness to try to contribute in the future?

mathiascode:
> So first you refactor without testing

It was confirmed to be working at the time. `connect()` + `send()` effectively does the same as `sendto()` for UDP sockets in Python, but might as well revert to the more idiomatic way since the change didn't make a difference.

> and then you shit on my fix by not merging my PR? What do you think that does to my willingness to try to contribute in the future?

A revert of the original commit seemed more appropriate. I apologize if this gave the wrong impression, I'd be happy to merge future PRs.

stefantalpalaru:
> It was confirmed to be working at the time. connect() + send() effectively does the same as sendto() for UDP sockets in Python

Not on Linux.

"If the socket sockfd is of type SOCK_DGRAM, then addr is the address to which datagrams are sent by default, and the only address from which datagrams are received." - https://man7.org/linux/man-pages/man2/connect.2.html

"In the current implementation, connection-oriented sockets are allowed to use only unicast addresses." - https://man7.org/linux/man-pages/man7/ip.7.html

So by calling connect() on a UDP socket with a multicast IP destination, you're telling the kernel to only accept incoming packets from that multicast IP address, but what you are actually expecting is the router to respond from its unicast address. Maybe that's why the kernel only wants unicast addresses there, because it's the only scenario that makes sense.

> A revert of the original commit seemed more appropriate. I apologize if this gave the wrong impression

I had to compare old and new UPnP code line by line, since debug messages gave no clue to why UPnP stopped working from last release to Git HEAD. It was not trivial, so to rob me of the symbolic credit that is having my commit added to Git history leaves a bitter taste. Be fair to your contributors and you'll get more contributions.

PR #2792, opened 2023-12-18
Chat View: use explicit username in append_line()

slook:
+ Fixes the issue where a part of the timestamp is tagged if the username is a short numeric.
+ Avoids the use of find() when inserting text for online chat messages.

Requires improvement of timestamp data type mixing string vs. integer.


mathiascode:
Thanks for working on this! I'll take a look once we've released 3.3.5. I think that would also be a good opportunity to create a 3.3.x branch and start development of 3.4.0 in master.

mathiascode:
Any changes/fixes in this branch that should be backported to 3.3.x? I see you've mentioned "Fixed: New room wall message was appended at the end instead of being prepended to start upon text entry" in the description, but is this an issue in master or exclusive to this branch?

slook:
> smaller changes/fixes in this branch that should be backported to 3.3.x?

#3201, #3202

> is this an issue in master or exclusive to this branch?

The issue exists in master, since we are currently unable to prepend any text whatsoever (without rewriting the entire contents).

However, the fix implemented in this PR is not fully effectual because for some unknown reason the textview doesn't seem to update itself properly when the popover is open so it's necessary to close and then reopen the popover in order to see the new message in any case.

mathiascode:
> The issue exists in master, since we are currently unable to prepend any text whatsoever (without rewriting the entire contents).

How do I reproduce it? It's true that we can't prepend text, but I'm pretty sure I wrote code that always clears the view before appending a new line.

slook:
_Profile of current master startup performance - `read_log()` happens BEFORE window.`present()`:_
_____________________________________________________________________________v ~2.9 seconds
image

_Profile of this PR startup performance - `read_log()` happens AFTER window.`present()`:_
__________________________________v ~1.5 seconds
image

mathiascode:
The biggest issue I'm having with the chat view right now is while scolling fast with a touchpad, or dragging the scrollbar to the top. Once you reach the very top of the chat view, GTK insists on sticking to the top when prepending new log lines, screwing up the scroll position in the process.

I tried to wrangle with the text view adjustments for an hour, but didn't get anywhere. It will probably require some delicate workaround...

mathiascode:
Seems like you can also get stuck in a loop loading log messages for a few seconds. It happens with a touchpad/smooth scrolling, when lightly nudging the scroll position up into the boundary for loading new messages.

slook:
> GTK insists on sticking to the top when prepending new log lines, screwing up the scroll position in the process

That sticking behaviour is caused because the "edge-reached" or "edge-overshoot" signal handler has to be used in this case, instead of our usual scroll position handler. Using the "edge-overshoot" signal makes it even more resistive than the "edge-reached" signal, but allows scrolling to occur even if the view was already at the top. I have switched it to "edge-reached" so you can feel the difference in behavior (however this breaks scroll-wheel scrolling). Firing both signals is also possible but results in a unpredictable number of read operations.

There is also a problem when using PageUp key then afterwards it is sometimes impossible to use PageDown immediately afterwards, and Down or mouse scrolling must be used instead, and I don't know what causes that to happen.

Those issues are somewhat mitigated by ensuring that more than the requested text arrives into the view before the next key-repeat, mouse scrolling or scroll-wheel event occurs. If it is too slow or not large enough then the scrollbar will reach the top-most position and the edge handler must kick in to resume it again, but the "edge-reached" handler doesn't fire in all cases such as if the view was already at the top (in which case the mouse or must be jiggled or the PageUp key must be used instead if the file isn't long enough to reveal the scroll bar).

Anyway, the inertial scrolling issue is irrelevant at the moment because the chunks are not loaded with the proper alignment. I have inserted a "!!! FIXME" line into the view to help identify where the gaps appear as a result of the faulty joining.

mathiascode:
Removing 3.4.0 milestone. We'll merge https://github.com/nicotine-plus/nicotine-plus/pull/3340 first, and probably work from there with another PR.

PR #2801, opened 2023-12-23
Browse Shares: multi-select in the folder view

stefantalpalaru:
It was already possible to select multiple files in the right panel. With this PR, it's now possible to select multiple folders in the left one.


mathiascode:
Thanks! I think this is something we can add. I'll review the changes when I have time.

mathiascode:
The folder context menu is currently reused for the dropdown menu in the path bar, which can result in the menu items targeting the wrong folder when you change the selection.

I think the easiest path forward would be to:

- Unset the active folder when multiple folders are selected, which in turn hides the path bar and dropdown menu
- Rename `self.selected_folder_path` to `self.active_folder_path` or `self.visible_folder_path`, and only use it in `set_selected_folder()`
- In all other cases, replace `self.selected_folder_path` with `self.folder_tree_view.get_selected_rows()`, like you've done in some cases already

pynicotine/gtkgui/userbrowse.py
```diff
@@ -252,7 +252,7 @@ def __init__(self, userbrowses, user):
         # Setup folder_tree_view
         self.folder_tree_view = TreeView(
             self.window, parent=self.folder_tree_container, has_tree=True, always_select=True,
```
mathiascode, line 254:
```suggestion
            self.window, parent=self.folder_tree_container, has_tree=True,
```

stefantalpalaru: Done.

stefantalpalaru:
Variable renamed as suggested. The path bar and file list view are now cleared when multiple folders are selected.

I'm not aware of any other cases where multiple folder selection needs to be handled.

PR #2813, opened 2023-12-30
userbrowse.py: recursive folder actions by default

### LINKED ISSUE #2812: Browse Shares: folder downloads should always be recursive
The first two options on the right click menu in the folder view are almost never what the user wants, because they do not download subfolders (which usually have artwork, or individual CDs). I propose we remove them and rename the next two options to remove "& Subfolders".

This will simplify the interface, make it more intuitive and protect the user from discovering this footgun the hard way.

In the off chance that someone actually wants to download only the top-level files in a folder, without subfolders, it would still be possible - by using the file view, on the right.

mathiascode:
Until we have something better, reuse the existing confirmation dialog used when downloading large folders from search results.

Closes #2812

### DIFF
--- a/pynicotine/gtkgui/userbrowse.py
+++ b/pynicotine/gtkgui/userbrowse.py
@@ -281,7 +281,6 @@ class UserBrowse:
 
         if user == config.sections["server"]["login"]:
             self.folder_popup_menu.add_items(
-                ("#" + _("Upload Folder…"), self.on_upload_folder_to),
                 ("#" + _("Upload Folder & Subfolders…"), self.on_upload_folder_recursive_to),
                 ("", None),
                 ("#" + _("Open in File _Manager"), self.on_file_manager),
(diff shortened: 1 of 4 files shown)


pynicotine/gtkgui/userbrowse.py
```diff
@@ -294,8 +293,6 @@ def __init__(self, userbrowses, user):
             )
         else:
             self.folder_popup_menu.add_items(
-                ("#" + _("_Download Folder"), self.on_download_folder),
-                ("#" + _("Download Folder _To…"), self.on_download_folder_to),
                 ("#" + _("Download Folder & Subfolders"), self.on_download_folder_recursive),
                 ("#" + _("Download Folder & Subfolders To…"), self.on_download_folder_recursive_to),
```
slook, line 297: The `D` and `T` keyboard accelerators are gone, perhaps they can be transposed into the remaining menu items instead.

slook (changes requested):

pynicotine/gtkgui/userbrowse.py
mathiascode: Will do once string freeze is over.

mathiascode:
Superseded by 5427c5ec212789c6433493df45457f78f3c200bd

PR #2823, opened 2024-01-11
autoban plugin added

zelmor:
This is a commit for an autoban plugin to be included by default. It handles banning and informing leechers, as well as removing the ban once they start contributing.

### DIFF
--- a/pynicotine/plugins/autobahn/__init__.py
+++ b/pynicotine/plugins/autobahn/__init__.py
@@ -1,0 +1,248 @@
+from pynicotine import slskmessages
+from pynicotine.pluginsystem import BasePlugin
+from pynicotine.config import config
+#from pynicotine.userbrowse import UserBrowse
+
+#def enable(plugins):
+#    global PLUGIN
+#    PLUGIN = Plugin(plugins)
+
+
+#def disable(plugins):
+#    global PLUGIN
+#    PLUGIN = None
+
+
+class Plugin(BasePlugin):
+    __name__ = "Autobahn"
+    settings = {
+        'ban_message': 'AUTOMATIC MESSAGE: You are not sharing enough and you were banned. You will be unbanned when you will contribute.',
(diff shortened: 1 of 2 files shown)


slook:
> autoban plugin to be included by default

It looks like a comprehenive plugin, but unfortunately this is not a feature to be included in the master package.

pynicotine/plugins/autobahn/__init__.py
```diff
@@ -0,0 +1,248 @@
...
+        except KeyError:
+            # we did not trigger this notification
+            return
+        if status == 'requesting':
+            #check if user file list has finished loading
+            #if not ready - request UserStats again - it is used as a timer hack to prevent locking UI
+            
+            #status check fails if user browse tab is closed manualy before fully loading
+            try:
+                total_size_text = self.frame.userbrowse.pages[user].share_size_label.get_text()
```
slook, line 73: We don't use references to `frame` in plugins anymore.

You have to consider headless mode, where there is no GUI.

pynicotine/plugins/autobahn/__init__.py
```diff
@@ -0,0 +1,248 @@
...
+            total_folders_text = self.frame.userbrowse.pages[user].num_folders_label.get_text()
+
+            total_size_array = total_size_text.split(" ")
+            total_size_number = float(total_size_array[0])
+            total_size_unit = total_size_array[1]
+            #self.log("after split %s, %s" % (total_size_number, total_size_unit))
+            if (total_size_unit == "B"):
+                total_size = total_size_number
+            elif (total_size_unit == "KiB"):
+                total_size = total_size_number * 1024
```
slook, line 137: You could use the `factorize()` function which is provided in the `utils` module:

https://github.com/nicotine-plus/nicotine-plus/blob/8a083b78201c4dce19be4279b4c6e05b7e4d17a7/pynicotine/utils.py#L136-L172

But it would be needed to get the raw size from the `core` anyway, such conversion is intended only for user inputs.

slook:
Closing sorry the plugin is not compatible due to breaking changes in 3.3.0.

Rings4:
Does anyone have a working version of this for Nicotine+ 3.3.x?

PR #2831, opened 2024-01-16
GUI: remove transfer tab highlights

mathiascode:
In their current state, these are mostly visual noise rather than being helpful. If you have a lot of transfers, the tabs are almost always highlighted. Auto-cleared transfers are not taken into account either.

If we bring back highlights in the future, we should be a lot more restrictive about when to activate them.

### DIFF
--- a/pynicotine/downloads.py
+++ b/pynicotine/downloads.py
@@ -465,8 +465,6 @@ class Downloads(Transfers):
         self._file_downloaded_actions(username, download_file_path)
         self._folder_downloaded_actions(username, transfer.folder_path)
 
-        finished = True
-        events.emit("download-notification", finished)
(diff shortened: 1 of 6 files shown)


mathiascode:
Thinking about this again, I like to have an indicator that transfers are working, but constantly highlighting the tab isn't helpful. How about the following solution?

- Downloads: Tab Changed when new rows are added, Tab Highlight when a folder finishes downloading
- Uploads: Tab Changed when new rows are added

slook:
Yes, that could be an improvement.

Work on this is not urgently needed until 3.3.1+

mathiascode:
I committed the change to master to get more feedback. Can be tweaked in future releases if necessary.

Issue #2260, opened 2022-11-16
TODO-list for Nicotine+ 3.3.0


mathiascode:
Nicotine+ 3.3.0 will be released mid to late 2023. This release will enable GTK 4 by default, revamp the command system, complete the Soulseek protocol implementation (distributed network), and introduce some new features.

In order to give a rough idea of where we currently stand, here's a list of required items before the release:

- [x] https://github.com/nicotine-plus/nicotine-plus/pull/2259
- [x] https://github.com/nicotine-plus/nicotine-plus/issues/2025
- [x] Refactor widget color system to avoid GTK 4 deprecations, implement https://github.com/nicotine-plus/nicotine-plus/issues/2074
- [x] Minor string improvements
- [x] Enable GTK 4 by default on every platform
- [x] https://github.com/nicotine-plus/nicotine-plus/issues/2141
- [x] Port all commands to the new plugin command system
- [x] https://github.com/nicotine-plus/nicotine-plus/issues/994
- [x] Implement tray icon and notification support in GTK 4 on Windows (https://github.com/nicotine-plus/nicotine-plus/pull/2584)
- [x] https://github.com/nicotine-plus/nicotine-plus/issues/885
- [x] Resize cursor gets stuck on macOS
- [x] Waiting for uploads to finish before quitting can get stuck
- [x] Finalize strings
- [x] Update release notes
- [x] Update screenshots
- [ ] Final testing

Depending on how quickly the above list is completed, here are things that would be nice to have:

- Finish treeview rewrite
- https://github.com/nicotine-plus/nicotine-plus/issues/1651 (move now playing functionality to a plugin, incomplete branch: https://github.com/mathiascode/nicotine-plus/tree/nowplaying-plugin)
- https://github.com/nicotine-plus/nicotine-plus/issues/771 (use new scheduler system)
- https://github.com/nicotine-plus/nicotine-plus/issues/110
- Wishlist overhaul with filters
- Other popular requests, check which ones

mathiascode:
I'll most likely not be able to work on Nicotine+ until June, so the new target for a 3.3.0 release is sometime during the second half of 2023.

toineenzo:
I see "Enable GTK 4 by default on every platform" has already been implemented. But on the latest Linux Test build I see the following in the about window: 3.3.0.dev5, Python 3.6.9, GTK 3.22.30. Still GTK 3?

mathiascode:
[...]

PR #2837, opened 2024-01-24
Version 3.3.0

mathiascode:
PR is mainly for testing CI in preparation for release. We could try to target February 1 as the release date if no serious bugs emerge, but we can revisit the situation in a week.


mathiascode:
@slook @hboetes Are we ready to release 3.3.0 today, or is there something left to do? I've tested on a few different distros, as well as Windows and macOS, and it seems to work well.

slook:
> is there something left to do? I've tested on a few different distros, as well as Windows and macOS, and it seems to work well.

You need to check the "Final testing" checkbox in https://github.com/nicotine-plus/nicotine-plus/issues/2260!

PR #2843, opened 2024-01-30
Add HIDDEN_FOLDER_NAMES class to shares.py for easier management to exclude multiple folders from shares

toineenzo:
Added the Synology #snapshot and #recycle folders to be excluded from Share as these folders contain data that have been already deleted which may be a security issues and may flood the SLSK network with unnecessary files.

### DIFF
--- a/pynicotine/shares.py
+++ b/pynicotine/shares.py
@@ -478,11 +478,9 @@ class Scanner:
 
         # If the last folder in the path starts with a dot, or is a Synology extended
         # attribute, snapshot or recycle bin folder, we exclude it
-        HIDDEN_FOLDER_NAMES = {"@eaDir", "#recycle", "#snapshot"}
-
         if filename is None:
             last_folder = os.path.basename(folder)
-
+            HIDDEN_FOLDER_NAMES = {"@eaDir", "#recycle", "#snapshot"}
             if last_folder.startswith(".") or last_folder in self.HIDDEN_FOLDER_NAMES:
                 return True
 
--- a/pynicotine/shares.py
+++ b/pynicotine/shares.py
@@ -478,13 +478,15 @@ class Scanner:
 
         # If the last folder in the path starts with a dot, or is a Synology extended
         # attribute, snapshot or recycle bin folder, we exclude it
+        HIDDEN_FOLDER_NAMES = {"@eaDir", "#recycle", "#snapshot"}
+        
         if filename is None:
             last_folder = os.path.basename(folder)
-
-            if last_folder.startswith(".") or last_folder in ["@eaDir", "#recycle", "#snapshot"]:
+            
+            if last_folder.startswith(".") or last_folder in self.HIDDEN_FOLDER_NAMES:
                 return True
 
-        # If we're asked to check a file we exclude it if it start with a dot
+        # If we're asked to check a file, we exclude it if it starts with a dot
         if filename is not None and filename.startswith("."):
             return True
 
--- a/pynicotine/shares.py
+++ b/pynicotine/shares.py
@@ -230,9 +230,10 @@ class Scanner:
     It handles scanning of folders and files, as well as building
     databases and writing them to disk.
     """
-    
+
+    # Add foldernames that need to be excluded from shares
     HIDDEN_FOLDER_NAMES = {"@eaDir", "#recycle", "#snapshot"}
-    
+
     def __init__(self, config_obj, queue, share_groups, share_db_paths, init=False, rescan=True,
                  rebuild=False, reveal_buddy_shares=False, reveal_trusted_shares=False):
 
--- a/pynicotine/shares.py
+++ b/pynicotine/shares.py
@@ -480,7 +480,9 @@ class Scanner:
         # attribute, snapshot or recycle bin folder, we exclude it
         if filename is None:
             last_folder = os.path.basename(folder)
+            
             HIDDEN_FOLDER_NAMES = {"@eaDir", "#recycle", "#snapshot"}
+            
             if last_folder.startswith(".") or last_folder in self.HIDDEN_FOLDER_NAMES:
                 return True
 
--- a/pynicotine/shares.py
+++ b/pynicotine/shares.py
@@ -230,7 +230,9 @@ class Scanner:
     It handles scanning of folders and files, as well as building
     databases and writing them to disk.
     """
-
+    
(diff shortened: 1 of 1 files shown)


pynicotine/shares.py
```diff
@@ -477,11 +477,11 @@ def is_hidden(folder, filename=None, entry=None):
         """Stop sharing any dot/hidden folders/files."""
 
         # If the last folder in the path starts with a dot, or is a Synology extended
-        # attribute folder, we exclude it
+        # attribute, snapshot or recycle bin folder, we exclude it
         if filename is None:
             last_folder = os.path.basename(folder)
 
-            if last_folder.startswith(".") or last_folder == "@eaDir":
+            if last_folder.startswith(".") or last_folder == ["@eaDir", "#recycle", "#snapshot"]:
```
slook, line 484: Condition is equal to entire list. You need `in` instead of `==`.

slook (changes requested):

pynicotine/shares.py
toineenzo: Thank you, fixed.

slook:
https://github.com/nicotine-plus/nicotine-plus/actions/runs/7718684220/job/21040477536?pr=2843#step:6:169
> `pynicotine/shares.py:484:61: R6201: Consider using set for membership test (use-set-for-membership)`

To use a set is as simple as using `{`curly brackets`}` in place of `[`square brackets`]` to contain the list of strings.

pynicotine/shares.py
```diff
@@ -477,11 +477,11 @@ def is_hidden(folder, filename=None, entry=None):
         """Stop sharing any dot/hidden folders/files."""
 
         # If the last folder in the path starts with a dot, or is a Synology extended
-        # attribute folder, we exclude it
+        # attribute, snapshot or recycle bin folder, we exclude it
         if filename is None:
             last_folder = os.path.basename(folder)
 
-            if last_folder.startswith(".") or last_folder == "@eaDir":
+            if last_folder.startswith(".") or last_folder in ["@eaDir", "#recycle", "#snapshot"]:
```
mathiascode, line 484: I would make the folder list a class constant, such as:
`HIDDEN_FOLDER_NAMES = {"@eaDir", "#recycle", "#snapshot"}`

then access it with

`if last_folder.startswith(".") or last_folder in self.HIDDEN_FOLDER_NAMES:`

toineenzo: Done

pynicotine/shares.py
```diff
@@ -477,14 +477,15 @@ def is_hidden(folder, filename=None, entry=None):
         """Stop sharing any dot/hidden folders/files."""
 
         # If the last folder in the path starts with a dot, or is a Synology extended
-        # attribute folder, we exclude it
+        # attribute, snapshot or recycle bin folder, we exclude it
         if filename is None:
             last_folder = os.path.basename(folder)
 
-            if last_folder.startswith(".") or last_folder == "@eaDir":
+            HIDDEN_FOLDER_NAMES = {"@eaDir", "#recycle", "#snapshot"}
```
mathiascode, line 484: Please move this to the beginning of the class, below the initial comment (line 232).

pynicotine/shares.py
```diff
@@ -477,14 +477,15 @@ def is_hidden(folder, filename=None, entry=None):
         """Stop sharing any dot/hidden folders/files."""
 
         # If the last folder in the path starts with a dot, or is a Synology extended
-        # attribute folder, we exclude it
+        # attribute, snapshot or recycle bin folder, we exclude it
```
mathiascode, line 480: This comment should preferably be more generic, so you don't have to update it if the list of hidden folders is updated in the future.

Perhaps something like this:

> If the last folder in the path starts with a dot, or uses the same name as well-known hidden folders, we exclude it

mathiascode:
I forgot the is_hidden() method is static. Change the function header to:

```
    @classmethod
    def is_hidden(cls, folder, filename=None, entry=None):
```

and `self.HIDDEN_FOLDER_NAMES` to `cls.HIDDEN_FOLDER_NAMES`

pynicotine/shares.py
mathiascode: No worries!

slook:
It's too late to implement this new feature whilst in release candidate stage, so I propose we pause this PR until the next development cycle in 3.3.1dev1.

pynicotine/shares.py
toineenzo: I've moved it below this:
```
class Scanner:
    """Separate process responsible for building shares.

    It handles scanning of folders and files, as well as building
    databases and writing them to disk.
    """
```

toineenzo:
> It's too late to implement this new feature whilst in release candidate stage, so I propose we shouldn't merge this PR until the next development cycle in 3.3.1dev1.



> It's too late to implement this new feature whilst in release candidate stage, so I propose we shouldn't merge this PR until the next development cycle in 3.3.1dev1.

Will the excluding of the @eaDir folder still be included in 3.3.0 stable as currently in the latest 3.3.0-rc3?

toineenzo:
> It's too late to implement this new feature whilst in release candidate stage, so I propose we shouldn't merge this PR until the next development cycle in 3.3.1dev1.

Will the excluding of the @ eadir folder still be included in 3.3.0 stable as currently in the latest 3.3.0-rc3?

pynicotine/shares.py
```diff
@@ -472,19 +475,19 @@ def rescan_dirs(self, permission_level):
...
-    @staticmethod
-    def is_hidden(folder, filename=None, entry=None):
+    @classmethod
+    def is_hidden(cls, folder, filename=None, entry=None):
         """Stop sharing any dot/hidden folders/files."""
 
-        # If the last folder in the path starts with a dot, or is a Synology extended
-        # attribute folder, we exclude it
+        # If the last folder in the path starts with a dot, or uses
+        # the same name as well-known hidden folders, we exclude it
```
slook, line 483: Remove this entire comment, the code is self-explainatory.

pynicotine/shares.py
```diff
@@ -231,6 +231,9 @@ class Scanner:
     databases and writing them to disk.
     """
 
+    # Add foldernames that need to be excluded from shares
```
slook, line 234: Remove this entire comment, the code is self-explainatory.

slook:
> Will the excluding of the @ eadir folder still be included in 3.3.0 stable as currently in the latest 3.3.0-rc3?

Yes.

mathiascode (approved):

PR #2890, opened 2024-02-20
GUI: rework window parenting

mathiascode:
Instead of explicitly specifying a parent window, let the application provide an appropriate one instead.


slook:
What is the advantage?

mathiascode:
Easier to maintain, in my opinion. No need to ensure we're always giving the correct parent window to a dialog, the previously focused window will always be used.

This is likely 3.3.x material, but there are some smaller unrelated cleanups that could be added to master now.

mathiascode:
On second thought, I don't want this change now.

PR #2892, opened 2024-02-21
userbrowse.py: handle usernames with invalid path chars in share lists

mathiascode:
- Add .json file extension to lists saved to disk
- Add metadata section containing the original username, in case it contained characters that were removed from the filename

### DIFF
--- a/pynicotine/userbrowse.py
+++ b/pynicotine/userbrowse.py
@@ -186,6 +186,9 @@ class UserBrowse:
     def load_shares_list_from_disk(self, file_path):
 
         file_path_encoded = encode_path(file_path)
+        suffix = None
+        metadata_index = None
+        username = None
 
         try:
             try:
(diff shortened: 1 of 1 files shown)


slook:
This bugs seems low priority, I'd suggest waiting to merge this until 3.3.2.dev1.

slook:
Can the new metadata field be used to contain other data in future, for example, favorite folders, file and folder counts?

Having such extra metadata would be faster than using iteration for counting items, in order for a feature or plugin to quickly check if the shares list needs updating to get newly added items. Not saying to add anything extra now, just saying the metadata field is a really good idea.

Does it need testing to see what happens if the list is loaded by an older version?

mathiascode:
Sorry, I let this one slip. I'll take a look at the code again.

mathiascode:
Still not sure about the format, not quite happy with it. I'll give it some more thought when I have time.

slook:
I'm not confident that the current JSON format meets our future needs.

Maybe we should consider a fully fledged database format the same as the local shares scanner.

mathiascode:
It needs to be human-readable. That was one of the reasons for abandoning the old pickle-based format.

slook:
I'm a bit worried that we could end up trying to re-invent the wheel here.

In general, what do you think about [sqlite3](https://docs.python.org/3/library/sqlite3.html)?

mathiascode:
Maybe sqlite could be an option in the far future, but I don't really want to do another migration anytime soon.

We could probably do something that doesn't touch the file contents for now, e.g. append the base64-encoded username to the end of the filename for problematic usernames. Something similar could be done for chat room and private chat logs too.

PR #2904, opened 2024-02-29
"User Actions" menu labels

slook:
+ Changed: Label of user actions popup menu to match that used everywhere else.


mathiascode:
Private chats should be updated as well.

mathiascode:
Thanks!

PR #2915, opened 2024-03-01
shares.py: don't block send_num_shared_folders_files()

slook:
+ Fixed: The numbers were prevented from updating because method only runs after `self.rescanning` is set to `False`


mathiascode:
Ugh, thanks for noticing.

PR #2917, opened 2024-03-02
Threaded transfers loader

### LINKED ISSUE #2910: App irreparably frozen after accidentally queueing up a very large download
Nicotine+ version:  3.3.2

Operating System/Distribution: macOS Monterey Version 12.7.3

### Describe the bug

Nicotine+ app has been rendered completely useless via the spinning beach ball of death no matter how many times I force quit and restart it after accidentally starting up a download of an absolutely massive folder

### Expected behavior

I'd expect to be able to clear all downloads and actually use the app like normal

The bug is omnipresent since this morning and doesn't go away regardless of how many times I force quit and restart the app

### Additional context

I'm not sure what would be relevant or helpful, so please let me know if I can share anything that might help someone troubleshoot this issue with me (having said that, I can't do much navigating around the app as it is frozen)

slook:
Draft concept to close #2910

On initial testing it doesn't appear to actually make any difference. I think Multiprocessing is needed?

Theoretically, this could be dangerous unless `self._allow_saving_transfers` is used to block most functions.

### DIFF
--- a/pynicotine/downloads.py
+++ b/pynicotine/downloads.py
@@ -30,6 +30,7 @@ import shutil
 import time
 
 from collections import defaultdict
+from threading import Thread
 
 from pynicotine import slskmessages
 from pynicotine.config import config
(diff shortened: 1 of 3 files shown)


slook:
Thank you I was confident that you would have a better suggestion like that, so I'd like us to work on #2952 instead.

Issue #2861, opened 2024-02-05
3.3.0: does not connect to server via tun0 (vpn)


rexendevar:
Nicotine+ version:  3.3.0
Operating System/Distribution: Pop OS 22.04

### Describe the bug
Since updating yesterday, Nicotine will no longer connect to the server when it is configured to connect via `tun0`. It will repeatedly throw timeout errors instead.

### Expected behavior
Nicotine should work just as well with the VPN as it does without it (aside from inevitable decrease in throughput).

### Steps to reproduce the bug
Set up and run the OpenVPN service on your operating system (`sudo service openvpn start`). I use Mullvad for the VPN but it shouldn't matter. 
Open Nicotine+ and go to the network settings. Set the network interface to `tun0` (the vpn) and keep the soulseek server at default configuration (`server.slsknet.org:2242`). Restart Nicotine and attempt to connect.

### Additional context
I'm on a college network and have no control over which ports are open on the internet. I don't think this affects things but I mention it just in case.

mathiascode:
If you enable "Miscellaneous" [debug logging](https://nicotine-plus.org/doc/DEVELOPING.html#debug-logging), the local IP address should be included in the logs (`Local IP address: x.x.x.x`). Does it look correct?

rigrig:
Same problem here (except with a different VPN) 

The debug log shows
`[Misc] Local IP address: <IP-address of the VPN interface>`

Version 3.2.9 works fine, but doesn't seem to log the IP.

mathiascode:
Can you try starting Nicotine+ 3.2.9 with the following command, using the IP logged in 3.3.0, and see if the issue happens there too?

`nicotine --bindip <IP>`

This should be the same way of binding to an interface that Nicotine+ 3.3.0 uses.

rigrig:
Yes, running Nicotine+ 3.2.9 with that argument also fails to connect.   
(I checked the VPN interface still has that same IP)

mathiascode:
Strange. In 3.3.0, if you unset the network interface while the VPN is active, does Nicotine+ connect? Does the IP address remain the same?

rigrig:
It did connect, the IP address changed to that of the default interface (`enp5s0` / `192.168.2.2`)

After selecting either `enp5s0` or my Wi-Fi interface it still connects, so the problem is only with the VPN interface.

mathiascode:
[...]

Issue #2518, opened 2023-05-29
Unable to Connect to SoulSeek Server when Dual Network Interface Is Configured


frnksmdlkedjnfr:
Nicotine+ version: 3.2.8  •  GTK 3.24.36
Operating System/Distribution: Windows 11 Pro for Workstations 22H2 22621.1778

**### Describe the bug**
My company isolates the Internet from the Intranet, so I had to write some extra route commands to access the Internet. However, while 99% of applications and websites work fine, Nicotine+ is an exception. It just won't connect to the SoulSeek server. No system proxy is being using. The following GIF describes the issue I'm having:
![录制_2023_05_29_09_29_36_278](https://github.com/nicotine-plus/nicotine-plus/assets/63704404/23692215-e205-4fd8-8bd7-ec00602f080a)

tracert looks normal:
![2023-05-29_09-53-22_Administrator Windows Pow
![2023-05-29_09-53-22_Administrator Windows PowerShell](https://github.com/nicotine-plus/nicotine-plus/assets/63704404/afc6add3-e39e-436f-8a3e-2d5b451f6447)
erShell](https://github.com/nicotine-plus/nicotine-plus/assets/63704404/98203b8d-98c9-4280-bb83-5a7eb7618f11)

**### Expected behavior**
![2023-05-29_09-59-06_Nicotine+](https://github.com/nicotine-plus/nicotine-plus/assets/63704404/730d8642-d351-4089-95b1-9143a1b8e649)

**### Steps to reproduce the bug**

1. Plug in the LAN cable;
2. Wait for the Ethernet to identify;
3. Execute the following CMD command to configure the route table:

`route delete 0.0.0.0 mask 0.0.0.0 10.134.48.1 metric 35 -p`
`route delete 10.0.0.0 mask 255.0.0.0 10.134.48.1 metric 50`

**### Additional context**
My route table:
![2023-05-29_10-05-01_Administrator Windows PowerShell](https://github.com/nicotine-plus/nicotine-plus/assets/63704404/3a6b8aef-3101-41fa-81b5-4f0e571378a5)

**Screenshots, logs, stacktraces or relevant information.**
1. 10 out of 10 times, unplug the LAN cable and Nicotine+ instantly connects;
2. With LAN cable connected, after 15 minutes or so, Nicotine+ would randomly connects;
3. My coworker from the IT department suspects that Nicotine+ isn't optimised for dual NIC routes.

slook:
[...]

Issue #871, opened 2020-11-24
Bind Nicotine+ to specified network adapter


dochdoch:
I'd like to be able to choose a network adapter to bind Nicotine+ to.

This is useful for VPNs, etc. qBittorrent has implemented this very nicely as an advanced preference via a simple dropdown menu.

gits68:
much better would be to have the possibility to bind the incoming & outcoming interfaces to différent network adapter.
for example, I've an ADSL and a 4G network, can't use the 4G network for download bcoz it is not possible to listen on it...
while it would be possible to download using the 4G network and upload using the ADSL one...

felipehw:
I agree with @dochdoch .
Actually, I need to use `nicotine  -b VpnIP` from CLI.  I always need to look at the actual IP from the VPN interface before invoking the command.
If we could (just as we do with qBittorrent) just set the interface from the GUI ... these boring procedures vanish.

mathiascode:
I assume this option is to ensure that only the VPN is used for traffic at all times, or is there more to it? Are there cases where the VPN isn't used by default? I'm not too familiar with this area myself, so any insight or contributions would be appreciated.

felipehw:
@mathiascode 

If your VPN turns off and you don't perceive ... Nicotine connects without VPN.
If we can attach Nicotine to an interface (like `tun0`), this kind of problem doesn't happen.

mathiascode:
This is now done for Linux and macOS. *BSD and Windows will require more work.

dochdoch:
Thank you very much for all your hard work!

felipehw:
@mathiascode Thanks a lot man!

mathiascode:
https://github.com/nicotine-plus/nicotine-plus/commit/5949bf0bd7bea04b8fd3b300e8b162a9a673ccd8 should add this functionality on *BSD. Would be much appreciated if any *BSD users could test this.

slook:
To test binding to a particular network adapter with Preferences > Network > Network Interface in the [latest unstable build](https://github.com/nicotine-plus/nicotine-plus/blob/master/doc/TESTING.md) of master branch 3.3.0.dev4

mathiascode:
Confirmed working on OpenBSD: https://github.com/nicotine-plus/nicotine-plus/commit/5949bf0bd7bea04b8fd3b300e8b162a9a673ccd8#commitcomment-92062623

I've also verified that it works on FreeBSD.

[...]

PR #2925, opened 2024-03-04
slskproto.py: broadcast with connect_ex() in local_ip_address()

slook:
+ Changed: to avoid exception 'OSError: [Errno 101] Network is unreachable' # (doesn't need to be reachable)

Edge-case platform-sepcific regression since add35dc0, exception is handled in `_bind_listen_port()` which throws to log:

```
Your branch is up to date with 'origin/master'.
~/git/slook/nicotine-plus $ python3 ./nicotine
[2024-03-04 10:54:53] Loading Python 3.11.8
[2024-03-04 10:54:53] Loading Nicotine+ 3.3.3.dev1
[2024-03-04 10:54:53] Translation files (.mo) are unavailable, using default English strings
[2024-03-04 10:54:53] Loading GTK 4.12.4
[2024-03-04 10:54:54] Loading plugin system
[2024-03-04 10:54:54] Loaded plugin Nicotine+ Commands
[2024-03-04 10:54:54] Reconnecting to server in 12 seconds
[2024-03-04 10:54:54] Cannot listen on port 2234. Ensure no other application uses it, or choose a different port. Error: [Errno 101] Network is unreachable
[2024-03-04 10:54:55] Rescanning shares…
[2024-03-04 10:54:55] Rescan complete: 0 folders found
[2024-03-04 10:55:06] Cannot listen on port 2234. Ensure no other application uses it, or choose a different port. Error: [Errno 101] Network is unreachable
[2024-03-04 10:55:08] Quitting Nicotine+ 3.3.3.dev1, application closing…
[2024-03-04 10:55:08] Quit Nicotine+ 3.3.3.dev1!
```

Connection cannot be established on Android Oreo 8.1.0 AOSP (non-GMS) in Termux native Linux shell or Debian GNU/Linux 6.2.1-PRoot-Distro aarch64, because for some reason this environment won't broadcast to any local broadcast address other than `0.0.0.0`. Connectivity in other apps, using wlan0 interface (served by WiFi mobile hotspot), is fine.

It's possible (but not tested) that other similar platforms might also be prevented from connecting due to this error. The exception is not intended this case, because we don't actually need to use the local broadcast address for anything else afterwards, see link:

> [socket.`connect_ex`(_address_)](https://docs.python.org/3.8/library/socket.html#socket.socket.connect_ex)
> _"Like connect(address), but return an error indicator instead of raising an exception for errors returned by the C-level connect() call (other problems, such as “host not found,” can still raise exceptions)."_

Alternatively, `sys.platform != "darwin"` could be used to set the broadcast ip address to `0.0.0.0` as it was before add35dc0, but using `connect_ex()` seems like the more proper solution, see log output:

```
~/git/slook/nicotine-plus $ git switch headless-termux-connect
Switched to branch 'headless-termux-connect'
~/git/slook/nicotine-plus $ python3 ./nicotine
[2024-03-04 10:55:20] Loading Python 3.11.8
[2024-03-04 10:55:20] Loading Nicotine+ 3.3.3.dev1
[2024-03-04 10:55:20] Translation files (.mo) are unavailable, using default English strings
[2024-03-04 10:55:21] Loading GTK 4.12.4
[2024-03-04 10:55:22] Loading plugin system
[2024-03-04 10:55:22] Loaded plugin Nicotine+ Commands
[2024-03-04 10:55:22] Listening on port: 2234
[2024-03-04 10:55:22] Connecting to server.slsknet.org:2242
[2024-03-04 10:55:22] Connected to server server.slsknet.org:2242, logging in…
[2024-03-04 10:55:22] UPnP: Failed to forward external port 2234: [Errno 101] Network is unreachable
[2024-03-04 10:55:22] Rescanning shares…
[2024-03-04 10:55:22] Rescan complete: 0 folders found
[2024-03-04 10:55:22] 1988 privileged users
[2024-03-04 10:55:22] You have no Soulseek privileges. While privileges are active, your downloads will be queued ahead of those of non-privileged users.
```

Now, the application can be run in either headless mode or with GTK 4 using X11/Xfce on this platform, see screenshot...

image

... showing "Online" after the fix in this PR.

### DIFF
--- a/pynicotine/slskproto.py
+++ b/pynicotine/slskproto.py
@@ -585,7 +585,7 @@ class NetworkThread(Thread):
 
             # Send a broadcast packet on a local address (doesn't need to be reachable,
             # but MacOS requires port to be non-zero)
-            local_socket.connect(("10.255.255.255", 1))
+            local_socket.connect_ex(("10.255.255.255", 1))
 
             # This returns the "primary" IP on the local box, even if that IP is a NAT/private/internal IP
             ip_address = local_socket.getsockname()[0]


mathiascode:
The intention was to only listen on the port on the current network interface, but I suppose we have no option but to fall back to all interfaces (0.0.0.0) on this platform.

slook:
> I suppose we have no option but to fall back to all interfaces (0.0.0.0) on this platform

After some further experimentation I become concerned about the effect of this on other systems, whose interface binding preference is set because it is for the purpose of preventing any leakage over the default interface.

I determine this concern because an invalid binding being set still allows for a connection to be established, but I cannot properly test because my network environment doesn't feature any UPnP nor VPN and is behind a public NAT. A slightly different design of the module might be required to avoid such fallback in the case of an interface binding, that is beyond my own capability to develop.

Further testing is required perhaps the testers in issues #2861 #2518 #871 and others can be alerted to try latest master to verify if binding fallback behaviour during VPN network outage is correct in the real world use case?

mathiascode:
This part of the code prevents connections if a specified interface isn't available: https://github.com/nicotine-plus/nicotine-plus/blob/9e5d5feb274baddac24c5ec042fbef05ae71cbb2/pynicotine/slskproto.py#L485-L488

In other words, this PR makes no difference in this case.

slook:
How does PortMapper ensure that the forwarding request is made to the correct specified bound interface, when binding is set?

PR #2930, opened 2024-03-05
search.py: minor code cleanup imports

slook:
+ Changed: Import only the required function `random` from the `random` package
+ Changed: Import only the required function `shlex` from the `shlex` package

It's not needed to load an entire package if only a single function is called from it.


mathiascode:
> It's not needed to load an entire package if only a single function is called from it.

The whole module will be imported regardless, but from import looks a little better here.

PR #2936, opened 2024-03-12
gtkgui: select selected file on_browse_folder()

slook:
+ Added: Select the selected file when browsing to the folder from a "_Browse Folder" context menu action

The `_basename` data already exist but they were discarded, yet there is no good reason to truncate the `path` because in most cases where this function is called (usually from a file selection) it seems expected to select a file in the shares, if it still exists there.

### DIFF
--- a/pynicotine/gtkgui/downloads.py
+++ b/pynicotine/gtkgui/downloads.py
@@ -199,11 +199,13 @@ class Downloads(Transfers):
 
         transfer = next(iter(self.selected_transfers), None)
 
-        if transfer:
-            user = transfer.username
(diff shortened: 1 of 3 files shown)


mathiascode:
I have a vague memory of some discussion about this. If multiple transfers are selected, I think we also wanted to select multiple files when browsing a folder?

slook:
That would of course be ideal, but not really essential. At the moment we don't have an easily available method of passing a multiple selection, because that would require much larger changes. I think a single selection is better than nothing, and in fact it is satisfactory enough for most use cases.

PR #2952, opened 2024-03-21
gtkgui: async rows

slook:
Superseeds PR #2917 to close issue #2910

> GTK is single-threaded. You'd need to implement something like ... adding rows in batches (currently, only user browse works somewhat, but there are several edge cases you need to handle).

### DIFF
--- a/pynicotine/gtkgui/transfers.py
+++ b/pynicotine/gtkgui/transfers.py
@@ -24,6 +24,7 @@
 
 from itertools import islice
 
+from gi.repository import GLib
 from gi.repository import GObject
 from gi.repository import Gtk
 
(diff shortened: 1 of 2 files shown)


slook:
When browsing a user's saved list the progress bar disappears before the tree is fully loaded and it's possible to navigate the folder tree view while items are still being added to it, which is feels great but the progress bar doesn't indicate that this is happening.

mathiascode:
I'd like to prioritize the port to GtkColumnView, which will fix the performance issues for GTK 4 users without the need for this PR. I might continue working on this, but I'm not sure, considering GTK 4 is turning 4 years old, and we want to phase out GTK 3 eventually.

PR #2960, opened 2024-04-01
Replace IP2Location binary database with tuples

mathiascode:
Removes IP2Location's custom binary database in favor of something more auditable.

Uses a bit more memory than before, but performance is better.


pynicotine/external/data/ipcountry.py
slook, line 1: If the ability is provided to get the CSV file is going to be provided within the package, and there is no executable code in here, then the existence of this large file seems unnecessary.

The fact that the file cannot be easily viewed because it is so large, makes it difficult to audit.

pynicotine/external/update_ipcountry_data.py
```diff
@@ -0,0 +1,95 @@
...
+    # work. If not, see <http://creativecommons.org/licenses/by-sa/4.0/>.
+
+    """)
+    country_codes = set()
+    ip_ranges = []
+    country_values = []
+
+    with zipfile.ZipFile(FILE_PATH, "r") as zip_file_handle:
+        with zip_file_handle.open(zip_file_handle.namelist()[-1]) as csv_file_handle:
+            for row in csv.reader(io.TextIOWrapper(csv_file_handle)):
```
slook, line 50: Perhaps the values could be loaded directly from a local copy of the csv file at runtime?

pynicotine/external/update_ipcountry_data.py
```diff
@@ -0,0 +1,95 @@
...
+    output.extend(b"# Country codes for IP ranges\n")
+    output.extend(b"country_values = (")
+
+    for country_code in country_values:
+        output.extend(f'{country_code.replace("-", "_")},'.encode("utf-8"))
+
+    output.extend(b")\n\n")
+
+    with open(os.path.join(CURRENT_PATH, "data", "ipcountry.py"), "wb") as file_handle:
+        file_handle.write(output)
```
slook, line 91: It isn't best practice to programmatically generate a (potentially executable) script file like this.

slook (changes requested):

mathiascode:
We don't have many options:
- No data should be downloaded during the build process, since we want reproducible builds.
- Downloading data on runtime from random URLs is something I won't accept.
- The original CSV file is about 12 MB, containing redundant data we don't need, so storing it in the git repo as-is is not an option.

What I could look into is generating a slimmed-down CSV file, storing it in the repo, and converting it to Python-compatible format on runtime.

pynicotine/external/update_ip_country_data.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+COUNTRY_DB_URL = "https://download.ip2location.com/lite/IP2LOCATION-LITE-DB1.CSV.ZIP"
+FILE_PATH, _HEADERS = urllib.request.urlretrieve(COUNTRY_DB_URL)
+CURRENT_PATH = os.path.dirname(os.path.realpath(__file__))
+MAX_IPV4_RANGE = 4294967295
+
+
+def update_ip_country_data():
+    """Dowmload IP2Location country data and convert it for use in Python."""
```
slook, line 33: Dowmload -> Download

pynicotine/external/data/ips.py
slook, line 1: ips.py -> ip_ranges.py

slook (changes requested):

setup.cfg
```diff
@@ -66,7 +66,7 @@ enable =
     use-implicit-booleaness-not-comparison-to-string,
     use-implicit-booleaness-not-comparison-to-zero
 ignore = tinytag.py
-ignore-paths = build, venv
+ignore-paths = build, pynicotine/external/data, venv
```
slook, line 69: It would be interesting to see what errors the linter says.

slook:
Nice job. I agree that your approach is sound, this is much easier to audit compared to the black box of the current binary file.

As there are two separate data files, it seems like it would be a good idea to compare file timestamps in CI to verify that they are both at least created on the same day/month. Perhaps store a variable containing the creation date when the update script was run, maybe even show that date in the generated about text or license file.

> how often are meaningful changes made?

Rarely, and in any case the accuracy of the country data is not much concern to end users, given the displayed flag icons are not really that critical to the proper functioning of the program. It is only necessary to update perhaps on an annual basis, or if there is some significant published change of policy for international address allocation.

data/scripts/update_ip_country_data.py
```diff
@@ -0,0 +1,103 @@
...
+                    raise ValueError("Invalid country code")
+
+                if address_to < 0 or address_to > max_ipv4_range:
+                    raise ValueError("Invalid IP address")
+
+                country_codes.add(country_code)
+                ip_ranges.append(address_to)
+                country_values.append(country_code)
+
+    output_ip_ranges = bytearray(b"""# IP2Location LITE (c) 2001-2024 Hexasoft Development Sdn. Bhd.
```
slook, line 62: Suggest automate the insertion of the copyright year text from the source license file.

pynicotine/external/README.md
```diff
@@ -1,38 +1,467 @@
...
+- Audio metadata parser, modified for Nicotine+
+- Original source code: https://github.com/devsnd/tinytag
 
 ## License
 
-MIT
+MIT License
+
+Copyright (c) 2014-2023 Tom Wallroth, Mat (mathiascode)  
+Copyright (c) 2020-2023 Nicotine+ Contributors
```
slook, line 449: 2024

pynicotine/tests/unit/networkfilter/test_country.py
```diff
@@ -0,0 +1,70 @@
...
+
+        for ip_address, country_code in (
+            ("0.255.255.255", ""),
+            ("1.0.0.0", "US"),
+            ("1.0.0.255", "US"),
+            ("1.0.1.0", "CN"),
+            ("9.255.255.255", "US"),
+            ("10.0.0.0", ""),
+            ("13.255.255.255", "US"),
+            ("14.0.0.0", "CN")
```
slook, line 58: Some testing of European ranges would be nice.

Consider making a test with some calls to a web API lookup service to check the country codes still match.

pynicotine/external/README.md
mathiascode: I'll change this once we update tinytag. The code was last updated in 2023.

mathiascode:
Timestamps are a good idea. I'll implement them.

setup.cfg
mathiascode: The linter uses too much memory when processing the file.

data/scripts/update_ip_country_data.py
```diff
@@ -0,0 +1,132 @@
...
+
+def parse_ip_country_data():
+    """Download and parse IP2Location country data."""
+
+    country_data_url = "https://download.ip2location.com/lite/IP2LOCATION-LITE-DB1.CSV.ZIP"
+    copyright_prefix = b"Copyright (c)"
+    expected_copyright_author = b"Hexasoft Development"
+    expected_license = b"Creative Commons Attribution-ShareAlike 4.0 International"
+    max_ipv4_range = 4294967295
+    file_path, _headers = urllib.request.urlretrieve(country_data_url)
```
slook, line 36: Perhaps there are data in `headers` that can be used to verify the authenticity of the source remote host?

Ideally the author's public key could be parsed to compare it for a match, albeit that might be a bit overkill... 

Perhaps something more trivial can be checked to satisfy the operator of the script that the domain is still under the same ownership organisation or else bail, or that the file is within a sane size that would be expected, etc.

pynicotine/networkfilter.py
```diff
@@ -407,9 +412,16 @@ def get_online_username(ip_address):
...
     def get_country_code(self, ip_address):
 
-        country_code = self._ip2location.get_country_code(ip_address)
+        try:
+            ip_num, = UINT32_UNPACK(inet_aton(ip_address))
+            ip_index = bisect_left(ip_ranges.values, ip_num)
+            country_code = country_codes.values[ip_index]
 
-        if country_code is None or country_code == "-":
+        except Exception as error:
```
slook, line 420: Exactly which type of Exception(s) might be expected here?

I only predict ValueError or IndexError, but I don't know about what types of errors the socket or Struct modules can throw.

slook (changes requested):

pynicotine/tests/unit/networkfilter/test_country.py
```diff
@@ -0,0 +1,90 @@
...
+    def tearDown(self):
+        core.quit()
+        self.assertIsNone(core.network_filter)
+
+    @classmethod
+    def tearDownClass(cls):
+        shutil.rmtree(DATA_FOLDER_PATH)
+
+    def test_ip_country_data_valid(self):
+        """Verify that IP country data is valid."""
```
slook, line 49: """Verify that IP country data _structure_ is valid."""

pynicotine/networkfilter.py
mathiascode: My main concern was invalid IP addresses from the network, but I realized that the message parser would already catch those.

data/scripts/update_ip_country_data.py
mathiascode: I added a size check. Anything else should be caught by the parser.

slook (approved):

mathiascode:
Thanks for the review! This turned out larger than expected, but we can easily modify the script to download CSV data from other IP country data providers in the future, if necessary.

PR #2974, opened 2024-04-11
added white icon theme set

philcryer:
This PR adds a White Icon Theme set generated in Inkscape from the example set. This fits in better with newer iterations of Gnome which use all white icon themes by default.


mathiascode:
Thank you for your PR. I will look closer at it after the next Nicotine+ release.

I believe some systems have a concept of symbolic tray icons that are recolored automatically for light/dark themes, but I don't know how widespread this is. Symbolic UI icons should be easy to support though. I'll have to check if I've already implemented this for custom icon themes.

PR #2979, opened 2024-04-16
Fix last.fm now playing

everlyy:
It wouldn't work because of a 403 HTTP status, but this was because
the program showed that you had to put a colon after username;apikey
I removed this from the UI and now it works.


mathiascode:
This requires new translated strings, which we want to avoid for new 3.3.x releases. I think it would be better to wait for https://github.com/nicotine-plus/nicotine-plus/pull/2406, which we could change to move the username and API key into separate text entries.

PR #3008, opened 2024-05-10
Use TCP_NODELAY on the server socket

stefantalpalaru:
Rationale: https://news.ycombinator.com/item?id=40310896

### DIFF
--- a/pynicotine/slskproto.py
+++ b/pynicotine/slskproto.py
@@ -1717,6 +1717,8 @@ class NetworkThread(Thread):
         sock.setblocking(False)
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_RCVBUF, self.SOCKET_READ_BUFFER_SIZE)
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_SNDBUF, self.SOCKET_WRITE_BUFFER_SIZE)
+        if hasattr(socket, "TCP_NODELAY"):
+            sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)  # pylint: disable=no-member
         self._bind_socket_interface(sock)
 
         try:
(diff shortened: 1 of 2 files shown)


mathiascode:
Sure, why not. We already have our own buffer, so Nagle's algorithm doesn't really benefit us anyway.

Could you set TCP_NODELAY on the peer sockets as well?

pynicotine/slskproto.py
```diff
@@ -1214,6 +1214,9 @@ def _init_server_conn(self, msg_obj):
         server_socket.setsockopt(socket.SOL_SOCKET, socket.SO_RCVBUF, self.SOCKET_READ_BUFFER_SIZE)
         server_socket.setsockopt(socket.SOL_SOCKET, socket.SO_SNDBUF, self.SOCKET_WRITE_BUFFER_SIZE)
 
+        if hasattr(socket, "TCP_NODELAY"):
```
mathiascode, line 1217: I believe TCP_NODELAY should be available on every OS we support. Do you know one that doesn't have it?

stefantalpalaru:
> Could you set TCP_NODELAY on the peer sockets as well?

No. Those are UDP.

pynicotine/slskproto.py
stefantalpalaru: Not aware of any without it. I was just copying the safety checks in https://github.com/nicotine-plus/nicotine-plus/blob/5ecf2df5718d5454cb63370556261a343022b561/pynicotine/slskproto.py#L1144-L1177

mathiascode: I think we can leave the check out for now.

mathiascode:
> > Could you set TCP_NODELAY on the peer sockets as well?
> 
> No. Those are UDP.

The Soulseek protocol uses TCP, not UDP. Peer sockets are created here:
https://github.com/nicotine-plus/nicotine-plus/blob/f8ecc7b63e8928430705afad9a6bd8a80346d148/pynicotine/slskproto.py#L1710
and here:
https://github.com/nicotine-plus/nicotine-plus/blob/f8ecc7b63e8928430705afad9a6bd8a80346d148/pynicotine/slskproto.py#L2386

stefantalpalaru:
I stand corrected. Also added for peer sockets.

pynicotine/slskproto.py
stefantalpalaru: The CPython code itself checks for the corresponding C define: https://github.com/python/cpython/blob/7e894c2f38f64aed9b259c8fd31880f1142a259d/Modules/socketmodule.c#L8561

stefantalpalaru: OK. Check removed.

mathiascode (approved):

PR #3023, opened 2024-06-02
slskproto.py: handle MemoryError

### LINKED ISSUE #3022: Python Memory Error
Nicotine+ version:  3.3.4
Operating System/Distribution: Windows 11 Pro 23H2

### Describe the bug
When I leave Nicotine+ open for a few hours, I get the following error pop-up.

### Expected behavior
I expected to Nic+ to stay minimized and running in the background.
But then after a hour I get the following error message.

Just opening Nic+ and leave it running for a few hours I get the message again.

### Additional context

<a href="https://imgbb.com/">Screenshot-2024-06-01-232458</a>

Nicotine+ Version: 3.3.4
GTK Version: 4.14.3
Python Version: 3.11.9 (win32)

Type: <class 'MemoryError'>
Value: 
Traceback:   File "D:/a/nicotine-plus/nicotine-plus/pynicotine/gtkgui/application.py", line 884, in _raise_exception
  File "C:/msys64/mingw64/lib/python3.11/threading.py", line 1045, in _bootstrap_inner
  File "D:/a/nicotine-plus/nicotine-plus/pynicotine/slskproto.py", line 2725, in run
  File "D:/a/nicotine-plus/nicotine-plus/pynicotine/slskproto.py", line 2492, in _process_ready_sockets
  File "D:/a/nicotine-plus/nicotine-plus/pynicotine/slskproto.py", line 2433, in _process_ready_input_socket
  File "D:/a/nicotine-plus/nicotine-plus/pynicotine/slskproto.py", line 2606, in _read_data

slook:
Fixes #3022

An exception handler already exists if `_read_data()` throws `OSError`, so it is inexpensive to handle the case of `MemoryError` here also. 

Closing the connection is probably the best action to take in such circumstances. The program should be able to recover itself enough to continue running, albeit in a limited capacity, refusing connections until either...

- demand for resources reduces (such as peers giving up), or;
- more system resources become available (such as another application terminating).

... As such normal operation would hopefully be resumed eventually, in theory, but this needs further testing in the wild to determine if further actions are required or not (for example, setting `MAX_SOCKETS` in accordance with available memory, or cleaning up old connection data that might be still hanging around in memory for no good reason).

### DIFF
--- a/pynicotine/slskproto.py
+++ b/pynicotine/slskproto.py
@@ -2440,7 +2440,7 @@ class NetworkThread(Thread):
                     self._close_connection(self._conns, sock)
                     return
 
-            except OSError as error:
+            except (OSError, MemoryError) as error:
                 log.add_conn(("Cannot read data from connection %(addr)s, closing connection. "
                               "Error: %(error)s"), {
                     "addr": conn_obj_established.addr,


mathiascode:
I think this might work as a last resort, but memory usage shouldn't keep growing in the first place. I'll try to get to the root of the issue first.

slook:
It is difficult to imagine exactly what the client is doing that would result in so much of the system resources being consumed. My initial guess would be serving a vast number of peers, likely with incoming queue requests or search requests, but there might be other causes, perhaps if the network is very slow (or limited?) then maybe it could take a while for data pending on the stack to be released.

In any case if an OOM error is not impossible then it needs to be handled in some way, but admittedly what is offered here is a workaround that only masks the problem, which could be detrimental as it might lead to indeterminate behaviours.

mathiascode:
3872336c0c6beb948f8789fa88c1db4c5976c520 should handle this better.

PR #3024, opened 2024-06-02
Implement file download dialog

### LINKED ISSUE #1659: implement dialog with download options for folder downloads
Hello. Thanks for the excellent Nicotine+ software. I have a feature request. 

This is one of my favourite SoulseekQT features not present in Nicotine+. When I download a folder from another user, the UI allows me to change the target folder name before starting the download. This is different from selecting the target container folder. It could be useful when the folder name is, for example, "CD1". I would like it to store that folder on my computer as "Artist name - year - album CD1".

Also, SoulseekQT allows selecting which files of the folder will be downloaded in case you don't want to download every file present in the folder. That's also a neat feature not present in Nicotine+

Thank you very much.

### LINKED ISSUE #2686: "download folder" sometimes only selects 1 file
Nicotine+ version:  3.3.0.dev5   •   Python 3.10.12   •   GTK 3.24.33
Operating System/Distribution: Ubuntu

### Describe the bug

Sometimes, when using the "download folder" option in the search result, only 1 file is added to que queue, and not the whole contents of the remote folder.
At first i though i misclick on the "download file" option, by error, but finally i found one occurence where i am sure i used the right option. After noticing that only one file is added, i go again to the search results and click again in "download folder", and then it works as expected.
I think it correlates to when i am searching for a single track (within an album), and then try to download the containing folder. This is, the other requested files are not part of the displayed search results.

### Expected behavior

If the folder content can't be retrieved, i'd expect that at least the operation is not performed (and avoid silently adding only 1 file), with a warning. 

I don't know how to precisely reproduce this. And in my experience it only happens sporadically

### Additional context

When this happens and only 1 file is queued, it is created within a named folder (like the remote folder). This should confirm that i was not using the "download files" action, because in that case the files would be created without replicating the remote folder structure.

mathiascode:
TODO:
- [ ] Code cleanups
- [x] Fix bug where progress bar doesn't disappear in some cases
- [x] Improve checkbox behavior
- [x] Show total number and size of selected files in title
- [x] Remove full path from rename dialog
- [ ] Better error message in info banner
- [ ] Performance improvements? Need to test with large folders.
- [ ] Probably more I'm forgetting right now

Maybe an option to add downloads in paused mode, but I want to implement https://github.com/nicotine-plus/nicotine-plus/issues/110 first to see if this is needed.

Closes #1659 
Closes #2686

### DIFF
--- a/pynicotine/downloads.py
+++ b/pynicotine/downloads.py
@@ -65,13 +65,11 @@ from pynicotine.utils import truncate_string_byte
 
 
 class RequestedFolder:
-    __slots__ = ("username", "folder_path", "download_folder_path", "request_timer_id", "has_retried",
-                 "legacy_attempt")
+    __slots__ = ("username", "folder_path", "request_timer_id", "has_retried", "legacy_attempt")
 
-    def __init__(self, username, folder_path, download_folder_path):
+    def __init__(self, username, folder_path):
         self.username = username
         self.folder_path = folder_path
-        self.download_folder_path = download_folder_path
         self.request_timer_id = None
         self.has_retried = False
         self.legacy_attempt = False
@@ -638,12 +636,7 @@ class Downloads(Transfers):
 
         # Check if a custom download location was specified
         if not download_folder_path:
-            requested_folder = self._requested_folders.get(username, {}).get(folder_path)
-
-            if requested_folder is not None and requested_folder.download_folder_path:
-                download_folder_path = requested_folder.download_folder_path
-            else:
-                download_folder_path = self.get_default_download_folder(username)
+            download_folder_path = self.get_default_download_folder(username)
 
         # Merge download path with target folder name
         return os.path.join(download_folder_path, target_folders)
@@ -762,25 +755,21 @@ class Downloads(Transfers):
 
         return self.get_incomplete_download_file_path(transfer.username, transfer.virtual_path)
 
-    def enqueue_folder(self, username, folder_path, download_folder_path=None):
+    def request_folder(self, username, folder_path):
 
         requested_folder = self._requested_folders.get(username, {}).get(folder_path)
 
         if requested_folder is None:
             self._requested_folders[username][folder_path] = requested_folder = RequestedFolder(
-                username, folder_path, download_folder_path
+                username, folder_path
             )
-
-        # First timeout is shorter to get a response sooner in case the first request
-        # failed. Second timeout is longer in case the response is delayed.
-        timeout = 60 if requested_folder.has_retried else 15
 
         if requested_folder.request_timer_id is not None:
             events.cancel_scheduled(requested_folder.request_timer_id)
             requested_folder.request_timer_id = None
 
         requested_folder.request_timer_id = events.schedule(
-            delay=timeout, callback=self._requested_folder_timeout, callback_args=(requested_folder,)
+            delay=15, callback=self._requested_folder_timeout, callback_args=(requested_folder,)
         )
 
         log.add_transfer("Requesting contents of folder %s from user %s", (folder_path, username))
@@ -982,15 +971,16 @@ class Downloads(Transfers):
             log.add_transfer("Folder content request for folder %s from user %s timed out, "
                              "giving up", (folder_path, username))
             del self._requested_folders[username][folder_path]
+            events.emit("folder-contents-timeout", username, folder_path)
             return
 
         log.add_transfer("Folder content request for folder %s from user %s timed out, "
                          "retrying", (folder_path, username))
 
         requested_folder.has_retried = True
-        self.enqueue_folder(username, folder_path, requested_folder.download_folder_path)
+        self.request_folder(username, folder_path)
 
-    def _folder_contents_response(self, msg, check_num_files=True):
+    def _folder_contents_response(self, msg):
         """Peer code 37."""
 
         username = msg.username
@@ -1014,34 +1004,8 @@ class Downloads(Transfers):
         if not msg.list and not requested_folder.legacy_attempt:
             log.add_transfer("Folder content response is empty. Trying legacy latin-1 request.")
             requested_folder.legacy_attempt = True
-            self.enqueue_folder(username, folder_path, requested_folder.download_folder_path)
+            self.request_folder(username, folder_path)
             return
-
-        for i_folder_path, files in msg.list.items():
-            if i_folder_path != folder_path:
-                continue
-
-            num_files = len(files)
-
-            if check_num_files and num_files > 100:
-                check_num_files = False
-                events.emit(
-                    "download-large-folder", username, folder_path, num_files,
-                    self._folder_contents_response, (msg, check_num_files)
-                )
-                return
-
-            destination_folder_path = self.get_folder_destination(username, folder_path)
-
-            log.add_transfer("Attempting to download files in folder %s for user %s. "
-                             "Destination path: %s", (folder_path, username, destination_folder_path))
-
-            for _code, basename, file_size, _ext, file_attributes, *_unused in files:
-                virtual_path = folder_path.rstrip("\\") + "\\" + basename
-
-                self.enqueue_download(
-                    username, virtual_path, folder_path=destination_folder_path, size=file_size,
-                    file_attributes=file_attributes)
 
         del self._requested_folders[username][folder_path]
 
--- a/pynicotine/gtkgui/dialogs/download.py
+++ b/pynicotine/gtkgui/dialogs/download.py
@@ -1,0 +1,680 @@
+# COPYRIGHT (C) 2024 Nicotine+ Contributors
+#
+# GNU GENERAL PUBLIC LICENSE
+#    Version 3, 29 June 2007
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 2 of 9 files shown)


slook:
on click "Download..." in Search with large selection spanning multiple users...

```
Nicotine+ Version: 3.3.5.dev1
GTK Version: 3.24.24
Python Version: 3.9.2 (linux)

Type: <class 'KeyError'>
Value: <Gtk.TreeIter object at 0x7f8795bd9280 (GtkTreeIter at 0x49a5a00)>
Traceback:   File "/home/user/Git/mathiascode/nicotine-plus/pynicotine/gtkgui/application.py", line 979, in on_process_thread_events
    return events.process_thread_events()
  File "/home/user/Git/mathiascode/nicotine-plus/pynicotine/events.py", line 294, in process_thread_events
    self.emit(event_name, *args, **kwargs)
  File "/home/user/Git/mathiascode/nicotine-plus/pynicotine/events.py", line 250, in emit
    function(*args, **kwargs)
  File "/home/user/Git/mathiascode/nicotine-plus/pynicotine/gtkgui/dialogs/download.py", line 308, in folder_contents_response
    self.initial_selected_iterators.remove(parent_iterator)
```

pynicotine/gtkgui/userbrowse.py
```diff
@@ -375,12 +380,7 @@ def __init__(self, userbrowses, user):
             )
         else:
             self.file_popup_menu.add_items(
-                ("#" + _("_Download File(s)"), self.on_download_files),
-                ("#" + _("Download File(s) _To…"), self.on_download_files_to),
-                ("", None),
-                ("#" + _("_Download Folder"), self.on_download_folder),
-                ("#" + _("Download Folder _To…"), self.on_download_folder_to),
-                ("", None),
+                ("#" + _("_Download…"), self.on_download_files),
```
slook, line 383: The dialog implementation is not required in the `file_popup_menu` because the `userbrowse` interface already provides the needed selection for what files to download.

Instead, consider replacing the "Download Folder & Subfolders _To…" function in `folder_popup_menu`, if it is possible to pass file data recursively from cached shares data, then the dialog would be helpful to preview exactly what will be queued and to choose a download location to download to.

slook (changes requested):

pynicotine/gtkgui/userbrowse.py
mathiascode: > Instead, consider replacing the "Download Folder & Subfolders _To…" function in `folder_popup_menu`, if it is possible to pass file data recursively from cached shares data (while keeping the full path structure), then the dialog would be helpful to preview exactly what will be queued and to choose a download location to download to.

I'll see what I can do. What I was concerned about was the four different download items in the file menu. Adding the download dialog for recursive folder downloads seems like a good solution that would allow us to remove the two folder download items from the file menu.

mathiascode: Should be working now.

mathiascode:
> on click "Download..." in Search with large contiguous selection spanning multiple users...

Thanks, fixed.

pynicotine/gtkgui/dialogs/download.py
```diff
@@ -0,0 +1,594 @@
...
+                self.num_selected_files += 1
+                self.total_selected_size += tree_view.get_row_value(iterator, "size_data")
+            else:
+                self.num_selected_files -= 1
+                self.total_selected_size -= tree_view.get_row_value(iterator, "size_data")
+
+        self.update_title()
+
+    def on_download_folder_selected(self):
+        self.has_custom_download_location = True
```
slook, line 533: This might not be `True` if the user happens to manually select their default download folder. This would lead to a bug where the "usernamesubfolders" setting is not respected in `on_download()`.

slook (changes requested):

pynicotine/gtkgui/ui/dialogs/download.ui
```diff
@@ -0,0 +1,291 @@
...
+                <property name="margin-end">6</property>
+                <property name="margin-start">6</property>
+                <property name="margin-top">6</property>
+                <property name="spacing">6</property>
+                <property name="visible">True</property>
+                <child>
+                  <object class="GtkButton" id="_rename_button">
+                    <property name="tooltip-text" translatable="yes">Rename…</property>
+                    <property name="visible">True</property>
+                    <signal name="clicked" handler="on_rename"/>
```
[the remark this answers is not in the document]
mathiascode: Thanks, I knew some things slipped my mind. GUI work is exhausting. :)

slook:
This needs to be rebased and pushed again, since you have changed a function name.

pynicotine/downloads.py
```diff
@@ -1006,39 +996,9 @@ def _folder_contents_response(self, msg, check_num_files=True):
         if not msg.list and not requested_folder.legacy_attempt:
             log.add_transfer("Folder content response is empty. Trying legacy latin-1 request.")
             requested_folder.legacy_attempt = True
-            self.enqueue_folder(username, folder_path, requested_folder.download_folder_path)
+            self.request_folder(username, folder_path, requested_folder.download_folder_path)
```
slook, line 999: https://github.com/nicotine-plus/nicotine-plus/actions/runs/9416175846/job/25938718002?pr=3024#step:6:15

> pynicotine/downloads.py:999:12: E1121: Too many positional arguments for method call (too-many-function-args)

slook (changes requested):

pynicotine/gtkgui/dialogs/download.py
```diff
@@ -0,0 +1,594 @@
...
+
+    def on_default_download_folder(self, *_args):
+        self.download_folder_button.set_path(config.defaults["transfers"]["downloaddir"])
```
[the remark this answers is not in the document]
mathiascode: Fixed the first issue. The rest will have to be changed in the preferences dialog as well.

pynicotine/gtkgui/dialogs/download.py
```diff
@@ -0,0 +1,672 @@
...
+
+    def folder_contents_response(self, msg):
+
+        username = msg.username
+        selected = self.select_all
+        has_added_file = False
+
+        for folder_path, files in msg.list.items():
+            if username not in self.pending_folders:
+                continue
```
slook, line 349: `break`

slook (changes requested):

slook:
On click Download Folder in search, the dialog does present and is operable for a few seconds but then it freezes the whole application for many minutes, it is up to over 500 seconds now and still waiting...

image

```
            self.iterators[key] = iterator = self.model.insert_with_values(  # pylint: disable=no-member
                parent_iterator, position, self._column_numbers, values
            )
```

There is no way of aborting the folder listing (it must be really large), so the application has to be killed in this case.

It would be ideal if the Cancel button could be pressed, or otherwise an arbitrary limit on the number of entries in the list is needed to avoid freezing.

mathiascode:
> ** Browse Shares can navigate to the same remote folder and list all 99161 files in about 5 seconds (after waiting about 20 seconds for getting the user's entire shares listing)... 

I'm guessing something about appending a large number of rows to a parent row is horribly inefficient.

mathiascode:
Should be a lot faster now. I disabled sorting while adding rows.

mathiascode:
We will still have to come up with some kind of protection, or as you said, maybe hide the file rows after a certain point. Currently trying to recursively download the root of a share will hammer the dialog.

pynicotine/gtkgui/dialogs/download.py
```diff
@@ -0,0 +1,680 @@
...
+        self.download_button.set_sensitive(True)
+
+    def set_failed(self):
+
+        if not self.indeterminate_progress:
+            return
+
+        self.set_finished()
+        self.info_bar.show_error_message(
+            _("Failed to request folder contents.")
```
slook, line 333: > Better error message in info banner

`"Failed to get folder contents from user $(username)s"`

However, a single failure doesn't mean that requests made to other users in the selection might still respond (and they could fail as well)...

`"Failed to get folder contents from $(num_failed_users)i users"`

pynicotine/downloads.py
```diff
@@ -741,25 +734,21 @@ def get_current_download_file_path(self, transfer):
...
-        # failed. Second timeout is longer in case the response is delayed.
-        timeout = 60 if requested_folder.has_retried else 15
-
         if requested_folder.request_timer_id is not None:
             events.cancel_scheduled(requested_folder.request_timer_id)
             requested_folder.request_timer_id = None
 
         requested_folder.request_timer_id = events.schedule(
-            delay=timeout, callback=self._requested_folder_timeout, callback_args=(requested_folder,)
+            delay=15, callback=self._requested_folder_timeout, callback_args=(requested_folder,)
```
slook, line 751: Often a failure is due to the user being Offline, so this delay is far too long to wait for the error to appear.

Suggest abandoning the `FolderContentsRequest()` in this case (since it is only creating unnecessary network traffic anyway) and calling back immediately with a new method like `requested_folder_offline()` with no delay, if possible.

slook:
This TODO item can be ticked now, since an issue was spotted and resolved satisfactorily, given the scope of this PR...

>> **Performance improvements? Need to test with large folders.**

> Should be a lot faster now. I disabled sorting while adding rows.

... Any further needed performance improvements can be revised in the future.

slook:
Bump we don't want this diverge for too long, it's not too far off!

mathiascode:
I could rebase I suppose, but I'll work on this when I have time.

slook:
Ace thank you, I'll give it another test run to make sure it's still solid. From what I remember this branch was working like a dream.

ngs428:
This will be a great enhancement.  Hope progress can be made to a successful outcome.

mathiascode:
We should be able to get this merged in the beginning of the 3.4.0 development cycle.

ngs428:
> We should be able to get this merged in the beginning of the 3.4.0 development cycle.

Awesome, thanks!!!🙏

mathiascode:
Rebased. More tweaks are needed, but I'll merge this as-is to get more feedback.

ngs428:
Excellent!  Thanks!

PR #3027, opened 2024-06-05
downloads.py: delete stale incomplete downloads

mathiascode:
Delete incomplete download files that are no longer in the list.

### DIFF
--- a/pynicotine/downloads.py
+++ b/pynicotine/downloads.py
@@ -107,6 +107,8 @@ class Downloads(Transfers):
 
     def _quit(self):
 
+        self._delete_stale_incomplete_downloads()
+
         super()._quit()
 
         self._folder_basename_byte_limits.clear()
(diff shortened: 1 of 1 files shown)


slook:
I can envisage a scenario that could make the program very slow to quit while gathering `allowed_incomplete_file_paths`, so it would be better to tidy up as we go... This should be better handled in `_clear_transfer()`.

mathiascode:
> I can envisage a scenario that could make the program very slow to quit while gathering `allowed_incomplete_file_paths`

I restored the `current_byte_offset` check (forgot this is why I added it in the first place two months ago).

The reason I want to do this on shutdown is to give some room for error, if someone is halfway through a download and removes it by mistake, or wants to change the download folder path. In the future, we will hopefully have better tools/protections for that.

slook:
I agree it is convenient that re-adding an accidentally deleted incomplete download results in resuming it from existing data.

mathiascode:
Besides, we have to use this approach to remove incomplete files that were removed from the list long ago.

PR #3031, opened 2024-06-09
mainwindow.ui: show normal mouse cursor over empty area of status bar

slook:
* Fixed: Show normal mouse cursor over empty area of status bar, instead of incorrectly showing text selection beam cursor...

_Before:_
image

_After:_
image

... It is done by using using a GtkBox as a filler widget for doing the hexpand, for want of a better way?

### DIFF
--- a/pynicotine/gtkgui/ui/mainwindow.ui
+++ b/pynicotine/gtkgui/ui/mainwindow.ui
@@ -1682,14 +1682,19 @@
         <property name="spacing">18</property>
         <property name="visible">True</property>
         <child>
-          <object class="GtkLabel" id="status_label">
-            <property name="ellipsize">end</property>
+          <object class="GtkBox">
             <property name="hexpand">True</property>
-            <property name="margin-start">12</property>
-            <property name="selectable">True</property>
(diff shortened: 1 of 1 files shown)


mathiascode:
Placing the label inside the new box would make the most sense.

slook:
Yes, that's more like what I wanted.

pynicotine/gtkgui/ui/mainwindow.ui
```diff
@@ -1682,14 +1682,20 @@
...
             <property name="hexpand">True</property>
-            <property name="margin-start">12</property>
-            <property name="selectable">True</property>
-            <property name="single-line-mode">True</property>
             <property name="visible">True</property>
-            <property name="xalign">0</property>
+            <child>
+              <object class="GtkLabel" id="status_label">
+                <property name="ellipsize">end</property>
+                <property name="hexpand">False</property>
```
mathiascode, line 1691: You can remove this line, since hexpand is false by default.

slook: Of course it is. Thanks.

PR #3032, opened 2024-06-10
gtkgui/transfers.py: don't display current_byte_offset when finished

slook:
+ Changed: Avoid one `human_size()` call for each Finished transfer, it's also faster to read and more logical to view.

image

### DIFF
--- a/pynicotine/gtkgui/transfers.py
+++ b/pynicotine/gtkgui/transfers.py
@@ -473,7 +473,7 @@ class Transfers:
     @staticmethod
     def get_hsize(current_byte_offset, size):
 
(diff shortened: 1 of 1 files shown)


pynicotine/gtkgui/transfers.py
```diff
@@ -472,6 +472,10 @@ def get_hqueue_position(queue_position):
 
     @staticmethod
     def get_hsize(current_byte_offset, size):
+
+        if current_byte_offset == size:
```
mathiascode, line 476:
```suggestion
        if current_byte_offset >= size:
```

PR #3034, opened 2024-06-14
Unwatch users when updates are no longer needed

mathiascode:
Currently, the list of watched users grows indefinitely until we disconnect/quit. Unwatch users when we no longer need to track them.

### DIFF
--- a/pynicotine/buddies.py
+++ b/pynicotine/buddies.py
@@ -124,7 +124,7 @@ class Buddies:
             return
 
         for username in self.users:
-            core.users.watch_user(username)
+            core.users.watch_user(username, context="buddies")
 
     def _server_disconnect(self, _msg):
 
@@ -165,7 +165,7 @@ class Buddies:
         events.emit("add-buddy", username, user_data)
 
         # Request user status, speed and number of shared files
-        core.users.watch_user(username)
+        core.users.watch_user(username, context="buddies")
 
     def remove_buddy(self, username):
 
@@ -176,6 +176,7 @@ class Buddies:
             core.chatrooms.update_completions()
             core.privatechat.update_completions()
 
+        core.users.unwatch_user(username, context="buddies")
         self.save_buddy_list()
         events.emit("remove-buddy", username)
 
--- a/pynicotine/interests.py
+++ b/pynicotine/interests.py
@@ -22,16 +22,30 @@ from pynicotine.core import core
 from pynicotine.events import events
 
 
+class SimilarUser:
+    __slots__ = ("username", "rating")
+
+    def __init__(self, username, rating=None):
+        self.username = username
+        self.rating = rating
+
+
 class Interests:
 
     def __init__(self):
 
+        self.similar_users = {}
+
         for event_name, callback in (
-            ("item-similar-users", self._similar_users),
+            ("item-similar-users", self._item_similar_users),
+            ("quit", self._quit),
             ("server-login", self._server_login),
             ("similar-users", self._similar_users)
         ):
             events.connect(event_name, callback)
+
+    def _quit(self):
+        self.similar_users.clear()
 
     def _server_login(self, msg):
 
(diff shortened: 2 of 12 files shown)


pynicotine/interests.py
```diff
@@ -134,6 +144,15 @@ def request_similar_users(self):
...
         """Server code 110 and 112."""
 
+        for username in self.similar_users:
+            if username not in msg.users:
+                core.users.unwatch_user(username, mode="interests")
+
+        self.similar_users.clear()
+
         for username in msg.users:
+            rating = msg.users[username] if isinstance(msg.users, dict) else None
```
slook, line 154: `msg.users` should always be a `dict` even if it is empty, so `isinstance()` shouldn't be required.

pynicotine/users.py
```diff
@@ -28,14 +28,16 @@
...
-    __slots__ = ("username", "upload_speed", "files", "folders")
+    __slots__ = ("username", "upload_speed", "files", "folders", "modes", "implicit")
 
     def __init__(self, username):
 
         self.username = username
         self.upload_speed = None
         self.files = None
         self.folders = None
+        self.modes = set()
```
slook, line 39: String matching to identify different modes isn't very nice.

Consider defining a new class such as `WatchMode()` for this purpose.

pynicotine/users.py
```diff
@@ -28,14 +28,16 @@
...
+        self.implicit = True  # True if no WatchUser message was sent (i.e. receiving updates for room user)
```
slook, line 40: Consider a `set()` to contain all the `rooms` that a user is in, avoids doing string concatenations.

If the set is empty the user would not be in any of the rooms we are in.

pynicotine/buddies.py
```diff
@@ -165,7 +165,7 @@ def add_buddy(self, username):
         events.emit("add-buddy", username, user_data)
 
         # Request user status, speed and number of shared files
-        core.users.watch_user(username)
+        core.users.watch_user(username, mode="userlist")
```
slook, line 168: Ambiguity with normal user. "userlist" -> "buddy"

slook (changes requested):

pynicotine/interests.py
mathiascode: It can be a list or dict depending on which server message it is (one doesn't include ratings). I'll probably come up with something better.

pynicotine/users.py
mathiascode: It doesn't feel right to make the users module responsible for knowing every mode in the application. Ideally you could call watch_user() from anywhere (say, a plugin) without being required to add a new mode here.

pynicotine/users.py
mathiascode: Similar concerns as the previous comment. I don't think the users module should know this.

mathiascode: I'll get rid of `mode.startswith("chatroom_")` in favor of a boolean parameter for watch_user().

pynicotine/users.py
slook: Perhaps get a value without needing to pass an argument with something in https://docs.python.org/3/library/inspect.html#module-inspect

pynicotine/buddies.py
slook: Rather than "buddy", actually "buddies" would match what is used in `core.enabled_components`.

mathiascode:
I renamed mode -> context, seemed a little clearer to me.

Also refactored transfer user watching so it doesn't unwatch/rewatch when changing the state of the last remaining transfer.

pynicotine/users.py
```diff
@@ -116,20 +118,86 @@ def request_set_status(self, status):
...
             return
 
-        if username in self.watched:
+        watched_user = self.watched.get(username)
+
+        if watched_user is None:
+            self.watched[username] = watched_user = WatchedUser(username)
+
+        if not context:
+            context = "unknown"
```
slook, line 142: Why is this needed?

If the function call is made improperly, then an error message should be logged.

pynicotine/users.py
```diff
@@ -116,20 +118,86 @@ def request_set_status(self, status):
...
+            # We can't unwatch ourselves
+            return
+
+        watched_user = self.watched.get(username)
+
+        if watched_user is None:
+            return
+
+        if not context:
+            context = "unknown"
```
slook, line 176: If the context is unspecified, then perhaps all contexts ought to be discarded?

pynicotine/users.py
mathiascode: I'd like to get the change into 3.3.x, but wasn't sure if any plugins are calling this function. I'll just mark watch_user() with no context as deprecated, and make the context mandatory in 3.4.0.

pynicotine/users.py
mathiascode: Too easy to misuse and break things. I'll make the context mandatory, since it's a new function.

Issue #3039, opened 2024-06-21
nowplaying - other should decode bytes


mayli:
Nicotine+ version:  latest
Operating System/Distribution:

### Describe the bug
nowplaying is returning b'xxxxx' when after executing external commands

### Expected behavior
show output without the `b''`

### Steps to reproduce the bug
use custom command in the nowplaying settings, click test

### Additional context
Screenshots, logs, stacktraces or relevant information.

https://github.com/nicotine-plus/nicotine-plus/blob/master/pynicotine/utils.py#L476
Is returing bytes by default, we need to add a `.decode('utf-8')` there.

slook:
Please create a PR (pull request) with the exact edit of the code required to fix the issue.

PR #3040, opened 2024-06-21
nowplaying: decode command output

mayli:


slook:
Fixes #3039

pynicotine/nowplaying.py
```diff
@@ -300,7 +300,7 @@ def other(self, command):
             return None
 
         try:
-            output = execute_command(command, returnoutput=True, hidden=True)
+            output = execute_command(command, returnoutput=True, hidden=True).decode()
             self.title["nowplaying"] = output
```
slook, line 304: Can we have

`            self.title["nowplaying"] = output.decode()`

That is the same effect but easier to read.

slook (changes requested):

mayli (commented):
address comment

pynicotine/nowplaying.py
mayli: Done.

mathiascode (approved):

PR #3041, opened 2024-06-22
treeview.py: remove legacy verify_grouping_mode() function

slook:
- Removed: Legacy function in favor of permanent config migration.

### DIFF
--- a/pynicotine/config.py
+++ b/pynicotine/config.py
@@ -703,11 +703,11 @@ class Config:
                 self.sections[section][option] = value
 
         # Map legacy folder/user grouping modes (3.1.0)
-        for (section, option) in [
+        for section, option in (
             ("searches", "group_searches"),
             ("transfers", "groupdownloads"),
             ("transfers", "groupuploads")
-        ]:
+        ):
             mode = self.sections[section].get(option, "folder_grouping")
 
             if mode == "0":
--- a/pynicotine/config.py
+++ b/pynicotine/config.py
@@ -702,6 +702,26 @@ class Config:
                 # Set default value
                 self.sections[section][option] = value
 
+        # Map legacy folder/user grouping modes (3.1.0)
+        for (section, option) in [
+            ("searches", "group_searches"),
+            ("transfers", "groupdownloads"),
+            ("transfers", "groupuploads")
+        ]:
+            mode = self.sections[section].get(option, "folder_grouping")
+
+            if mode == "0":
+                mode = "ungrouped"
+
+            elif mode == "1":
+                mode = "folder_grouping"
+
+            elif mode == "2":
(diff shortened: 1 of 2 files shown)


pynicotine/config.py
```diff
@@ -699,6 +699,30 @@ def _set_config(self):
...
+
+                    elif mode == "2":
+                        mode = "user_grouping"
+
+                    # Verify mode validity
+                    elif mode not in {"ungrouped", "folder_grouping", "user_grouping"}:
+                        mode = "folder_grouping"
+
+                    self.sections[section][option] = mode
+                    continue
```
mathiascode, line 724: This code block isn't currently reached. It should be moved down after line 728 (change `if (section, option) in [...]` to `for section, option in (...)`).

pynicotine/config.py
```diff
@@ -699,6 +699,30 @@ def _set_config(self):
...
+
+                    elif mode == "1":
+                        mode = "folder_grouping"
+
+                    elif mode == "2":
+                        mode = "user_grouping"
+
+                    # Verify mode validity
+                    elif mode not in {"ungrouped", "folder_grouping", "user_grouping"}:
+                        mode = "folder_grouping"
```
mathiascode, line 721: We can remove this code. We should only take care of migrating old config values.

mathiascode (changes requested):

pynicotine/config.py
slook: Oh dear, the method has become too complex. Perhaps we need a dedicated `_migrate_config()` function, or such?

mathiascode: Most likely, but don't include it in this PR.

pynicotine/config.py
```diff
@@ -702,6 +702,26 @@ def _set_config(self):
                 # Set default value
                 self.sections[section][option] = value
 
+        # Map legacy folder/user grouping modes (3.1.0)
+        for (section, option) in [
+            ("searches", "group_searches"),
+            ("transfers", "groupdownloads"),
+            ("transfers", "groupuploads")
+        ]:
```
mathiascode, line 710:
```suggestion
        for section, option in (
            ("searches", "group_searches"),
            ("transfers", "groupdownloads"),
            ("transfers", "groupuploads")
        ):
```

Small nit. This is the code style we use elsewhere nowadays.

pynicotine/config.py
mathiascode: Looks like the linter agrees too

mathiascode (approved):

pynicotine/config.py
mathiascode: Done in 778e2748d8ace661c2549eaf973a985f1d65162b

PR #3047, opened 2024-06-28
ChatView: don't timestamp old messages separator in append_log_lines()

slook:
Minor UI change:

+ Changed: Don't timestamp the `--- old messages above ---` separator line, because the complexity required in the code it is not really worthwhile for this purpose, actually it is visually easier to spot it with no timestamp.

Minor code cleanups:

+ Changed: the lowercase `login_username` string is required for each line in lines, so it needs doing only once beforehand
+ Changed: Call the `read_log()` method directly from gui module instead of parsing config values into the widget
+ Removed: no caller gets the value that was returned by append_line()

This structure of methods is part of what will be required for future enhancements that are planned in draft PR #2792 (please try running that branch and let me know what you think of the overall idea), but this PR is ready to be merged, if desired.

### DIFF
--- a/pynicotine/gtkgui/chatrooms.py
+++ b/pynicotine/gtkgui/chatrooms.py
@@ -660,12 +660,13 @@ class ChatRoom:
 
     def prepend_old_messages(self):
 
-        log_lines = self._read_old_messages()
+        log_lines = log.read_log(
+            folder_path=log.room_folder_path,
+            basename=self.room,
+            num_lines=config.sections["logging"]["readroomlines"]
+        )
 
         self.chat_view.append_log_lines(log_lines, login_username=config.sections["server"]["login"])
-
-    def _read_old_messages(self):
-        return log.read_log(log.room_folder_path, self.room, config.sections["logging"]["readroomlines"])
 
     def populate_room_users(self, users):
 
--- a/pynicotine/gtkgui/chatrooms.py
+++ b/pynicotine/gtkgui/chatrooms.py
@@ -569,7 +569,7 @@ class ChatRoom:
         )
 
         self.setup_public_feed()
-        self.read_room_logs()
+        self.prepend_old_messages()
 
     def load(self):
         GLib.idle_add(self.read_room_logs_finished)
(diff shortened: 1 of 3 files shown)


pynicotine/gtkgui/chatrooms.py
```diff
@@ -658,14 +658,14 @@ def read_room_logs_finished(self):
...
-            folder_path=log.room_folder_path,
-            basename=self.room,
-            num_lines=config.sections["logging"]["readroomlines"],
-            timestamp_format=config.sections["logging"]["rooms_timestamp"]
-        )
+        log_lines = self._read_old_messages()
+
+        self.chat_view.append_log_lines(log_lines, login_username=config.sections["server"]["login"])
+
+    def _read_old_messages(self):
```
mathiascode, line 667: Would probably make more sense to get rid of this function for now and just call log.read_log() directly?

pynicotine/gtkgui/chatrooms.py
```diff
@@ -658,14 +658,14 @@ def read_room_logs_finished(self):
...
+        return log.read_log(log.room_folder_path, self.room, config.sections["logging"]["readroomlines"])
```
mathiascode, line 668: Could you pass the arguments as keyword arguments instead? Would be easier to see what they do at a glance.

pynicotine/gtkgui/privatechat.py
```diff
@@ -394,14 +394,14 @@ def read_private_log_finished(self):
...
-            basename=self.user,
-            num_lines=config.sections["logging"]["readprivatelines"],
-            timestamp_format=config.sections["logging"]["private_timestamp"]
-        )
+        log_lines = self._read_old_messages()
+
+        self.chat_view.append_log_lines(log_lines, login_username=config.sections["server"]["login"])
+
+    def _read_old_messages(self):
+        return log.read_log(log.private_chat_folder_path, self.user, config.sections["logging"]["readprivatelines"])
```
mathiascode, line 404: See previous comments.

pynicotine/gtkgui/widgets/textview.py
```diff
@@ -49,7 +47,6 @@ class TextView:
 
     except TypeError:
         # Broken cursor theme, but what can we do...
-        log.add_debug("Cannot load cursors from theme, falling back to default cursor")
```
mathiascode, line 52: You suggested this log message originally.

pynicotine/gtkgui/widgets/textview.py
```diff
@@ -405,24 +400,19 @@ def __init__(self, *args, chat_entry=None, status_users=None, username_event=Non
...
-
-        lines = log.read_log(folder_path, basename, num_lines)
+    def append_log_lines(self, log_lines, login_username=None):
 
-        if not lines:
+        if not log_lines:
             return
 
-        login = config.sections["server"]["login"]
+        login_lower = login_username.lower() if login_username else None
```
mathiascode, line 408: login_username_lower

pynicotine/gtkgui/chatrooms.py
slook: Yes, it's not required until textview is made capable of prepending lines as in PR #2792. In there, I found that a seg fault may occur when rapidly infinite scrolling (such as when holding down the PageUp key) unless GLib.idle_add() was used, which doesn't accept multiple arguments hence why this separate function call was implemented.

pynicotine/gtkgui/widgets/textview.py
slook: We have finished debugging that issue now, so it is no longer required (and so, the log import is not needed anymore for this is the only call to it).

Iirc our conclusion was that some cursor themes use different naming schemes, and indeed there is nothing we can do about that.

PR #3049, opened 2024-06-30
uploads: refactor code related to pending shutdown

mathiascode:
The previous code was bad/sloppy for many reasons:
- Core should not be aware of uploads, only the uploads module should
- No events for starting/stopping a pending shutdown for uploads
- Implemented in the wrong module
- Gave me a headache when reading it

Based on previous feedback from users, also change the behavior to always show a confirmation dialog on quit if uploads are active (unless the window is hidden/running in the background).

### DIFF
--- a/pynicotine/events.py
+++ b/pynicotine/events.py
@@ -182,6 +182,8 @@ EVENT_NAMES = {
     "upload-denied",
     "upload-failed",
     "upload-file-error",
+    "uploads-shutdown-request",
+    "uploads-shutdown-cancel",
 
     # User info
     "user-info-progress",
--- a/pynicotine/gtkgui/transfers.py
+++ b/pynicotine/gtkgui/transfers.py
@@ -102,7 +102,7 @@ class Transfers:
     UNKNOWN_STATUS_PRIORITY = 1000
 
     path_separator = path_label = retry_label = abort_label = None
-    transfer_page = user_counter = file_counter = expand_button = expand_icon = grouping_button = None
+    transfer_page = user_counter = file_counter = expand_button = expand_icon = grouping_button = status_label = None
 
     def __init__(self, window, transfer_type):
 
@@ -365,6 +365,16 @@ class Transfers:
             return translated_status
 
         return status
+
+    def update_limits(self):
+        """Underline status bar bandwidth labels when alternative speed limits
+        are active."""
+
+        if config.sections["transfers"][f"use_{self.type}_speed_limit"] == "alternative":
+            add_css_class(self.status_label, "underline")
+            return
+
+        remove_css_class(self.status_label, "underline")
 
     def update_num_users_files(self):
         self.user_counter.set_text(humanize(len(self.users)))
--- a/pynicotine/gtkgui/uploads.py
+++ b/pynicotine/gtkgui/uploads.py
@@ -24,6 +24,8 @@
 
 import os
 
+from gi.repository import Gtk
+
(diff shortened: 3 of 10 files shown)


pynicotine/events.py
```diff
@@ -182,6 +182,8 @@
     "upload-denied",
     "upload-failed",
     "upload-file-error",
+    "uploads-start-shutdown",
+    "uploads-stop-shutdown",
```
slook, line 186: This contradiction in terms is rather odd. Suggest:

`"uploads-shutdown-requested"`
`"uploads-shutdown-cancelled"`

or simply (the event doesn't necessarily have to be exclusive to the uploads module):

`"shutdown-request"` (...`ed`)
`"shutdown-cancel"` (...`led`)

pynicotine/gtkgui/uploads.py
```diff
@@ -99,6 +105,22 @@ def destroy(self):
         self.upload_speeds.destroy()
         super().destroy()
 
+    def start_shutdown(self):
```
slook, line 108: `shutdown_requested()` or `request_shutdown()`

pynicotine/gtkgui/uploads.py
```diff
@@ -99,6 +105,22 @@ def destroy(self):
...
+        toggle_status_action = self.window.lookup_action("toggle-status")
+
+        toggle_status_action.handler_block_by_func(self.window.on_toggle_status)
+        self.window.user_status_button.set_active(True)
+        toggle_status_action.handler_unblock_by_func(self.window.on_toggle_status)
+
+        self.window.user_status_icon.set_from_icon_name(icon_name, *icon_args)
+        self.window.user_status_label.set_text(_("Quitting..."))
+
+    def stop_shutdown(self):
```
slook, line 121: `shutdown_cancelled()` or `cancel_shutdown()`

pynicotine/uploads.py
```diff
@@ -706,6 +697,25 @@ def clear_uploads(self, uploads=None, statuses=None, denied_message=None):
...
+
+        if self.pending_shutdown:
+            return
+
+        self.pending_shutdown = True
+        self._check_upload_queue()
+
+        events.emit("uploads-start-shutdown")
+
+    def stop_shutdown(self):
```
slook, line 711: `shutdown_cancelled()` or `cancel_shutdown()`

slook (changes requested):

pynicotine/gtkgui/transfers.py
```diff
@@ -102,7 +102,7 @@ class Transfers:
     UNKNOWN_STATUS_PRIORITY = 1000
 
     path_separator = path_label = retry_label = abort_label = None
-    transfer_page = user_counter = file_counter = expand_button = expand_icon = grouping_button = None
+    transfer_page = user_counter = file_counter = expand_button = expand_icon = grouping_button = status_label = None
```
slook, line 105: If we have a reference to the status_label widget in this module, then `set_connection_stats()` can be moved in here too.

pynicotine/events.py
mathiascode: `uploads-shutdown-request` and `uploads-shutdown-cancel` perhaps? I want to make it clear that the request came from the uploads module, so we can react accordingly in the GUI. We don't want to cause confusion between this event and the existing `schedule-quit`.

slook: Yes, go for that.

PR #3053, opened 2024-07-04
Improvements to IP country data format

mathiascode:
A few months ago, I introduced a new IP country data format using Python files. While the intention was to find a balance between file size, memory usage and performance, I failed to realize a few things:

- Generating the bytecode (.pyc) files uses 300+ MB memory momentarily. We don't want any sudden spikes in memory usage.
- The .pyc files are much larger than the .py files, ~8 MB vs ~3 MB. Some packages, such as the Arch package, ship multiple .pyc files for different optimization levels, resulting in ~16 MB of wasted space.
- Importing the data modules the first time a .pyc file is generated is quite slow.

This commit tries a different approach, using text files the same size as the old .py files, without the need for bytecode files. This fulfills the requirement of keeping the data files readable, which was the primary motivation for changing the format in the first place.


slook:
This file format is more like `.csv` than `.txt`.

Is there still a compelling reason to have the data split across two files, or would it not be better to have them on two rows within the same file instead?

mathiascode:
> Is there still a compelling reason to have the data split across two files, or would it not be better to have them on two rows within the same file instead?

No good reason, really. I've combined the files into one.

data/scripts/update_ip_country_data.py
```diff
@@ -76,64 +76,44 @@ def parse_ip_country_data():
...
-                country_values.append(country_code)
+                ip_range_values.append(address_to)
+                ip_range_countries.append(country_code)
 
-    return country_codes, ip_ranges, country_values, copyright_notice
+    return ip_range_values, ip_range_countries, copyright_notice
 
 
 def update_ip_country_data():
     """Update country data files."""
```
slook, line 86: files -> file

slook (approved):

PR #3054, opened 2024-07-05
config.py: remove 'maxresults' option

mathiascode:
Someone asked what the purpose of this option is, and I didn't have a good answer. It's another one of those options that have been around forever, and nobody bothered looking into.

Hard code the maximum number of returned results to 500 instead. This should leave room for useful results in almost every case, but also avoid wasting bandwidth by returning thousands of results for popular search terms.

### DIFF
--- a/pynicotine/config.py
+++ b/pynicotine/config.py
@@ -291,7 +291,6 @@ class Config:
             "searches": {
                 "expand_searches": True,
                 "group_searches": "folder_grouping",
-                "maxresults": 150,
                 "enable_history": True,
                 "history": [],
                 "enablefilters": False,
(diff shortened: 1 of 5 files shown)


mathiascode:
@slook Why the thumbs down?

slook:
Primarily because I'm not in favour of hard-coded parameters.

I do agree that this option is not really of much importance to the end user, but it is an important parameter for the functioning of the client that strongly influences the pattern of behaviour on the network while distributing search results, so therefore it should remain configurable.

mathiascode:
Bumped the default to 300: df05a68c3831e1085eeabb47310a1eda73f9f32f
Lowered the limit in the preferences dialog a bit: 5509e1470f03709ce9ff16656a91f761275baf70

PR #3070, opened 2024-08-03
settings/search.ui: add revealer for default filters

slook:
+ Added: Revealer for default filter entries, this saves vertical space and better guides how the function is activated.
+ Changed: "Free Slot" checkbox to "Free Slot:" switch (involves a string change).

_Collapsed:_
image

_Revealed:_
image


mathiascode:
I like the direction this is going in. What do you think about something like this?
expanded
collapsed

slook:
Yes that is even better :) Please go ahead and commit it.

mathiascode:
Thanks!

PR #3118, opened 2024-09-07
TextSearchBar: get selection content in text entry on Find

slook:
+ Added: Populate the Find text entry with selected text on initializing new text search in chat, private chat or log window

Not tested in GTK 4 which has `get_selection_content()` but `get_selection_bounds()` should work in both GTK 3 and GTK 4.

~If multiple lines are selected then just show the Find bar with a blank entry as before.~

It would be nicer to have a `get_selection_content()` helper function in `textview.py` but the Class isn't available from `textedit.py` only the widget can be referenced from there with the way it is at the moment.

### DIFF
--- a/pynicotine/gtkgui/widgets/textentry.py
+++ b/pynicotine/gtkgui/widgets/textentry.py
@@ -630,12 +630,24 @@ class TextSearchBar:
 
     def set_visible(self, visible):
 
+        self.search_bar.set_search_mode(False)
+
         if visible:
+            text_buffer = self.textview.get_buffer()
+            text_buffer_selection_bounds = text_buffer.get_selection_bounds()
+
+            if not self.entry.get_text() and text_buffer_selection_bounds:
+                selection_start, selection_end = text_buffer_selection_bounds
+
+                selection_content = text_buffer.get_text(
+                    selection_start, selection_end, include_hidden_chars=True).strip(" \n")
+
+                if "\n" not in selection_content:
+                    self.entry.set_text(selection_content)
+
             self.search_bar.set_search_mode(True)
             self.entry.grab_focus()
             return
-
-        self.search_bar.set_search_mode(False)
 
         if self.focus_widget is not None and self.focus_widget.get_sensitive():
             self.focus_widget.grab_focus()
--- a/pynicotine/gtkgui/widgets/textentry.py
+++ b/pynicotine/gtkgui/widgets/textentry.py
@@ -630,23 +630,22 @@ class TextSearchBar:
 
     def set_visible(self, visible):
 
-        self.search_bar.set_search_mode(False)
+        self.search_bar.set_search_mode(visible)
 
         if visible:
             text_buffer = self.textview.get_buffer()
-            text_buffer_selection_bounds = text_buffer.get_selection_bounds()
+            selection_bounds = text_buffer.get_selection_bounds()
 
-            if not self.entry.get_text() and text_buffer_selection_bounds:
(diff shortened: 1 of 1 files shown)


mathiascode:
Preferably the text would remain selected in the entry, but here's an issue where the text is selected briefly before the caret is moved to the end.

slook:
It wasn't intended, but imo it looks good when the text is briefly selected.

pynicotine/gtkgui/widgets/textentry.py
```diff
@@ -629,13 +629,25 @@ def on_show_search_accelerator(self, *_args):
...
+
+                selection_content = text_buffer.get_text(
+                    selection_start, selection_end, include_hidden_chars=True).strip(" \n")
+
+                if "\n" not in selection_content:
+                    self.entry.set_text(selection_content)
+
             self.search_bar.set_search_mode(True)
             self.entry.grab_focus()
             return
```
mathiascode, line 649: Having used some GTK-based text editors a lot, these are a couple things I expect:
- Selected text is inserted into the search entry, no matter how many lines
- Selecting a different piece of text in the text view and pressing Ctrl+F again starts a new search
- Cursor is placed at the end of search entry (ideally the text would be selected to allow typing to overwrite it, but that's not possible with the current implementation)

```suggestion
        self.search_bar.set_search_mode(visible)

        if visible:
            text_buffer = self.textview.get_buffer()
            selection_bounds = text_buffer.get_selection_bounds()

            if selection_bounds:
                selection_start, selection_end = selection_bounds
                selection_content = text_buffer.get_text(
                    selection_start, selection_end, include_hidden_chars=True)

                if self.entry.get_text().lower() != selection_content.lower():
                    self.entry.set_text(selection_content)

            self.entry.grab_focus()
            self.entry.set_position(-1)
            return
```

slook:
> limitation due to the way we implement search result highlights

We don't really take much advantage of the features offered by the widget. It would be better to insert marks at the bounds of all the ranges and then tag the matches with a highlighted style to visually identify all the results at once, with a focus rectangle to show the current match, instead of just making a regular selection.

> Having used some GTK-based text editors a lot

You might be familiar with [Scintilla](https://www.scintilla.org/) which is what [Geany](https://www.geany.org/) uses for its editor widget. It can be easily configured via a standard Addon to "Mark all occurances of a word when double-clicking it" which is very fast and convenient because it doing it that way helps to avoid any need to enter text into the Find toolbar or dialog box...

image

mathiascode:
> > limitation due to the way we implement search result highlights
> 
> We don't really take much advantage of the features offered by the widget. It would be better to insert [TextMark](https://docs.gtk.org/gtk3/class.TextMark.html)s at the bounds of all the ranges and then tag the matches with a highlighted style to visually identify all the results at once, with a focus rectangle to show the current match, instead of just making a regular selection.

Definitely, that would be the way to go in the future.

> You might be familiar with [Scintilla](https://www.scintilla.org/) which is what [Geany](https://www.geany.org/) uses for its editor widget. It can be easily configured via a bundled Addon called [Automark](https://github.com/geany/geany-plugins/tree/master/automark) to "Mark all occurances of a word when double-clicking it" which is very fast and convenient because it doing it that way helps to avoid any need to enter text into the Find toolbar or dialog box...

I haven't used Automark, but it sounds like something that shouldn't be too difficult to implement.

slook:
I was mistaken its not "Automark" but rather "Markword" (this one is included as standard with the base install of Geany), link:
https://github.com/geany/geany-plugins/blob/master/addons/src/ao_markword.c

PR #3121, opened 2024-09-09
application.py: check for auto-connect on headless startup

monowii:


mathiascode:
That option was ignored intentionally, since headless mode is basically useless without connecting. Is there a good reason you want to check the option?

monowii:
when developing plugins, I found out this was helpful when restarting the client many times

slook:
> That option was ignored intentionallyy, since headless mode is basically useless without connecting.

Iirc, the headless implementation was before the `connect` command, without which it would have been impossible to connect. 

Obviously the `connect` command does exist now, so the proposed change in this PR is appropriate.

> when developing plugins, I found out this was helpful when restarting the client many times

It isn't necessary to restart the client, since the `plugin reload <plugin_name>` command has recently been added for this purpose.

mathiascode:
I think we need some kind of log message in headless mode when this option is disabled, e.g. "Type /connect to log in". The only indicator that you're connected is the connection-related log messages, which people might not know or forget about.

slook:
Perhaps fire `core.disconnect` which logs the "Disconnected from server" message.

mathiascode:
Unfortunately it's not that simple. You have to be connected to the server for that log message to appear.

I think I'll merge this as-is, then we can figure out something better in 3.4.0.

PR #3153, opened 2024-10-11
downloads.py: only remove actual stale incomplete downloads

### LINKED ISSUE #3152: Files deleted on quit
Nicotine+ version:  3.3.5
Operating System/Distribution: Garuda GNU/Linux

So, I recently lost *all* of the files stored in the root of my downloads folder. I'm using the same folder for finished downloads, incomplete downloads, and received files, and it is also shared. To test this, I placed an empty file in the root, shut down Nicotine+, and it disappeared. Then, I renamed my settings folder to set everything back to defaults, and sure enough, the empty file is still there when I shut down Nicotine+.

This problem only cropped up when I reinstalled my system for some reason. I'm still using the same settings I've used for years with no problem.

mathiascode:
If the user's incomplete download folder contains other unrelated files, ensure we don't delete them.

Fixes #3152

### DIFF
--- a/pynicotine/downloads.py
+++ b/pynicotine/downloads.py
@@ -525,12 +525,21 @@ class Downloads(Transfers):
         if not self._allow_saving_transfers:
             return
 
+        incomplete_download_folder_path = self.get_incomplete_download_folder()
+
+        if not incomplete_download_folder_path.startswith(config.data_folder_path):
+            # Only delete incomplete downloads inside Nicotine+'s data folder
+            return
+
         allowed_incomplete_file_paths = {
             encode_path(self.get_incomplete_download_file_path(transfer.username, transfer.virtual_path))
             for transfer in self.transfers.values()
             if transfer.current_byte_offset and transfer.status != TransferStatus.FINISHED
         }
-        incomplete_download_folder_path = self.get_incomplete_download_folder()
+        prefix = b"INCOMPLETE"
+        prefix_len = len(prefix)
+        md5_len = 32
+        md5_regex = re.compile(b"[0-9a-f]{32}", re.IGNORECASE)
 
         try:
             with os.scandir(encode_path(incomplete_download_folder_path)) as entries:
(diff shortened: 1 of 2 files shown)


pynicotine/downloads.py
```diff
@@ -531,6 +531,10 @@ def _delete_stale_incomplete_downloads(self):
             if transfer.current_byte_offset and transfer.status != TransferStatus.FINISHED
         }
         incomplete_download_folder_path = self.get_incomplete_download_folder()
```
slook, line 533: `if self.get_default_download_folder() == incomplete_download_folder_path:` then there is a possibility that the contents of the folder might be huge. In that case, all the shenanigans in this method ought to be avoided entirely.

slook (changes requested):

pynicotine/downloads.py
mathiascode: Maybe better to check if we're in the Nicotine+ data folder, and only perform a cleanup in this case? We still want to guarantee that no junk remains in Nicotine+'s "internal" storage, but we'd avoid deleting anything outside of it.

slook: That gives different behaviour depending upon what the paths are set to, which makes it difficult to predict what the program will do on any given instance, whereas it shouldn't matter where the incomplete folder is set to.

Perhaps it would be better to delete only those incomplete files which have been created/accessed during a given session, that would involve keeping a transient list of explicit path/filenames which are destined for deletion, rather than just arbitrarily wiping the entire contents of a folder...

... Really, in hindsight an indiscriminate cleanup operation like this ought only be manually triggered by the user. There is also no provision nor possibility of securely erasing the temporary data, which might be something users need.

slook: If we can't decide what the best way is, then I suggest either merging what you have so far or removing the feature entirely for now in order to expedite a patch release.

mathiascode: i'll merge as-is, we can think more about the feature when there's more time.

PR #3163, opened 2024-10-19
Strip trailing newline characters from password entries

### LINKED ISSUE #3156: Incorrect processing of passwords with special characters
Nicotine+ version:  3.2.9
Operating System/Distribution: Ubuntu 24.04.1 LTS

### Describe the bug

When application is started for the first time, one enters username and password. All works fine during the first run. 
At the second run (after restarting the application), the application fails to read the password from configs properly, prompting a window to enter "correct" credentials again.

Come up with a username and use `@dRtif8r%?8Ea8O!3jW%"p\#5` as the password.

Most likely the application fails to read `#` (interpreting it as a comment):

```
$ head ~/.config/nicotine/config
[server]
server = ('server.slsknet.org', 2242)
login = someusername
passw = @dRtif8r%?8Ea8O!3jW%"p\#5
interface = 
ctcpmsgs = False
autosearch = []
autoreply = 
portrange = (2234, 2239)
upnp = True
```

slook:
Fixes #3156

+ Changed: Disallow end-trailing newline characters from the end of password string inputs

Even though the server allows such strings, the config file is unable to. Assume it isn't the user's intention to include a newline character at the end of their password, this may occur unintentionally when pasting from a password manager or document.

A multi-line password (containing newline characters within the string) is still allowed, since this seems to be accepted by both the server and the config file.


slook:
I updated this PR to truncate-multiline rather than only stripping the trailing newline.

PR #3166, opened 2024-10-24
transfers.py: reset last_byte_offset when aborting transfer

### LINKED ISSUE #3162: 3.3.5 has same bug latest Test clients have about upload showing wrong upload
Nicotine+ version:  3.3.5
Operating System/Distribution: Windows

I went from latest test clients to Stable to avoid this bug but after going to newer builds it started happening again, possibly related to "Upload I/O error: Remote client does not support large file transfers: cannot fit 'int' into an offset-sized integer" bug

but my total upload data in Statistics went from some normal TB value to 60 EiB however much that is, i really don't want this bug to be messing up my upload stats

i don't see it fixed in new 3.3.6 release notes

### Additional context
Untitled

mathiascode:
Fixes #3162

### DIFF
--- a/pynicotine/transfers.py
+++ b/pynicotine/transfers.py
@@ -402,6 +402,10 @@ class Transfers:
         transfer.legacy_attempt = False
         transfer.size_changed = False
 
+        # Reset last byte offset to avoid incorrect offset subtractions between
+        # previous and new transfer sessions when updating statistics
+        transfer.last_byte_offset = None
+
         if transfer.sock is not None:
             core.send_message_to_network_thread(CloseConnection(transfer.sock))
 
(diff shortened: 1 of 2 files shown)


mathiascode:
Shouldn't affect anything else, but I'd appreciate testing to verify that I didn't break anything.

slook:
Well, I've done a couple transfers in each direction, paused, aborted, resumed a bunch of times, and all the copies are bit perfect and the stats are correct.

However, it's impossible for us to test for the all things that could actually go wrong! Is the CI test capable of adding a transfer to self and at least compare that the sizes are equal?

slook (approved):
Well, I've done a couple transfers in each direction, paused, aborted, resumed a bunch of times, and all the copies are bit perfect and the stats are correct.

However, it's impossible to test for the all unknown things that could actually go wrong!

PR #3186, opened 2024-11-12
Reconnect when changing listening port or network interface

### LINKED ISSUE #2704: Allow N+ to update Port without restarting
One should be able to change the Port, hit Apply, and the change will automatically take effect without having to close and restart the program.

### Additional context
MANY other programs (ex, qBittorrent, uTorrent, etc) have this functionality.

mathiascode:
Closes #2704 

Two translatable strings are modified, but since the only change is removing the `(requires a restart)` part from them, we could modify the translated strings manually and include this PR in Nicotine+ 3.3.7.

### DIFF
--- a/pynicotine/core.py
+++ b/pynicotine/core.py
@@ -92,7 +92,7 @@ class Core:
 
         for event_name, callback in (
             ("quit", self._quit),
-            ("server-reconnect", self.connect)
+            ("server-reconnect", self._server_reconnect)
         ):
             events.connect(event_name, callback)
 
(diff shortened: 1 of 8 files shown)


minicoz:
This would be awesome to get, i always have to restart the docker container when I update the listening port due to my vpn changing port. It would be great if I could just disconnect and reconnect.

mathiascode:
Reconnecting is enough in Nicotine+ 3.3.x (Menu -> Disconnect, then Menu -> Connect), but the "requires restart" description is a leftover from older versions, since we didn't have a good replacement for it. We'll just auto-reconnect when changing the port now to make things easier.

slook:
Automatically reconnecting would also make sense when changing username and/or server, as well.

slook (approved):

slook:
@minicoz It is in 3.3.7rc1 now you can try it https://github.com/nicotine-plus/nicotine-plus/blob/master/doc/TESTING.md

minicoz:
@slook thanks, but i think im getting stuck since im using the requests library that is not installed in the nicotine runtime. 

I have to call the control place local api endpoint and im trying to figure out how to do that since I keep getting 
`ModuleNotFoundError: No module named 'requests'`

PR #3217, opened 2024-12-11
Show folder count while rescanning shares

mathiascode:
We should give users a sign that the scanner is actually doing something.

The strings are a bit of a mess currently, since we can't add new translatable strings in 3.3.x, but we can improve the situation in 3.4.x.

### DIFF
--- a/pynicotine/gtkgui/mainwindow.py
+++ b/pynicotine/gtkgui/mainwindow.py
@@ -59,6 +59,7 @@ from pynicotine.gtkgui.widgets.theme import set_use_header_bar
 from pynicotine.gtkgui.widgets.window import Window
 from pynicotine.logfacility import log
 from pynicotine.slskmessages import UserStatus
+from pynicotine.utils import humanize
 from pynicotine.utils import open_folder_path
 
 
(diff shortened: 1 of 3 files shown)


pynicotine/shares.py
```diff
@@ -262,6 +262,7 @@ def __init__(self, queue, share_groups, share_db_paths, init=False, rescan=True,
         self.processed_share_names = set()
         self.processed_share_paths = set()
         self.current_file_index = 0
+        self.current_folder_count = 0
```
slook, line 265: Local scope only inside `scan_shared_folder()` like `folder_paths`

slook (changes requested):

pynicotine/shares.py
mathiascode: That will reset the count when we scan the next top-level folder.

pynicotine/gtkgui/mainwindow.py
```diff
@@ -1149,17 +1154,30 @@ def set_connection_stats(self, total_conns=0, **_kwargs):
...
         self.scan_progress_spinner.start()
 
-    def shares_scanning(self):
+    def shares_scanning(self, folder_count=None):
+
+        label = _("Scanning Shares")
+
+        if folder_count is not None:
+            self.scan_progress_label.set_label(
+                f"{_('Scanning Shares')} ({humanize(folder_count)})")
```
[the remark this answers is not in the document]
mathiascode: Humanize shouldn't be an issue here.

PR #3268, opened 2025-02-04
logfacility.py: Optimize reading of log lines

mathiascode:
Turns out using deque for reading lines from a large file is really slow, even when you only want to show a fraction of the contents in the GUI. Use a more efficient implementation for reading lines instead.

Some profiling results (150 MB log file):

Master:
master

#2792:
pr-2792

This PR:
pr

@slook This change is relevant for #2792.

### DIFF
--- a/pynicotine/logfacility.py
+++ b/pynicotine/logfacility.py
@@ -77,6 +77,8 @@ class Logger:
         UnknownPeerMessage
     }
 
+    READ_BLOCK_SIZE = 4096
+
     def __init__(self):
 
         current_date_time = time.strftime("%Y-%m-%d_%H-%M-%S")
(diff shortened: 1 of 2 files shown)


slook:
Genius! This is like the way I had originally wanted to do this, because it avoids the need to count the number of lines...

However, my mathematics are not good enough to figure out how this solution would be integrated with my new infinite scrolling feature - although it seems like it should be doable - would you be willing to accept the challenge of integrating your solution into #2792 for me? I guess we keep track of the current block instead of the current line, and somehow align it to the nearest newline?

pynicotine/logfacility.py
```diff
@@ -220,6 +222,38 @@ def update_folder_paths(self):
...
+
+        file_handle.seek(0, os.SEEK_END)
+
+        blocks = deque()
+        block_number = 1
+        file_size = file_handle.tell()
+        lines_left = num_lines + 1
+
+        while lines_left > 0:
+            read_offset = file_size - (block_number * self.READ_BLOCK_SIZE)
```
slook, line 235: `log_file.read_offset -= (block_number * self.READ_BLOCK_SIZE)`

pynicotine/logfacility.py
```diff
@@ -220,6 +222,38 @@ def update_folder_paths(self):
     def open_log(self, folder_path, basename):
         self._log_file_operation(folder_path, basename, self.open_log_callback)
 
+    def _get_log_lines(self, file_handle, num_lines):
+
+        file_handle.seek(0, os.SEEK_END)
+
+        blocks = deque()
+        block_number = 1
+        file_size = file_handle.tell()
```
slook, line 231: Suggest storing the starting `read_offset` in the `LogFile` object, because that would then allow the log history to be prepended at any time after application startup (for example when switching to the relevant chat tab), thereby improving startup time even further.

pynicotine/logfacility.py
mathiascode: Got rid of the block number entirely. I used it while experimenting, but it became redundant eventually.

mathiascode:
> However, my mathematics are not good enough to figure out how this solution would be integrated with my new infinite scrolling feature - although it seems like it should be doable - would you be willing to accept the challenge of integrating your solution into #2792 for me? I guess we keep track of `read_offset` instead of the current lineno, and somehow align it to the nearest newline?

I could look closer at it at some point, but yeah, should be enough to store the `read_offset` in the `LogFile`, adjusting it to the nearest newline before doing so. When we need to read more, we just read more blocks starting the last offset. We wouldn't have to track the number of lines anymore.

pynicotine/logfacility.py
mathiascode: We can do this in your PR.

PR #3271, opened 2025-02-06
logfacility.py: READ_BLOCK_SIZE -> io.DEFAULT_BUFFER_SIZE

slook:
+ Changed: Use built-in constant instead of READ_BLOCK_SIZE use [io.DEFAULT_BUFFER_SIZE](https://docs.python.org/3/library/io.html#io.DEFAULT_BUFFER_SIZE):

> _An int containing the default buffer size used by the module’s buffered I/O classes. [open()](https://docs.python.org/3/library/functions.html#open) uses the file’s [blksize](https://docs.python.org/3/library/os.html#os.stat_result.st_blksize) (as obtained by os.stat()) if possible._

On my system this is `8192` instead of `4096`.

Imported the entire `io` module since other functions from it are likely to be required in future.

### DIFF
--- a/pynicotine/logfacility.py
+++ b/pynicotine/logfacility.py
@@ -16,6 +16,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import io
 import os
 import sys
 import time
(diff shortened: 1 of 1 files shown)


mathiascode:
Nice! Didn't know about that one.

PR #3276, opened 2025-02-11
Build: convert svg icons to png when necessary

### LINKED ISSUE #3273: The only valid location for the SVG icon on Linux is `/usr/share/icons/hicolor/scalable/apps`
Nicotine+ version:  3.3.7
Operating System/Distribution: Linux

### Describe the bug

Currently the icon is present in every category under `/usr/share/icons/hicolor` but should be present only in `/usr/share/icons/hicolor/scalable` because it's not a raster image of the respective sizes. 

### Expected behavior

Nicotine+'s icon is only present under `/usr/share/icons/hicolor/scalable/apps/org.nicotine_plus.Nicotine.svg`.

mathiascode:
Fixes #3273

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt update
-          sudo apt install gettext
+          sudo apt install gettext librsvg2-bin
           python -m pip install build
 
       - name: Build package
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -117,6 +117,7 @@ jobs:
             ${{ matrix.prefix }}-gettext-tools
             ${{ matrix.prefix }}-gtk4
             ${{ matrix.prefix }}-libadwaita
+            ${{ matrix.prefix }}-librsvg
             ${{ matrix.prefix }}-python-build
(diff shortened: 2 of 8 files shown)


slook:
https://github.com/marketplace/actions/converting-svg-files-to-png-files#example-workflow-file
> _"This action is designed to convert all images in SVG (Scalable Vector Graphics) format located in a specified directory into PNG (Portable Network Graphics) format. The converted PNG images are then saved in a separate target directory."_

https://github.com/joergnapp/convert-svg-to-png
https://github.com/joergnapp/convert-svg-to-png/blob/main/svg-to-png.sh#L20
https://github.com/joergnapp/convert-svg-to-png/blob/main/action.yml

mathiascode:
Now that I think about it, we might get away with librsvg alone (which was already an implicit dependency) if we convert the SVGs with Gdk.Pixbuf.

mathiascode:
If we use GdkPixbuf on build time, we'd have to install PyGObject in a virtual environment when running `python -m build`, which requires installing a compiler and Python header files. Too painful.

I'm going with https://github.com/nicotine-plus/nicotine-plus/pull/3279 instead, since we already had raster icons in the repo anyway (Windows and macOS).

PR #3280, opened 2025-02-12
Add lowercase file path mapping for Soulseek NS clients

### LINKED ISSUE #3277: Soulseek NS unable to download folder from search results
i started noticing anomalies when trying to download (it fails with file not shared) and browsing clients (list never loads).
when i talked to these people (at least the messaging works...) common denominator was that they're using nicotine+ !

go figure now...
i don't think nicotine can be called soulseek client - it's more of a leecher as when i asked if they can download from me - sure, they can...

mathiascode:
Make things a bit easier for our users, so they don't have to deal with private messages from Soulseek NS users about not sharing files.

Fixes #3277

### DIFF
--- a/pynicotine/shares.py
+++ b/pynicotine/shares.py
@@ -255,7 +255,8 @@ class Scanner:
     __slots__ = ("queue", "share_groups", "share_dbs", "share_db_paths", "init",
                  "rescan", "rebuild", "reveal_buddy_shares", "reveal_trusted_shares",
                  "files", "streams", "mtimes", "word_index", "processed_share_names",
-                 "processed_share_paths", "current_file_index", "current_folder_count")
+                 "processed_share_paths", "current_file_index", "current_folder_count",
+                 "lowercase_paths")
 
     HIDDEN_FOLDER_NAMES = {"@eaDir", "#recycle", "#snapshot"}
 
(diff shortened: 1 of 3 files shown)


mathiascode:
Thankfully, only file transfers are affected, so we don't have to do anything special with folder content requests.

pynicotine/transfers.py
```diff
@@ -96,6 +97,7 @@ def __init__(self, username, virtual_path=None, folder_path=None, size=0, file_a
         self.legacy_attempt = False
         self.retry_attempt = False
         self.size_changed = False
+        self.is_path_mismatched = False
```
slook, line 100: `is_case_mismatched` or `is_lowercase_path`

pynicotine/shares.py
```diff
@@ -762,13 +771,20 @@ def remove_db_file(db_path):
             import shutil
             shutil.rmtree(db_path_encoded)
 
-    def virtual2real(self, virtual_path):
+    def virtual2real(self, virtual_path, check_lowercase=False):
```
slook, line 774: `is_case_mismatched` or `is_lowercase_path`

slook (changes requested):

PR #3287, opened 2025-02-18
Update SLSKPROTOCOL.md: peer connection message order

boristopalov:
It seems there are a couple intermediate steps skipped in the docs on outgoing connection attempts to peers. Specifically, when attempting an outgoing connection to another user, the client first sends a `GetPeerAddress` message to the server to get the user's connection details, and only after receiving a response from the server do they attempt a direct connection request. These 2 steps are added in the docs

I was following the logic [here](https://github.com/nicotine-plus/nicotine-plus/blob/master/pynicotine/slskproto.py#L754)


doc/SLSKPROTOCOL.md
```diff
@@ -2290,21 +2290,23 @@ a peer. In Nicotine+, these messages are defined in slskmessages.py.
 
 1.  User A sends [ConnectToPeer](#server-code-18) to the Server with a unique
     token (indirect connection request)
-2.  User A sends a [PeerInit](#peer-init-code-1) to User B (direct connection
-    request)
-3.  The Server sends a [ConnectToPeer](#server-code-18) response to User B with
+2.  User A sends [GetPeerAddress](#server-code-3) to the Server with the same token and connection type
```
mathiascode, line 2293: GetPeerAddress doesn't take a token or connection type. You could mention the token in step 4 instead, and explain why we send GetPeerAddress here.

mathiascode:
Could you also add GetPeerAddress to the legacy connection message order (should be step 1)?

boristopalov:
Sorry for the slow response- I haven't forgotten about this. I've updated the wording to make it clearer.

I've also added a typical flow for: searching for a file > connecting to a peer with a result > downloading the result. I'm writing a client in Go and having some type of example would have saved me some time. Please let me know if it looks off or if there's a better place to put it.

Thanks for your hard work on this project!

mathiascode:
Thanks, I'll review the changes soon.

mathiascode:
Sorry for the late response. The contents look good, just some minor formatting issues that I can fix after merging.

doc/SLSKPROTOCOL.md
```diff
@@ -2289,22 +2289,25 @@ a peer. In Nicotine+, these messages are defined in slskmessages.py.
...
-    token (indirect connection request)
-2.  User A sends a [PeerInit](#peer-init-code-1) to User B (direct connection
-    request)
-3.  The Server sends a [ConnectToPeer](#server-code-18) response to User B with
-    the same token.  
+    token (indirect connection request) and User B's username. The token is stored
+    for later use (see steps 4 and 5).
+2.  User A sends [GetPeerAddress](#server-code-3) to the Server with User B's username.
+3.  The Server responds to User A with [GetPeerAddress](#server-code-3), which gives User A the IP of User B.
+4.  User A sends a [PeerInit](#peer-init-code-1) to User B (direct connection
```
mathiascode, line 2296: I didn't catch this before merging, but the token in the PeerInit is always 0, and goes unused (probably a leftover from the early days of the Soulseek protocol).

PR #3336, opened 2025-03-15
DEVELOPING.md: don't mention oldest Python version number

slook:
- Removed: Inaccurate mention of Python `3.6`
+ Added: Link to [Status of Python versions chart](https://devguide.python.org/versions/#python-release-cycle) on the Python website
+ Added: When incrementing `B` mention the creation of old feature branches to enable critical bug fix maintenance
+ Added: Fourth point to keep in mind about old branch maintenance

### DIFF
--- a/doc/DEVELOPING.md
+++ b/doc/DEVELOPING.md
@@ -29,8 +29,8 @@ Developing in a single language is also easier for everyone involved in the
 project.
 
 We aim to support the oldest [Python version](https://devguide.python.org/versions/)
-still receiving security updates, up to the latest.
-
+still receiving security updates, up to the latest. In any case, there's no rush to
+bump the minimum version requirement unless it brings a new feature that we need.
 
 ### GTK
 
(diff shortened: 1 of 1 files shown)


doc/DEVELOPING.md
```diff
@@ -28,9 +28,8 @@ allows for running Nicotine+ on almost any system without compiling anything.
 Developing in a single language is also easier for everyone involved in the
 project.
 
-We aim to support the oldest minor Python 3 version still used by supported
-releases of distributions and operating systems. The minimum version Nicotine+
-currently supports is 3.6.
+We aim to support the oldest [Python version](https://devguide.python.org/versions/)
```
mathiascode, line 31: Debian-based distros will probably lag behind. We'll have to see how this works in practice.

doc/DEVELOPING.md
```diff
@@ -249,13 +248,15 @@ within 24 hours. In order to preserve author information for commits, use the
...
     significantly change the workflow of the application. In practice, this is
     unlikely to happen.
-  - B is incremented when significant features or new dependencies are added,
+  - `B` is incremented when significant features or new dependencies are added,
     version requirements are increased, or in the case of major changes to
-    existing code that are deemed too intrusive for a smaller release.
-  - C is incremented when bug fixes, performance improvements or smaller
+    existing code that are deemed too intrusive for a smaller release. Then the
+    previous feature branch (e.g. `3.3.x`) continues to be maintained only with
+    critical backports, until it's dependencies reach beyond end-of-life.
```
mathiascode, line 258: Realistically, the previous branch will be maintained until the next `B` is released (3.3.x will reach EOL once 3.4.0 is released). There isn't much point in maintaining the older branch after that, since Debian-based distros will not package such releases (only major software like web browsers receive major version bumps, everything else is frozen with the exception of security backports).

doc/DEVELOPING.md
mathiascode: In any case, there's no rush to bump the version requirements. New Python versions don't bring that many changes we need.

doc/DEVELOPING.md
```diff
@@ -28,9 +28,9 @@ allows for running Nicotine+ on almost any system without compiling anything.
 Developing in a single language is also easier for everyone involved in the
 project.
 
-We aim to support the oldest minor Python 3 version still used by supported
-releases of distributions and operating systems. The minimum version Nicotine+
-currently supports is 3.6.
+We aim to support the oldest [Python version](https://devguide.python.org/versions/)
+still receiving security updates, up to the latest. In any case, there's no rush to
+bump the minimum version requirement unless it brings a new feature that we need.
```
mathiascode, line 33:
```suggestion
still receiving security updates, up to the latest. There is no rush to bump the
minimum version requirement, since LTS distributions may use older Python
versions, but a smaller range of versions is easier to test.
```

PR #3337, opened 2025-03-16
DEVELOPING.md: simplify Continuous Integration Testing section

slook:
- Removed: Outdated and irrelevant detail about running the test suite locally, since GitHub Actions make this task superfluous.

### DIFF
--- a/doc/DEVELOPING.md
+++ b/doc/DEVELOPING.md
@@ -145,60 +145,43 @@ by the Nicotine+ team.
 
 ## Continuous Integration Testing
 
-It is important that all patches pass unit testing. Unfortunately developers
-make all kinds of changes to their local development environment that can have
-unintended consequences. This means sometimes tests on the developer's computer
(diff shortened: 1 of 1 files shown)


mathiascode:
I'm not sure how useful the Autopkgtest section is. Admittedly, I've never used the commands there.

mathiascode:
How about something like this? On second thought, it's probably not worth mentioning autopkgtest, since it's Debian-specific and GitHub Actions already runs it.

# Continuous Integration Testing

It is important that all changes pass unit and integration testing. To properly validate that things are working, continuous integration (CI) is implemented using GitHub Actions workflows, which run tests on various platforms for each commit.

### Creating Tests

...

### Running Tests

Although GitHub Actions runs tests automatically when pushing changes to the repository, it is possible to run them locally. This can be helpful to quickly test smaller changes, but should not be relied on for changes that affect packaging or are likely to break on other platforms.

To run the tests using the Python unittest module:
`python3 -m unittest`

PR #3339, opened 2025-03-19
shares.py: accurate current_folder_count during rescan

slook:
This fixes the scanner progress label getting stuck on the same high number for ages in common cases, where the tree is shallow. The problem was that most of the subfolders got counted from the outset well before their files were actually scanned...

+ Fixed: Increment and emit `current_folder_count` as its files are about to be scanned, instead of when initially encountered.
+ Fixed: Also count root folders, so that the logged finished number matches up with actual Shared Folders shown in browse.
+ Fixed: Don't count duplicate folder paths, because such nested folders aren't shared twice.
+ Fixed: Possible risk of infinite loop, `break` out of both loops on `ScannerState.FAILURE` to halt any further processing of the `scanner_queue` because it is not needed to keep looping since the preceding error message gets logged beforehand.

The 100 modulo workaround is no longer required. The difference in scanning performance is negligible or none, for rescan or rebuild respectively. It is reassuring to see constant progress being made in the status bar, especially when doing a full rebuild.


mathiascode:
> The 100 modulo workaround is no longer required. The difference in scanning performance is negligible or none, for rescan or rebuild respectively. It is reassuring to see constant progress being made in the status bar, especially when doing a full rebuild.

Still an uncomfortable amount of CPU time spent with a lot of small folders:

image

Most of this time is wasted, since the GUI doesn't have a chance to update on each call.

slook:
Does this improve it?

PR #3340, opened 2025-03-19
ChatView: implement clickable room tags and prepend_log_lines

### LINKED ISSUE #2482: Public chat room feed: Don't notify on new messages + clickable room names

### LINKED ISSUE #3200: username highlighting bug
Nicotine+ 3.3.6
macOS 15.1

In the chat window, when the username matches the current timestamp, the timestamp gets highlighted instead of the username.
See below.

Screenshot 2024-11-20 at 20 45 19

lowkey gave me a chuckle
cheers

slook:
The is the largest group of changes that I have planned for 3.4.x, the TextView and ChatView widgets are entirely refactored as per draft PR #2792...

+ Added: Global room tags, close #2482
+ Fixed: Part of the timestamp was incorrectly tagged if the `username` is a short numeric that's also in timestamp, close #3200
- Changed: Rename `append_line()` -> `add_line()` and new `prepend` option to insert old log lines in reverse order

This PR will make it possible to implement new features such as lazy-loading, infinite scrolling, separate timestamp fonts, etc, in the future if desired, but those are not implemented here, it is only refactored in preparation.

For now, there shouldn't be any noticeable difference of functionality in the UI. The renamed method in the widget isn't directly callable from plugins so it shouldn't cause any effect on that either. Testing and feedback is welcomed, as always, thank you.


mathiascode:
Great job, this works well!

Could we add primary and secondary callbacks to tags instead of a single callback? Opening rooms when right-clicking a room name is unexpected.

I would also leave out the room name color for now, or maybe experiment with using an underline.

pynicotine/gtkgui/widgets/textview.py
```diff
@@ -404,10 +407,70 @@ def __init__(self, *args, chat_entry=None, status_users=None, username_event=Non
...
+        if username:
+            opener, closer = ("* ", " ") if message_type == "action" else ("[", "] ")
+
+            yield (opener, tag)
+            yield (username, self._get_user_tag(username))
+            yield (closer, tag)
+
+        # Highlight urls, if found and tag them
+        for hypertext, tag in list(self._generate_hypertext(message, tag=tag)):
+            yield (hypertext, tag)
```
mathiascode, line 456:
```suggestion
        yield from self._generate_hypertext(message, tag=tag)
```

pynicotine/logfacility.py
```diff
@@ -254,14 +254,14 @@ def _get_log_lines(self, log_file, num_lines):
...
-        lines = None
+        lines = []
         log_file = None
 
         try:
             log_file = self._get_log_file(folder_path, basename, should_create_file=False)
 
             if log_file is not None:
-                lines = self._get_log_lines(log_file, num_lines)
+                lines = reversed(self._get_log_lines(log_file, num_lines))
```
mathiascode, line 264: I think reversed() would be a better fit in gtkgui for now.

mathiascode:
One bug I noticed: when colored usernames are disabled, only the username in the latest log line remains uncolored. All other usernames use the remote/local text color.

slook:
> One bug I noticed: when colored usernames are disabled, only the username in the latest log line remains uncolored. All other usernames use the remote/local text color.

I can't replicate this behaviour. Are you talking about it occurring when the preference is just changed, or at all times while colored usernames are disabled?

mathiascode:
> > One bug I noticed: when colored usernames are disabled, only the username in the latest log line remains uncolored. All other usernames use the remote/local text color.
> 
> I can't replicate this behaviour. Are you talking about it occurring when the preference is just changed, or at all times while colored usernames are disabled?

Seems to only happen to lines that are restored from history.

1. Preferences -> UI -> Disable colored usernames
2. Preferences -> UI -> Set a remote text color

image

slook:
> Could we add primary and secondary callbacks to tags instead of a single callback?

Done.

> I would also leave out the room name color for now, or maybe experiment with using an underline.

I made it follow the same font style as user tags but with the color of URLs, so underline can be set if that's what you want.

mathiascode:
> > One bug I noticed: when colored usernames are disabled, only the username in the latest log line remains uncolored...
> 
> It is not only the latest log line that is correctly uncolored, any remote line above an action or mention is also features an uncolored user tag as well.
> 
> The only workaround that I have found so far is by setting an explicit color for the uncolored user tags, but no such default color exists in the config and I don't know how to get the default text foreground color. So as it is, the bug remains.

Is there an issue with the way tags are applied? It works correctly in master, and I don't think the logic changed here? (usernames should only have the username tag)

slook:
The logic has somewhat changed, I suppose, it seems like what is happening is that some element of the style of the previous tag is being inherited when the color is None. Maybe it is some undocumented fallback mechanism that we are not aware of.

> Edit: It's something with the line breaks again. If I remove this code ... the names are uncolored, as expected.

Hooray! I suppose the line breaks are added in backwards so the logic has changed in that way. But what to do about it then?

slook:
> Why not pass a primary and secondary callback to create_tag(), instead of having to deal with return values?

I thought it seemed like that would be an unnecessarily large amount of data to store within each and every tag, considering that the additional complexity of separate events for each button aren't really required since all we want to do is block an action.

mathiascode:
> The logic has somewhat changed, I suppose, it seems like what is happening is that some element of the style of the previous tag is being inherited when the color is None. Maybe it is some undocumented fallback mechanism that we are not aware of.
> 
> > Edit: It's something with the line breaks again. If I remove this code ... the names are uncolored, as expected.
> 
> Hooray! I suppose the line breaks are added in backwards so the logic has changed in that way. But what to do about it then?

It smells like a bug in GTK. If I insert the line break before the rest of the line, things work as expected:

```
        if prepend and self.textbuffer.get_char_count() > 0:
            self.textbuffer.insert(iterator, "\n")
            iterator = self.get_iter_at_line(line_number)
```

mathiascode:
> > Why not pass a primary and secondary callback to create_tag(), instead of having to deal with return values?
> 
> I thought it seemed like that would be an unnecessarily large amount of data to store within each and every tag, considering that the additional complexity of separate events for each button aren't really required since all we want to do is block an action.

I don't think storing a reference to an additional callback will have a noticeable impact. It's only one tag per username after all.

It didn't seem that complex to implement when I tried some time ago, basically just store `tag.primary_callback` and `tag.secondary_callback` instead of `tag.callback`, and use the correct one in the primary/secondary click handlers.

slook:
> If I insert the line break before the rest of the line, things work as expected:

Nice observation, but I don't think that solution will work as expected in the future case when there is already content in the text view before the log is prepended. All your doing there really is inserting the newline at the last iterator position, which could well be after the most recent message, whereas it is necessary to define the desired insertion point at line `0` for prepending.

> use the correct one in the primary/secondary click handlers

In this PR I combined both handlers into a single method. However your proposal is agreeably a more proper way.

mathiascode:
> Nice observation, but I don't think that solution will work as expected in the future case when there is already content in the text view before the log is prepended. All your doing there really is inserting the newline at the last iterator position, which could well be after the most recent message, whereas it is necessary to define the desired insertion point at line `0` for prepending.

My code inserts the line break at the start iterator, and updates it to point at the start of the text buffer before inserting the rest of the line.

mathiascode:
Thanks! Could we keep `text_line.insert(0, ("\n", None))`, to avoid creating another iterator?

mathiascode:
I looked into how to optimize text insertions, and came up with the following:

```python
    def _insert_line(self, text_line, prepend=False):

        line_number = 0 if prepend else self.textbuffer.get_line_count()
        iterator = self.get_iter_at_line(line_number)
        offset = iterator.get_offset()
        second_iterator = None

        # None tag on line breaks to prevent wrong color glitch on next line
        if prepend and self.textbuffer.get_char_count() > 0:
            text_line.append(("\n", None))

        elif line_number > 1 or self.textbuffer.get_char_count() > 0:
            text_line.insert(0, ("\n", None))

        self.textbuffer.insert(iterator, "".join(text for text, _tag in text_line))

        for text, tag in text_line:
            if tag is None:
                offset += len(text)
                continue

            if second_iterator is None:
                second_iterator = iterator.copy()

            iterator.set_offset(offset)
            offset += len(text)
            second_iterator.set_offset(offset)

            self.textbuffer.apply_tag(tag, iterator, second_iterator)

        self._remove_old_lines(line_number)
```

The good news is that it's much faster, and doesn't suffer from the username color bug!

pynicotine/gtkgui/widgets/textview.py
```diff
@@ -153,50 +184,22 @@ def _remove_old_lines(self, num_lines):
...
+            line.append((" ", None))
 
         # Highlight urls, if found and tag them
-        if self.parse_urls and ("://" in line or "www." in line or "mailto:" in line):
-            # Match first url
-            match = self.URL_REGEX.search(line)
-
-            while match:
-                self._insert_text(line[:match.start()], tag)
+        line.extend(list(self._generate_hypertext(message)))
```
mathiascode, line 200:
```suggestion
        line.extend(self._generate_hypertext(message))
```

mathiascode (approved):

PR #3368, opened 2025-05-03
search.py: implement folder-level filtering

### LINKED ISSUE #1023: Album/Folder Level Result Filters
Add result filters that match for the whole folder instead of file level:
Minimum Matching Files: Only shows results for folders that have at least X number of files that match the file type filter. So if a folder has 20 images but only one song that matches, filter out the folder.
Maximum Matching Files: Only shows results for folders that have at most X number of files that match the file type filter. So if an album has multiple rereleases you can filter out the rereleases easier.
Contains: Only show results for folders that have at least one file that matches text or !text. Ex. Main Search term "Animal Tracks" with Folder Filter Contains "Mess Around" would show results matching for https://en.wikipedia.org/wiki/Animal_Tracks_(British_album)

mathiascode:
The code and UI needs some tweaks/cleanups (I want to avoid setting `self.found_exclude_filter` while filtering), but the feature works.

Closes #1023

### DIFF
--- a/pynicotine/gtkgui/search.py
+++ b/pynicotine/gtkgui/search.py
@@ -441,6 +441,7 @@ class Search:
             self.filters_button,
             self.filters_container,
             self.filters_label,
+            self.folder_level_filtering_button,
             self.grouping_button,
             self.results_button,
             self.results_label,
@@ -470,6 +471,8 @@ class Search:
         self.active_filter_count = 0
         self.num_results_found = 0
         self.num_results_visible = 0
+        self.found_include_filter = False
+        self.found_exclude_filter = False
 
         # Use dict instead of list for faster membership checks
         self.selected_users = {}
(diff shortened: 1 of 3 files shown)


pynicotine/gtkgui/search.py
```diff
@@ -789,6 +792,46 @@ def set_filters(self, stored_filters):
...
+                current_folder = folder_path
+
+            if self.found_exclude_filter:
+                continue
+
+            folder_results.append(row)
+
+            if self.check_filter(row):
+                self.found_include_filter = True
+                update_ui = True
```
slook, line 830: > I want to avoid setting `self.found_exclude_filter` while filtering

```
else:
     self.found_exclude_filter = True
```

At least having one instance of it here would be better than it being scattered amongst functions.

PR #3388, opened 2025-05-18
preferences.py: remove "Enable colored usernames" option toggle

slook:
- Removed: "Enable colored usernames" option toggle from User Interface Preferences

This will also remove the `"usernamehotspots"` config setting, because there doesn't seem to be any advantage having uncolored username tags in chat, and disabling the option causes `None` color to be set in the theme implementation [which is problematic](https://github.com/nicotine-plus/nicotine-plus/pull/3340#issuecomment-2888560770).


mathiascode:
I already did this once, but restored the option because a long-time user needed it.

PR #3393, opened 2025-05-22
ui: set "no-emoji" input hint on GtkEntry widgets

slook:
- Removed: "Insert Emoji" menu item from context menu of GtkEntry widgets where emoji are definitely invalid

This PR doesn't deal with Entry fields that are programmatically created within ComboBox and Spinner widgets, dialogs, etc.


mathiascode:
Spinners have the no-emoji hint set in later GTK 4 versions anyway.

PR #3397, opened 2025-05-24
Implement string pluralization

mathiascode:
https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html

Replaces #2466

### DIFF
--- a/pynicotine/plugins/core_commands/__init__.py
+++ b/pynicotine/plugins/core_commands/__init__.py
@@ -19,6 +19,7 @@
 from pynicotine.pluginsystem import BasePlugin
 from pynicotine.shares import PermissionLevel
 from pynicotine.slskmessages import UserStatus
+from pynicotine.utils import humanize
 
 
 class _CommandGroup:
@@ -293,10 +294,18 @@ class Plugin(BasePlugin):
         output_text = ""
 
         if not search_query:
-            output_text += _("Listing %(num)i available commands:") % {"num": num_commands}
+            output_text += ngettext(
+                "Listing %(num)s available command:",
+                "Listing %(num)s available commands:",
+                num_commands
+            ) % {"num": humanize(num_commands)}
         else:
-            output_text += _('Listing %(num)i available commands matching "%(query)s":') % {
-                "num": num_commands,
+            output_text += ngettext(
+                'Listing %(num)s available command matching "%(query)s":',
+                'Listing %(num)s available commands matching "%(query)s":',
+                num_commands
+            ) % {
+                "num": humanize(num_commands),
                 "query": search_query
             }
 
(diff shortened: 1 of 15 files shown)


pynicotine/users.py
```diff
@@ -428,26 +431,30 @@ def _privileged_users(self, msg):
...
+        if seconds <= 0:
             log.add(_("You have no Soulseek privileges. While privileges are active, your downloads "
                       "will be queued ahead of those of non-privileged users."))
 
             if self._should_open_privileges_url:
                 self.open_privileges_url()
+
+        elif seconds > 86400:
+            days = ceil(seconds / 86400)
+            log.add(_("%(num)s days of Soulseek privileges left"), {"num": humanize(days)})
```
slook, line 445: Considering a new function in `utils.py` for any case where an approximate estimation of duration is desired...

- `_("%(duration)s of Soulseek privileges left"), {"duration": human_duration(seconds)}`
_**About 2 years** of Soulseek privileges left_
_**Over 1 year** of Soulseek privileges left_
_**365 days** of Soulseek privileges left_
_**1 day** of Soulseek privileges left_
_**23 hours** of Soulseek privileges left_
_**1 hour** of Soulseek privileges left_
_**59 minutes** of Soulseek privileges left_
_**1 minute** of Soulseek privileges left_
_**59 seconds** of Soulseek privileges left_
_**Less than 1 second** of Soulseek privileges left_

- `_("Wishlist wait period set to %(duration)s"), {"duration": human_duration(seconds)}`
_Wishlist wait period set to **12 minutes**_

- `_("Reconnecting to server in %(duration)s"), {"duration": human_duration(server_timeout)}`
_Reconnecting to server in **5 minutes**_
_Reconnecting to server in **1 minute**_
_Reconnecting to server in **20 seconds**_

... where only a single unit of measurement is needed, i.e. year(s), day(s), hour(s), minute(s) or second(s) for a sane indication of the rough passing of time. No need to bother with week(s), up to about one years' worth of days are easy enough to comprehend but no more so approximate year(s) (without worrying about leap years) are wanted in this case, imo.

Or call the function `human_time()`, `human_eta()`, `human_length_approx()` or something else. These time comprehensions are nice but they don't really belong in the users module.

pynicotine/plugins/core_commands/__init__.py
```diff
@@ -19,6 +19,7 @@
 from pynicotine.pluginsystem import BasePlugin
 from pynicotine.shares import PermissionLevel
 from pynicotine.slskmessages import UserStatus
+from pynicotine.utils import humanize
```
slook, line 22: Not required here because there are unlikely to be over 1000 custom plugin commands. If there are that many, then the user has larger problems to worry about other than the proper localization of numerals in the help!

slook (changes requested):

pynicotine/users.py
mathiascode: I have a feeling approximate durations (about, over, less than) as separate strings are problematic in some languages.

slook: Then it would be best to just keep it simple for now, not do years, only go up to days and all the way down to 1 second.

PR #3398, opened 2025-05-25
utils.py: new human_duration() function using ngettext()

slook:
+ Added: 8 pluralized translatable strings that represent basic units of time in day(s), hour(s), minute(s) or second(s).
+ Changed: 3 translatable strings for the sake of demonstrating the use of `ngettext()` where `$(duration)s` are involved.

These time strings aren't ever concatenated with each other, only a singular unit of measurement needs to be displayed at a time for these rounded values because they don't need to have the same degree of precision as something like an audio track length.

For instance, it is clearer to say "**_n_ days** of Soulseek privileges left" because it isn't useful to know the exact number of hours, minutes nor seconds until the amount left falls low enough for it to become "**_n_ hours** of Soulseek privileges left", and so on.

https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html
https://www.gnu.org/software/gettext/manual/html_node/Translating-plural-forms.html

Related to PR #3397.

### DIFF
--- a/pynicotine/i18n.py
+++ b/pynicotine/i18n.py
@@ -105,4 +105,4 @@ def apply_translations(language=None):
     _set_system_language(language)
 
     # Install translations for Python
-    gettext.install(TRANSLATION_DOMAIN, LOCALE_PATH)
+    gettext.install(TRANSLATION_DOMAIN, LOCALE_PATH, names=["ngettext"])
--- a/pynicotine/search.py
+++ b/pynicotine/search.py
@@ -39,6 +39,7 @@ from pynicotine.slskmessages import SEARCH_TOKENS_ALLOWED
 from pynicotine.slskmessages import UserSearch
 from pynicotine.slskmessages import WishlistSearch
 from pynicotine.utils import TRANSLATE_PUNCTUATION
+from pynicotine.utils import human_duration_approx
 
 
 class SearchRequest:
@@ -421,8 +422,10 @@ class Search:
         self.wishlist_interval = msg.seconds
 
         if self.wishlist_interval > 0:
-            log.add_search(_("Wishlist wait period set to %s seconds"), self.wishlist_interval)
-
+            log.add_search(
+                _("Wishlist wait period set to %(duration)s"),
+                {"duration": human_duration_approx(self.wishlist_interval)}
+            )
             events.cancel_scheduled(self._wishlist_timer_id)
             self._wishlist_timer_id = events.schedule(
                 delay=self.wishlist_interval, callback=self._do_next_wishlist_search, repeat=True)
--- a/pynicotine/slskproto.py
+++ b/pynicotine/slskproto.py
@@ -87,6 +87,7 @@ from pynicotine.slskmessages import UserStatus
 from pynicotine.slskmessages import WatchUser
 from pynicotine.slskmessages import increment_token
 from pynicotine.slskmessages import initial_token
+from pynicotine.utils import human_duration_approx
 from pynicotine.utils import human_speed
 
 
@@ -1095,7 +1096,8 @@ class NetworkThread(Thread):
             self._server_timeout_value *= 2
 
         self._server_timeout_time = time.monotonic() + self._server_timeout_value
-        log.add(_("Reconnecting to server in %s seconds"), self._server_timeout_value)
+        log.add(_("Reconnecting to server in %(duration)s"),
+                {"duration": human_duration_approx(self._server_timeout_value)})
 
     @staticmethod
     def _set_server_socket_keepalive(sock, idle=10, interval=2):
--- a/pynicotine/users.py
+++ b/pynicotine/users.py
@@ -33,6 +33,7 @@ from pynicotine.slskmessages import UnwatchUser
 from pynicotine.slskmessages import UserStatus
 from pynicotine.slskmessages import WatchUser
 from pynicotine.utils import UINT32_LIMIT
+from pynicotine.utils import human_duration_approx
 from pynicotine.utils import open_uri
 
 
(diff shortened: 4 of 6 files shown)


slook:
The `human_duration()` function has 3 optional arguments which can adjusted by a developer (including a plugin developer) to control the points at which the final unit of measurement is chosen. By default, for any given input number of `seconds` the output string that is returned will be in the unit of...

- hour/hours if the duration is less than a day (e.g. "23 hours"), or;
- minute/minutes if the duration is less than an hour (e.g. "59 minutes"), or;
- second/seconds if the duration is less than a minute (e.g. "59 seconds").

However, for example, I decided the privileges string shall ideally be expressed in the more precise unit of hour/hours if there are less than 3 days (with `max_hours=72`), because this helps the privileged user determine the rough time of day upon which their privileges will expire, not merely the vague number of days until they are going to be expiring.

This is accomplished in this PR by calling `human_duration(seconds, max_hours=72)` from `shares.py` (since 72 hours equals 3 days). It results in outputs such as:

- ...
- 4 days of Soulseek privileges left _(i.e. more than 96 hours)_
- 3 days of Soulseek privileges left _(i.e. somewhere between `72` to 96 hours)_
- 71 hours of Soulseek privileges left
- 70 hours of Soulseek privileges left
- ...
- 25 hours of Soulseek privileges left
- 24 hours of Soulseek privileges left _(i.e. just over 1 day)_
- ...
- 2 hours of Soulseek privileges left
- 1 hour of Soulseek privileges left
- 59 minutes of Soulseek privileges left
- ...
- 2 minutes of Soulseek privileges left
- 1 minute of Soulseek privileges left
- 59 seconds of Soulseek privileges left
- ...
- 2 seconds of Soulseek privileges left
- 1 second of Soulseek privileges left

We might, for instance, abolish reporting of seconds by calling `human_duration(seconds, max_seconds=-1, max_hours=72)`...

- 2 minutes of Soulseek privileges left
- 1 minute of Soulseek privileges left
- 0 minutes of Soulseek privileges left _(i.e. somewhere between 1 to 59 seconds)_
- _... or it could be preferred to simply report something like "Less than a day" for values under 24 hours_

For the "Reconnecting to server in ..." string, a developer implementing a status widget might decide that it would be desirable to have more precision with `max_seconds=120` such that, after falling below "2 minutes" then "1 minute" is skipped in favour of instead measuring shorter durations in seconds as the upcoming event (whatever that may be) becomes imminent...

- Reconnecting to server in 5 minutes
- Reconnecting to server in 4 minutes
- Reconnecting to server in 3 minutes
- Reconnecting to server in 120 seconds
- Reconnecting to server in 119 seconds
- ...
- Reconnecting to server in 2 seconds
- Reconnecting to server in 1 second

Other use cases could be how long ago a buddy was last seen, when a statistics record began etc, and this or another function will be expanded to improve the display of estimated time remaining for transfers that take over a day, to resolve https://github.com/nicotine-plus/nicotine-plus/discussions/1533#discussioncomment-2011019 .

The upshot is that the delineations of the duration unit might not always be in the obvious unit as per a standard clock, but rather the chosen unit of measurement depends upon what level of precision is deemed sensible for a given purpose at any given time.

hboetes:
That's a lot to explain in 2 sentences, how bout?

> time display accuracy; higher value, more accurate(123 seconds), -1, less accurate (2 minutes)

slook:
Maybe the max_ optional arguments are too complicated and niche. I may just remove the options as fixing the thresholds to standardized clock values could be satisfactory enough for most use cases and would also make the interface be simpler to understand and more consistent/predicable.

The main concern I had that caused me to request your review was about the feasibility of programmatically embedding the strings which the utility function generates into other translatable strings (thereby reducing the total number of strings in the program), as opposed to having all the needed time expressions baked into the main strings themselves (potentially resulting in needing a larger number of strings).

I don't know if the end result would be grammatically correct in all cases/languages, or not? Do you think this strategy might be an acceptable approach for us to take?

slook:
> Could you explain to me what can be configured here, more verbosely?

Ask a silly question... :)

hboetes:
Not at all,

> Approximate long time span rounded to nearest unit. Set unit max higher for more precision or -1 to use less precise units.

This sentence is hard to understand. Whereas:

> time display accuracy; higher value, more accurate(123 seconds), -1, less accurate (2 minutes)

is easy to understand. To be able to explain complicated matters in laymen's terms is a valued skill.

slook:
Thanks, I changed the arguments of the function to True or False options, considering that logically the only cases where changing the default behaviour would make sense are the 120 seconds and 72 hours cases.

Have you seen PR #3397 yet? There are a lot of string changes/new strings in there. I was thinking that it might be wise to start with a smaller PR like this one to give you all a chance to figure out if any special configuration is needed in Weblate to pickup the gettext calls. Hopefully the context of each of the pluralized string pairs will appear understandably for the translators.

mathiascode:
Could we use this for https://github.com/nicotine-plus/nicotine-plus/discussions/1533#discussioncomment-2011019?

mathiascode:
Remove the keyword arguments, always expand seconds/hours, and reduce the hour threshold from 72 to 48? I don't string concatenation is worth considering here, at least not initially.

I would also consider some kind of rounding, since showing e.g. 3 days left for 3 days and 59 minutes can be misleading.

My main concern with the name was that it could be confused with audio duration, since we use the "duration" term in the GUI. The difference between `human_length` and `human_duration` isn't obvious to me the first time hearing them. It's too late to change the `human_length` name, and I don't know what would be a better name for this new function yet.

mathiascode:
`human_duration_approx()`?

slook:
> I would also consider some kind of rounding, since showing e.g. 3 days left for 3 days and 59 minutes can be misleading.

I don't see how that can be misleading, your example would seem close enough to me?

Although if you meant 3 days left for "3 days and 23 hours" then you would have a point, and that is one reason why I originally chose a 72 hour threshold, since by that point the difference between 3 to 4 days isn't as relevant as it is between 2 to 3. The problem with rounding being baked in is that it would make the function produce an incorrect output in cases where it may be wanted to append a timestamp for example "3 days 23:00:00". As such, I think it is better to always round down rather than up.

PR #3400, opened 2025-05-27
plugin_debugger: Plugin Debugger Settings

slook:
+ Added: Settings for easily toggling each caller on or off as desired

image

### DIFF
--- a/pynicotine/plugins/plugin_debugger/__init__.py
+++ b/pynicotine/plugins/plugin_debugger/__init__.py
@@ -1,4 +1,4 @@
-# COPYRIGHT (C) 2020-2023 Nicotine+ Contributors
+# COPYRIGHT (C) 2020-2025 Nicotine+ Contributors
 # COPYRIGHT (C) 2009 quinox <quinox@users.sf.net>
 #
 # GNU GENERAL PUBLIC LICENSE
@@ -17,6 +17,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from inspect import currentframe
+
 from pynicotine.pluginsystem import BasePlugin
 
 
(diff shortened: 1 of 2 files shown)


pynicotine/plugins/plugin_debugger/__init__.py
```diff
@@ -26,114 +28,131 @@ def __init__(self, *args, **kwargs):
...
+
+            self.settings[method_name] = ("Verbose" not in getsource(method))
+            self.metasettings[method_name] = {
+                "description": f"{method_name}()",
+                "type": "bool"
+            }
+
         self.log("__init__()")
 
+    def _trigger_log(self, parameter_arguments=""):
```
mathiascode, line 47: It should be possible to turn this into a decorator that extracts the function names and arguments automatically.

pynicotine/plugins/plugin_debugger/__init__.py
```diff
@@ -26,114 +28,131 @@ def __init__(self, *args, **kwargs):
 
         super().__init__(*args, **kwargs)
 
+        self.settings = {}
+        self.metasettings = {}
+
+        for method_name, method in getmembers(self, ismethod):
+            if method.__module__ != __name__ or method_name.startswith("_"):
+                # Convenience function from BasePlugin, don't include these
+                continue
```
mathiascode, line 37: `Plugin.__dict__` should give you only the members of the subclass. As a bonus, it will keep the order the methods are defined.

pynicotine/plugins/plugin_debugger/__init__.py
```diff
@@ -26,114 +28,131 @@ def __init__(self, *args, **kwargs):
...
+
+            self.settings[method_name] = ("Verbose" not in getsource(method))
```
mathiascode, line 39: I would prefer a list of method names instead of inspecting the source.

pynicotine/plugins/plugin_debugger/__init__.py
slook: Yes but the more tricky part is binding the live values of each argument. To do that is going to result in quite a lot of code and is beyond my current level of understanding.

PR #3401, opened 2025-05-27
PluginSettings: settings and plugin_name object naming cleanup

slook:
No breaking changes in this PR. It is a minor code cleanup only.

Code cleanups related to logical naming for plugin settings, metasettings and pluginsettings dialog only:

- Renamed: `plugin` -> `plugin_name` where it is a string of the plugin.internal_name and not the plugin object itself
- Renamed: `plugin_id` -> `plugin_name` everywhere
- Renamed: `plugin_name` -> `plugin_human_name` where the string is plugin.human_name and not plugin.internal_name
- Renamed: `name` -> `option_name` where the key is the name of an option and not either of the names of the plugin
- Renamed: `settings` -> `metasettings` where the object is a plugin's metasettings dict() and not a plugin's config settings
- Renamed: `plugin_settings` -> `plugin_settings_dialog` where the widget is a GtkDialog and not plugin.settings
- Renamed: `plugin_settings()` -> `load_plugin_settings()` as the callable is a method and not plugin.settings
- Renamed: `get_plugin_settings` -> `get_plugin_metasettings` as output is plugin.metasettings and not plugin.settings
- Renamed: `on_plugin_settings` -> `on_plugin_metasettings` as the input is plugin.metasettings and not plugin.settings
- Renamed: `on_plugin_settings` -> `on_plugin_settings_button` it is a GtkButton handler and not plugin.settings
- Renamed: `update_settings()` -> `load_options()` as values are read into UI widgets and not written out to the config file
- Renamed: `on_enable_plugin` -> `on_enable_plugin_toggle` is a GtkSwitch handler and not triggered by enable_plugins()

This PR doesn't address why changed plugin settings are not immediately written into the config file. There must be something which is the wrong way around or missing somewhere but I can't work out what it is because the misleading names in the code make it too difficult to follow at the moment. I hope this will help me to clarify things a bit.

After merging then I propose a new function `PluginHandler().save_plugin_settings()` and slight (non-breaking) refactoring of the plugin options loading and plugin settings saving mechanisms, with the aim of simplification, if possible.

No breaking changes in this PR. It is a minor code cleanup only.

### DIFF
--- a/pynicotine/gtkgui/ui/settings/plugin.ui
+++ b/pynicotine/gtkgui/ui/settings/plugin.ui
@@ -57,7 +57,7 @@
                       <object class="GtkSwitch" id="enable_plugins_toggle">
                         <property name="valign">center</property>
                         <property name="visible">True</property>
-                        <signal name="notify::active" handler="on_enable_plugins"/>
+                        <signal name="notify::active" handler="on_toggle_enable_plugins"/>
                       </object>
                     </child>
                   </object>
@@ -131,7 +131,7 @@
                                     <property name="sensitive">False</property>
                                     <property name="tooltip-text" translatable="yes">Settings</property>
                                     <property name="visible">True</property>
-                                    <signal name="clicked" handler="on_plugin_settings"/>
+                                    <signal name="clicked" handler="on_click_plugin_settings"/>
                                     <child>
                                       <object class="GtkBox">
                                         <property name="spacing">6</property>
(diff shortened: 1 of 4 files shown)


slook:
> Call config.write_configuration() in the plugin settings dialog? plugin.settings is already set in config.sections

Yes that is a solution that would work, but what I'd like is a method that allows the config to be written to disk whenever the plugin changes its settings by itself, without depending on the gui. Anyway, whatever we decide to do that's irrelevant for this PR.

pynicotine/gtkgui/ui/settings/plugin.ui
```diff
@@ -131,7 +131,7 @@
                                     <property name="sensitive">False</property>
                                     <property name="tooltip-text" translatable="yes">Settings</property>
                                     <property name="visible">True</property>
-                                    <signal name="clicked" handler="on_plugin_settings"/>
+                                    <signal name="clicked" handler="on_click_plugin_settings"/>
```
mathiascode, line 134: on_show_plugin_settings?

PR #3414, opened 2025-06-09
cli.py: remove "log-message" event when stdout is gone

slook:
- Fixed: Sometimes the terminal console seems dead after the program terminates abnormally

### DIFF
--- a/pynicotine/cli.py
+++ b/pynicotine/cli.py
@@ -16,7 +16,6 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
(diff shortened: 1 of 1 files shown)


mathiascode:
The change makes sense. How can I reproduce the issue?

slook:
Can't remember exactly, sorry. Various errors seemed to cause it, and maybe it's more likely when running under `py-spy`.

I think I noticed it when writing a headless plugin that used `input()` (which isn't possible because `cli.prompt()` is needed instead), then the program hangs and needs to be killed (SIG 9) such that cleanup in quit() isn't done in that scenario.

Issue #3413, opened 2025-06-09
ctrl+PgUp-PgDn


82d:
Hot keys for turning the tabs in the search window like in a browser
This will add consistence and will become more convenient

Nicotine+ 3.4.0.dev1
GTK 3.24.49   •   Python 3.13.3

mathiascode:
Added in https://github.com/nicotine-plus/nicotine-plus/pull/3415

PR #3415, opened 2025-06-09
ctrl+PAGE_UP/PAGE_DOWN for tab cycle

ramseyharrison:
I saw [this issue](https://github.com/nicotine-plus/nicotine-plus/issues/3413) today and figured it was an easy addition. Let me know if what I did was correct (first contribution here).

### DIFF
--- a/pynicotine/gtkgui/application.py
+++ b/pynicotine/gtkgui/application.py
@@ -256,8 +256,8 @@ class Application:
             ("win.show-log-pane", ["<Primary>l"]),
             ("win.reopen-closed-tab", ["<Primary><Shift>t"]),
             ("win.close-tab", ["<Primary>F4", "<Primary>w"]),
-            ("win.cycle-tabs", ["<Control>Tab"]),
-            ("win.cycle-tabs-reverse", ["<Control><Shift>Tab"]),
+            ("win.cycle-tabs", ["<Control>Tab","<Control>Page_Up"]),
+            ("win.cycle-tabs-reverse", ["<Control><Shift>Tab","<Control>Page_Down"]),
 
             # Other accelerators (logic defined elsewhere, actions only used for shortcuts dialog)
             ("accel.cut-clipboard", ["<Primary>x"]),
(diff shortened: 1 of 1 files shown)


slook:
Nice find, but I think something better should be possible.

These shortcuts keys should be natively supported by the Gtk.Notebook widget (what is used for the tabs), so it shouldn't be needed to add them again here. Doing so only avoids the problem without identifying where the block is happening, and it might cause some kind of conflict unless it's solved properly.

It's strange that it works in chats and when the focus is in the search filters, but not when the focus is in the search TreeView widget. It seems it is consuming the keys for horizontal scrolling or something like that.

The same kind of focus trapping also occurs in the "Folders" TreeView widget in Browse Shares, where vertical scrolling is activated but works fine on empty tabs (for example when browsing users who are offline or don't exist). So that needs fixing as well.

Edit: On second thought, it is possible that your solution might be the correct way to override the unwanted scrolling, but I'm not sure.

slook (changes requested):

mathiascode:
This is the correct approach, as we need to apply these shortcuts globally (similar to browsers), not only when a notebook is focused. Trying to deal with conflicting shortcuts on a widget level here is a nightmare anyway.

pynicotine/gtkgui/application.py
```diff
@@ -256,8 +256,8 @@ def _set_up_action_accels(self):
             ("win.show-log-pane", ["<Primary>l"]),
             ("win.reopen-closed-tab", ["<Primary><Shift>t"]),
             ("win.close-tab", ["<Primary>F4", "<Primary>w"]),
-            ("win.cycle-tabs", ["<Control>Tab"]),
-            ("win.cycle-tabs-reverse", ["<Control><Shift>Tab"]),
+            ("win.cycle-tabs", ["<Control>Tab","<Control>Page_Up"]),
```
mathiascode, line 259: Should be page down here to cycle forward.

pynicotine/gtkgui/application.py
```diff
@@ -256,8 +256,8 @@ def _set_up_action_accels(self):
...
+            ("win.cycle-tabs-reverse", ["<Control><Shift>Tab","<Control>Page_Down"]),
```
mathiascode, line 260: Should be page up here to cycle backwards.

mathiascode (approved):
Thanks!

PR #3416, opened 2025-06-10
cli.py: redisplay prompt and input echo while printing log messages

slook:
Experimental draft PR

+ Added: Simultaneously print log output to the console without destroying the command prompt and input echo

Unlike printed log output messages, input prompts don't end with a `\n` newline character until they are consumed from the buffer, which means it is always safe to print a `\r` carriage return character at the start of every log line so that the current prompt and any input text can appear to stay in one piece where it should be, i.e. kept together at the bottom of the typewriter console.

Alternatively, the appropriate number of `\b` backspace characters (depending on the length of the prompt_string + line_buffer) could be printed instead of a carriage return, in order to make sure that a longer input line is cleared before writing over it with a shorter output line.

This method of relocating the last line is the least hacky solution that I have found. It can all be performed without resorting to using escape code sequences or external libraries, some of which may or may not be supported in various consoles on different platforms. However, the readline module isn't the same everywhere, so a more reliable method of peeking the input buffer might be better. For now, this is only good enough for the purpose of seeing what the idea is like.

### DIFF
--- a/pynicotine/cli.py
+++ b/pynicotine/cli.py
@@ -16,6 +16,15 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+try:
+    # Enable line editing and history
+    import readline
+    ENABLE_READLINE = True
+
+except ImportError:
+    # Readline is not available on this OS
+    ENABLE_READLINE = False
+
 import os
 import sys
 import time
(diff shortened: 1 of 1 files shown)


mathiascode:
Do you also have an issue where get_line_buffer() keeps returning the last buffer before the prompt line was cleared (with either the backspace or return key)?

slook:
No, I don't think I have that issue.

mathiascode:
Something strange with my setup. It happens in my main terminal, but not in my toolboxes. I'll merge this, since your implementation looks good.

PR #3438, opened 2025-07-05
shares.py: declare processed_share sets locally

slook:
There is no good reason to declare these class wide, nothing else needs to access them.

+ Changed: declare the `processed_share_names` set within the only method where it used
+ Changed: declare the `processed_share_paths` set within the only method where it used


slook:
Closing, I initially failed to understand that a Scanner() instance runs for each `permission_level`.

PR #3442, opened 2025-07-23
transfers.py: Replace bare key shortcuts with more sensible ones

mathiascode:
We currently have a R shortcut for resuming/retrying transfers, and T shortcut for pausing/aborting transfers. There are two issues with these shortcuts:

- They are bare keys, which conflict with 'type to search' in the transfer list.
- They are right next to each other, which is good for reachability, but makes it too easy to press the wrong key.

I've replaced the shortcuts with Ctrl+S (S as in start) for resuming/retrying transfers, and Ctrl+T (T as in terminate) for pausing/aborting transfers.

The goal is to use shortcuts that are easy to activate with a single hand, and fairly close to each other, but not close enough to cause confusion. We also want to avoid using shortcuts that are very common and have different meanings in other contexts. While Ctrl+S is commonly used for saving files, it seemed appropriate in this context (to resume a transfer to save it to disk).

### DIFF
--- a/pynicotine/gtkgui/application.py
+++ b/pynicotine/gtkgui/application.py
@@ -260,8 +260,8 @@ class Application:
             ("accel.download-to", ["<Primary>Return"]),
             ("accel.file-properties", ["<Alt>Return"]),
             ("accel.back", ["BackSpace"]),
-            ("accel.retry-transfer", ["r"]),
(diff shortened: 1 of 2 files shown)


mathiascode:
Let's give it a shot.

PR #3444, opened 2025-07-23
Replace deprecated GtkShortcutsWindow with custom implementation

mathiascode:
This replaces the final GTK-provided "toy" dialog, which is deprecated and slated for removal in GTK 5.

Some benefits of our custom keyboard shortcuts dialog:

- Accessible for screen reader users in all GTK versions
- Respects header bar preference
- Matches our other dialogs, and can easily be customized in the future
- No need for the old hack to center the dialog

dialog


slook:
Any chance of two columns?

mathiascode:
It might be possible with GtkFlowBox, but if you place each section in a GtkFloxBoxChild, all sections will have the height of the tallest one. Maybe dividing sections between two GtkFlowBoxChild widgets could work.

I don't want to bring back the issues of GtkShortcutsWindow (large static window size, having to carefully craft sections to avoid gaps and making the window too large, slow pagination).

slook:
Good points. I think what you've done looks nice, and like you say it can easily be customized in the future.

slook:
I think it was better with dim ~key~name labels and bright values, i.e. the way around as the statistics in User Profiles.

I also liked the centered section layout where the keys and values are close together because that made it easier to read, albeit I'd understand if you decided against that if it made the overall alignment appear too jumbled.

slook:
Correction: I meant _dim name_ labels and _bright keys_, same as in File Properties or User Profiles.

mathiascode:
I'll try some alignments and post screenshots.

mathiascode:
Here's what the other options look like:
option1
option2

slook:
The centered one, but make the headings centered as well.

mathiascode:
Feels off to me.

shortcuts

slook:
Hmm, yeah its a bit janky, perhaps each section needs delineating in a GtkFrame or other similar outline.

Maybe try that with the fully justified version as well.

mathiascode:
I reduced the size of the window a bit and went with the justified variant. We'll see how that goes.

PR #3455, opened 2025-08-09
iconnotebook.py: add delay before removing tab highlight

### LINKED ISSUE #3440: Scrolling tabs should not change page in focus (v3.3.10)
### Problem of version 3.3.10

When there are more tabs than can be displayed at once, scrolling the tabs by rolling the scroll wheel whilst hovering over tabs scrolls through pages/windows simultaneously. 

In the case of private chat pages this prevents me from reliably accessing a chat page containing new messages from the other party if the corresponding tab isn't visible at that time. This is because as soon as a chat page is displayed, the "new messages" indicator on the tab disappears. Thus, I have no way of knowing which chat actually had those new messages.

This was already a problem in version 3.2.1. The difference to version 3.3.10 (which I've got installed now) is that in version 3.2.1 the scroll wheel had no effect on tabs. Now it does, but the actual problem persists.

Scrolling tabs should leave the displayed page alone. In this way I can bring a tab with a "new messages" indicator into view without the indicator disappearing. In the interest of consistency this behaviour this should apply not only to chat pages/windows, but to all tabbed contexts.

If the tab row or column isn't overflowing (so all tabs are visible), scrolling should have no effect.

### Additional context

N/A

slook:
Fixes #3440

+ Added: Wait to dismiss tab highlight after short delay to allow transient switching such as mouse wheel scrolling

250ms feels about right to avoid the main complaint of the issue which is unintentional clearing of indicators while navigating through overflowed pages using the mouse scroll wheel. It is also effective when tabbing using the keyboard as well.

A much longer delay, such as a few seconds, could be considered so the unread item is actually seen, but some might perceive that as laggy.

### DIFF
--- a/pynicotine/gtkgui/widgets/iconnotebook.py
+++ b/pynicotine/gtkgui/widgets/iconnotebook.py
@@ -552,21 +552,11 @@ class IconNotebook:
 
     def remove_tab_changed(self, page):
 
-        if self.parent_page is not None:
-            if self.window.current_page_id != self.parent_page.id:
-                # Another tab has since been selected within 250ms
-                return
-
-            GLib.source_remove(self.switch_page_delay_timer)
-            self.switch_page_delay_timer = None
-            self._remove_unread_page(page)
-
-        if self.get_current_page() != page:
-            # Another page has since been selected within 250ms
-            return
-
         tab_label = self.get_tab_label(page)
         tab_label.remove_changed()
+
+        if self.parent_page is not None:
+            self._remove_unread_page(page)
 
     def _append_unread_page(self, page, is_important=False):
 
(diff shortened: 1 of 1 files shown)


mathiascode:
Finally got around to testing this, and it feels pretty good.

Could you store the ID returned by GLib.timeout_add() in the class, and cancel the old timeout request ID before the next one, if needed? Just to avoid building up a lot of them when scrolling fast.

It would also be nice if we could avoid these timeout_add calls entirely if a tab isn't highlighted.

slook:
I have made the timer id stored, but I struggled to bypass the call in the case where the tab is not unread because for some reason it caused a failure to clear the unread status from "Public " global feed tab unless it is important/mention.

For now this seems good. You can merge or open to further suggestions if required.

pynicotine/gtkgui/widgets/iconnotebook.py
```diff
@@ -558,12 +552,22 @@ def request_tab_changed(self, page, is_important=False, is_quiet=False):
...
+            GLib.source_remove(self.switch_page_delay_timer)
+            self.switch_page_delay_timer = None
             self._remove_unread_page(page)
 
+        if self.get_current_page() != page:
+            # Another page has since been selected within 250ms
+            return
+
+        tab_label = self.get_tab_label(page)
+        tab_label.remove_changed()
```
mathiascode: Since we track the timer now, you should be able to undo all these changes, and just set `self.switch_page_delay_timer = None` in remove_tab_changed().

PR #3482, opened 2025-09-16
SLSKPROTOCOL.md: update `HaveNoParent` description

fstxz:
When the peer has no parent, the client sends `true`, but "have parents" description reads like it  _does_ have a parent. This PR replaces "have parents" with "no parent", like in [slskmessages.py](https://github.com/nicotine-plus/nicotine-plus/blob/a68f5aa808a3198977b700e6bd61b8465d0196c3/pynicotine/slskmessages.py#L1842)


mathiascode:
Thanks! Can you also change the "last updated" date at the top of SLSKPROTOCOL.md?

fstxz:
@mathiascode done!

PR #3489, opened 2025-09-25
Use os.fsencode() and os.fsdecode() for pathlike byte strings

### LINKED ISSUE #3485: Critical Error (constant crashing)
Nicotine+ version:  

Nicotine+ Version: 3.3.10
GTK Version: 4.16.12
Python Version: 3.12.9 (win32)

Type: <class 'SystemError'>
Value: endswith() argument 1 (impossible<bad format char>)
Traceback:   File "D:/a/nicotine-plus/nicotine-plus/pynicotine/gtkgui/application.py", line 945, in on_process_thread_events
  File "D:/a/nicotine-plus/nicotine-plus/pynicotine/events.py", line 335, in process_thread_events
  File "D:/a/nicotine-plus/nicotine-plus/pynicotine/events.py", line 288, in emit
  File "D:/a/nicotine-plus/nicotine-plus/pynicotine/events.py", line 280, in emit
  File "D:/a/nicotine-plus/nicotine-plus/pynicotine/search.py", line 484, in _file_search_request_distributed
  File "D:/a/nicotine-plus/nicotine-plus/pynicotine/search.py", line 737, in _process_search_request
  File "D:/a/nicotine-plus/nicotine-plus/pynicotine/search.py", line 608, in _create_search_result_list

### Describe the bug
Crashes CONSTANTLY, doesn't stay open

### Expected behavior
crashes, forces restart, repeat

### Additional context

slook:
Closes #3485 (possibly - testing todo)

- https://peps.python.org/pep-0519/ "Adding a file system path protocol" (Python-Version: 3.6)
_" If a str is desired and the encoding of bytes should be assumed to be the default file system encoding, then [os.fsdecode()](https://docs.python.org/3/library/os.html#os.fsdecode) should be used."_

- https://peps.python.org/pep-0529/ "Change Windows filesystem encoding to UTF-8" (Python-Version: 3.6)
_"On Windows bytes cannot round-trip all characters used in paths, as Python internally uses the *A functions and hence the encoding is “whatever the active code page is”. Since the active code page cannot represent all Unicode characters, the conversion of a path into bytes can lose information without warning or any available indication."_

- https://docs.python.org/3/library/sys.html#sys.getfilesystemencoding
_"[os.fsencode()](https://docs.python.org/3/library/os.html#os.fsencode) and [os.fsdecode()](https://docs.python.org/3/library/os.html#os.fsdecode) should be used to ensure that the correct encoding and errors mode are used."_

### DIFF
--- a/pynicotine/shares.py
+++ b/pynicotine/shares.py
@@ -565,7 +565,7 @@ class Scanner:
             try:
                 with os.scandir(encode_path(folder_path, prefix=False)) as entries:
                     for entry in entries:
-                        basename = basename_escaped = entry.name.decode("utf-8", "replace")
+                        basename = basename_escaped = os.fsdecode(entry.name)
 
                         if "\\" in basename:
                             # Substitute backslashes with backslash sentinels in basenames. This is necessary
--- a/pynicotine/utils.py
+++ b/pynicotine/utils.py
@@ -185,7 +185,7 @@ def encode_path(path, prefix=True):
 
         path = LONG_PATH_PREFIX + path
 
-    return path.encode("utf-8")
+    return os.fsencode(path)
 
 
 def human_duration_approx(seconds):


mathiascode:
The largest issue I have with this is that os.fsdecode() uses the 'surrogateescape' error handler, which will likely bring back a bunch of UnicodeEncodeErrors when trying to print those paths or encode them to utf-8.

Especially here, this change would likely crash immediately: https://github.com/nicotine-plus/nicotine-plus/blob/e2a80b357b2177915d979f209990c65dbd0f3c0d/pynicotine/shares.py#L191

https://github.com/pallets/click/issues/2395

PR #3496, opened 2025-09-28
Implement share filters

### LINKED ISSUE #910: Avoid .DS_Store and @eaDir in shares
Hi, could you avoid, at least, .DS_Store (MacOS) files and @eaDir (Synology) directories in shares,
so, they don't appear in the user browse file list.
Also, seen on mac shares, ._file (70 bytes) which seems to be a kind of .DS_Store,
I suppose this is the mac equivalent of windows streams ? no need to share them...
other candidate : Thumbs.db
would be interrested to be configurable at the server side in the Preferences->Transfers->Shares
the same way they are in Downloads->Download filters at the client side.
take care, @eaDir is a full recursive directory to ignore, while the others are simple files.

### Additional context
sample:
slsk://cletusbuckley/%40%40xmnwl/Downloads/VA%20-%20Strange%20Sounds%20Vol.%201/

mathiascode:
Closes #910

### DIFF
--- a/pynicotine/config.py
+++ b/pynicotine/config.py
@@ -184,6 +184,15 @@ class Config:
                 "shared": [],
                 "buddyshared": [],
                 "trustedshared": [],
+                "share_filters": [
+                    ".*",
+                    ".*\\",
+                    "@eaDir\\",
+                    "#recycle\\",
+                    "#snapshot\\",
+                    "desktop.ini",
+                    "Thumbs.db"
+                ],
                 "uploadbandwidth": 50,
                 "use_upload_speed_limit": "unlimited",
                 "uploadlimit": 1000,
--- a/pynicotine/gtkgui/dialogs/preferences.py
+++ b/pynicotine/gtkgui/dialogs/preferences.py
@@ -585,18 +585,26 @@ class SharesPage:
         _("Buddies"): PermissionLevel.BUDDY,
         _("Trusted buddies"): PermissionLevel.TRUSTED
     }
+    FILTER_LEVELS = (
+        _("File-level"),
+        _("Folder-level")
+    )
 
     def __init__(self, application):
 
         (
             self.container,
             self.file_manager_button,
+            self.filter_list_container,
+            self.filter_list_page,
             self.rescan_daily_toggle,
             self.rescan_hour_container,
             self.rescan_on_startup_toggle,
             self.reveal_buddy_shares_toggle,
             self.reveal_trusted_shares_toggle,
-            self.shares_list_container
+            self.shares_list_container,
+            self.shares_list_page,
+            self.stack
         ) = self.widgets = ui.load(scope=self, path="settings/shares.ui")
 
         self.application = application
@@ -605,6 +613,7 @@ class SharesPage:
         self.shared_folders = []
         self.buddy_shared_folders = []
         self.trusted_shared_folders = []
+        self.share_filters = []
 
         items = []
         for hour in range(24):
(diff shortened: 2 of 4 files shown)


pynicotine/gtkgui/icons/hicolor/scalable/actions/mail-send-symbolic.svg
slook, line 1: ?

JRDetwiler:
Hey, thanks so much for adding this feature! I confirmed that a filter of `dont_share\` works for my [use case](https://github.com/nicotine-plus/nicotine-plus/issues/910#issuecomment-3315901849) I described in the issue (i.e. excluding `~/Music/dont_share` and everything inside it).

My main feedback is that the filter syntax was hard to figure out, going in blind from the UI alone. May I suggest two things?

* The filters are quite unforgiving of OS-native glob expressions. For example, I'm on Linux, and I tried out globs (and regexes) like 
  ```
  ./dont_share/*
  ./dont_share
  dont_share\.*
  dont_share\*
  dont_share
  ```
  and so on. None of them filtered out my directory, except for `dont_share\`, and that's not even testing deeper directories, or more liberal pattern matching. Could I suggest [`pathlib`](https://docs.python.org/3/library/pathlib.html) for OS-agnostic path filtering instead of `re`?
* Could the `Add Share Filter` syntax explanation be filled out a bit more to ease users into the feature? I don't think it's necessary to add links about glob syntax, but a few common use case examples that mention the target OS would probably go a long way.

Thanks again!

mathiascode:
I think the expectation of glob-like syntax is what causes the most issues there, it's not something offered right now. We could possibly support a more glob-like syntax, but it depends on how fast the filters will be. Keep in mind that some users share millions of files, which means potentially running the filters on millions of paths.

slook:
A pattern without any backslashes such as `dont_share` should apply to both files and folders where the string appears anywhere in the path name, not only filenames.

JRDetwiler:
> Keep in mind that some users share millions of files, which means potentially running the filters on millions of paths.

That's a really good point, I apologize. `Path.glob` isn't the right tool when taking that into account (I double checked, and `Path.glob` is consistently slower in every benchmark I found, several times over in some). I still think there's value in supporting a more general glob-like filter and I'd be happy to see it in the future, but I wouldn't hold up this PR for it.

Tracerneo:
This does not really fit my usecase, but maybe my suggestion may be a better approach.
I have `~/Music/` and `~/Music/Private/`.
I would like to limit `~/Music/Private/` only to `Trusted buddies`. I added the path to my shares, set it to `Trusted`, but it does not seem to work (does not overwrite `Public` `~/Music/` share).
Maybe implementing it this way would be a better option, and adding a `Hidden/Exclude` as an extra option.

mathiascode:
I've made some changes to hopefully make the filters a bit more straightforward to use:
- UI option to set the filter type (file-level or folder-level)
- Automatically convert path separators to backslashes (always used in Soulseek paths)
- Filter out entire folders if all files inside it were filtered

Let me know how these changes work for you.

mathiascode:
> This does not really fit my usecase, but maybe my suggestion may be a better approach. I have `~/Music/` and `~/Music/Private/`. I would like to limit `~/Music/Private/` only to `Trusted buddies`. I added the path to my shares, set it to `Trusted`, but it does not seem to work (does not overwrite `Public` `~/Music/` share). Maybe implementing it this way would be a better option, and adding a `Hidden/Exclude` as an extra option.

This is more of a separate request, but it might be possible to split out an explicitly shared subfolder and apply its own permission level. I'll have to take a closer look at the code.

slook:
> This is more of a separate request, but it might be possible to split out an explicitly shared subfolder

The suggestion seems to be driving at a good point. Really if a "Trusted" folder happens to be a subfolder of a lower permission level such as "Public", then it is indeed unexpected that it be made available to anyone who is not a Trusted Buddy.

The scanner simply needs to go through the groups backwards starting with the highest "Trusted" level and doing the lowest "Public" level last, since an already scanned folder won't be added twice.

mathiascode:
Merging this now to get some more testing in. We can track the other request in #3545.

PR #3514, opened 2025-10-07
dialogs/download.py: add quick (un)selection of files for download

mathiascode:
Adds context menu items to select/unselect files, and enables multi-selection in the treeview, allowing users to use rubberbanding or keyboard shortcuts to quickly select a group of files for download.

Depends on https://github.com/nicotine-plus/nicotine-plus/pull/3513


mathiascode:
There's an issue where pressing enter only affects the last row of the selection, but that's an issue elsewhere too.

PR #3532, opened 2025-11-02
events.py: append to list of callbacks instead of insert last

### LINKED ISSUE #3531: #73b30d7 breaks the ability to browse user shares
Nicotine+ version:  3.4.0.dev1 (latest master)
Operating System/Distribution: ArchLinux

### Describe the bug
Since 73b30d70eb48802051bf845c785e7c6911c902fc, browsing user shares always returns an empty list, and all users show zero shared files, even though search results still include their files. No such problem before this particular commit.

### Expected behavior
User shares should be visible.

Right click on any search result -> Browse Folder. Alternatively, open any user's profile -> Browse Shares.

### Additional context

slook:
Fixes #3531

### DIFF
--- a/pynicotine/events.py
+++ b/pynicotine/events.py
@@ -250,10 +250,12 @@ class Events:
         if event_name not in EVENT_NAMES:
             raise ValueError(f"Unknown event {event_name}")
 
-        # Event and log modules register callbacks first, but need to quit last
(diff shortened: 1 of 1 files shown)


mathiascode:
Right, -1 actually inserts before the last item. Another thing we have to remember to never use...

PR #3541, opened 2025-11-08
fix crash when closing the last private chat or user profile tab

### LINKED ISSUE #3540: crash after closing last private chat tab or last user profile tab.
Nicotine+ version:  latest from git
Operating System/Distribution: ubuntu-25.04

zniavre reported the following problem on IRC, which he could reproduce by closing the last tab on either private chat or user profile tabs.

```
Nicotine+ Version: 3.4.0.dev1
GTK Version: 3.24.49
Python Version: 3.13.3 (linux)

Type: <class 'AttributeError'>
Value: 'NoneType' object has no attribute 'remove_changed'
Traceback:   File "/home/zniavre/Documents/nicotine-plus/pynicotine/gtkgui/widgets/iconnotebook.py", line 697, in on_remove_tab_changed
    self.remove_tab_changed(new_page)
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/home/zniavre/Documents/nicotine-plus/pynicotine/gtkgui/widgets/iconnotebook.py", line 563, in remove_tab_changed
    tab_label.remove_changed()
    ^^^^^^^^^^^^^^^^^^^^^^^^
```

I couldn't reproduce the problem. But showed the crash output and asked the bot which proposed a few changes, for which I created a PR.

hboetes:
aims to fix #3540

### DIFF
--- a/pynicotine/gtkgui/widgets/iconnotebook.py
+++ b/pynicotine/gtkgui/widgets/iconnotebook.py
@@ -421,6 +421,14 @@ class IconNotebook:
 
     def remove_page(self, page, page_args=None):
 
+        # Cancel any pending delayed tab-remove callback to avoid races
+        if getattr(self, "switch_page_delay_timer", None) is not None:
+            try:
+                GLib.source_remove(self.switch_page_delay_timer)
+            except Exception:
+                pass
+            self.switch_page_delay_timer = None
+
         self.widget.remove_page(self.page_num(page))
         self._remove_unread_page(page)
         self.popup_menu_pages.clear()
@@ -560,6 +568,10 @@ class IconNotebook:
     def remove_tab_changed(self, page):
 
         tab_label = self.get_tab_label(page)
+        if tab_label is None:
+            # Tab was removed while the delayed timer was pending; nothing to do.
+            return
+
         tab_label.remove_changed()
 
         if self.parent_page is not None:


mathiascode:
I would go with this approach instead (there's already a `on_page_removed()` method, you can replace it):

```
    def on_remove_page(self, _notebook, new_page, _page_num):

        if self.switch_page_delay_timer is not None and self.get_n_pages() <= 0:
            GLib.source_remove(self.switch_page_delay_timer)
            self.switch_page_delay_timer = None

        self._remove_unread_page(new_page)
```

This should ensure the timer is removed when the last remaining tab is removed. The code/special case in `remove_tab_changed()` would no longer be needed.

slook:
That condition doesn't look right to me. Surely we want the timer removed for all tab switches and closures, not only if `get_n_pages() <= 0`?

mathiascode:
> That condition doesn't look right to me. Surely we want the timer removed for all tab switches and closures, not only if `get_n_pages() <= 0`?

The only case where the `switch-page` signal isn't emitted is when removing the last tab. If we remove the timer every time a tab is closed, that would break things if you managed to close another tab before the current one is marked as read.

slook:
Right. This is turning out to be a more logically complex mechanism that I had envisaged. Rather than adding more edge case handling workarounds, it would be good if it could somehow be simplified by only having one place where the timer is removed.

hboetes:
Let me play for rubber ducky, @slook 
Can you explain the problem to my simple mind?

slook:
Ideally, resetting the timer would occur as close as possible to the point just after when the wanted tab (if it still exists) is activated, rather than needing to be done or avoided at several other arbitrary places in the module for each calling method.

One simple `is not None` test should suffice to serve this purpose without resorting to trying past an exception because the `self.switch_page_delay_timer` attribute is always guaranteed to exist since it is declared during initialization at... https://github.com/nicotine-plus/nicotine-plus/blob/3d42d8eeab72f000ef439e561145defc6e55720e/pynicotine/gtkgui/widgets/iconnotebook.py#L266 ... even if it doesn't currently happen to contain a timer identifier.

If this cannot be implemented cleanly, i.e. without convoluted hacks for mysterious edge cases, then I'd rather the delayed switch feature, which was added in PR #3455, be reverted.

The original problem for which this feature was originally made is described therein and in issue #3440. Tl/dr; the switch timer was proposed as a solution to a problem that shouldn't exist in the first place, that being the widget lacks a built-in way of navigating or scrolling through tabs when there are more tabs than will fit on the screen, so all the widget can do is step through while activating each of them one-by-one, which sucks.

hboetes:
Just my 2ct, would it be an idea to mark a tab read only after say 2 seconds of being on it?

slook:
Yes it is an idea. A delay of about 3 to 5 seconds is normally expected for a messaging application, that is the default behaviour in Thunderbird for example. I tried 1 second and it just made the interface feel laggy, but a longer time makes it obvious that the delay is intentional.

I also had it consider if the view was scrolled to the bottom or not, since really it is incorrect to mark the tab as read unless and until the last line is actually visible. But I gave up with that idea because it would open up a whole other can of worms.

mathiascode:
Alternatively, we could do something like this:

```
    def on_remove_tab_changed(self, new_page):

        self.switch_page_delay_timer = None

        # Check if tab was removed before timer expired
        if self.page_num(new_page) != -1:
            self.remove_tab_changed(new_page)
```

hboetes:
I can't reproduce the crash and I have no idea what I'm doing. Can I assign the issue/pr to you @mathiascode and @slook ?

mathiascode:
I can commit my latest proposed fix to master. I'd say it's good enough.

mathiascode:
8b1e752bbc49613c88146a827d27975acfc86be1

PR #3551, opened 2025-11-25
Check for timer_id before canceling scheduled events

Diode-exe:
This code makes sure that timer_id exists before cancelling it.


mathiascode:
Is there a specific problem you're trying to solve? This change shouldn't be necessary, since the timers are always created when connecting to the server.

Diode-exe:
Error handling. I thought it was helpful. Sorry.

mathiascode:
No worries!

PR #3568, opened 2025-12-03
preferences.py: add icon column for readability of shared folder

mathiascode:
This mirrors the information in the "Unavailable Shares" dialog when rescanning.

### DIFF
--- a/pynicotine/gtkgui/widgets/theme.py
+++ b/pynicotine/gtkgui/widgets/theme.py
@@ -260,6 +260,9 @@ FILE_TYPE_ICON_LABELS = {
 PRIVATE_ICON_LABELS = {
     "changes-prevent-symbolic": _("Private")
 }
+SHARED_FOLDER_ICON_LABELS = {
+    "dialog-warning-symbolic": _("Unreadable Folder")
+}
 USER_STATUS_ICON_LABELS = {
     "nplus-status-available": _("Online"),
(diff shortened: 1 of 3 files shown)


slook:
I think it looks better with the icon to the left of the associated Folder path that the warning is about, like this:

image

Otherwise in the position you have it by default it looks like the warning is in reference to the user groups. Alternatively, the simplest and most visually appealing position is on the far left at column index 0, so consider swapping the order of the Folder and Virtual Folder columns (afterall, the local Folder path is not Accessible To peers, whereas the Virtual Folder is), like this:

image

Also, the icon column shouldn't be resizable and should be attached to the related Folder column (refer to the Filetype icon in the search results tree).

pynicotine/gtkgui/widgets/theme.py
```diff
@@ -260,6 +260,9 @@ def set_global_style(isolated_mode=False):
 PRIVATE_ICON_LABELS = {
     "changes-prevent-symbolic": _("Private")
 }
+READABLE_ICON_LABELS = {
+    "dialog-warning-symbolic": _("Unreadable")
```
slook: Consider that this same column could be used to indicate the presence of other kinds of errors logged during the actual rescan process of Scanner() and maybe a tick or something for a successfully completed rescan. Suggest call it `FOLDER_ICON_LABELS`.

PR #3597, opened 2026-01-03
Redesign buddy share visibility option in Preferences dialog

mathiascode:
Make it more difficult to enable the option by accident, and provide a description about what it does.

image

image


mathiascode:
Unsure about the wording, but we'll see how it works out for translators.

PR #3599, opened 2026-01-04
Implement chat keyword mentions

### LINKED ISSUE #3484: Allow specifying a list of additional words to highlight
Users should be able to add words to be highlighted in chatroom messages. This could be done through the chats section of the preferences dialog.

I am not familiar with GTK's features (I tried to use Glade to edit the ui file but quickly decided I'd be better off copy-pasting the XML), but if it is possible to put checkboxes in the list of words then they could be used to decide if only whole words should match, or any text (a similar improvement could also be made to the censored words list).

### Additional context
I have already implemented this in my [fork](https://github.com/archiecarrot123/nicotine-plus), checking for any text rather than only whole words. Checking for only whole words would probably be the preferred solution if checkboxes cannot be made, as I have encountered more false positives than whole words only would've given me false negatives.

mathiascode:
Allows you to specify a list of keywords/usernames to be notified about in chats. Messages containing keywords are highlighted.

Closes #3484

### DIFF
--- a/pynicotine/chatrooms.py
+++ b/pynicotine/chatrooms.py
@@ -460,18 +460,41 @@ class ChatRooms:
             self.update_completions()
             core.privatechat.update_completions()
 
-    def get_message_type(self, user, text):
+    def get_message_type(self, username, message):
 
-        if text.startswith("/me "):
+        if message.startswith("/me "):
             return "action"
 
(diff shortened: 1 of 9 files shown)


mathiascode:
Merging this for testing and further feedback. We can make changes later if needed.

PR #3278, opened 2025-02-11
README.md: update main program description for clarity

slook:
It's about time we updated the main description statement, because we're under-selling it at the moment...

- Changed: "aims to be" -> "is" to reflect the progress made in recent years, there is no longer any doubt in this assertion.
- Added: "fully compliant" following the implementation and completion of distributed network capabilities.
- Removed: "comprehensive set of features" as this was too vague and doesn't help search bots nor humans.

Also for consideration, "fast", "performant", "responsive", "efficient" would also be appropriate adjectives that accurately describe the runtime characteristics of the program nowadays. From a user experience standpoint, something like "intuitive" might be inspire confidence if we want to avoid being overly technical?

We should do more to highlight these attributes, since they are things that are important to us and is something users do appreciate also. SEO is important, not sure if this length/number of paragraphs is correct for package distribution purposes?

PR #3615, opened 2026-01-17
Update application description for clarity

mathiascode:
Do a better job at explaining what Nicotine+ does.

Replaces #3278

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,11 +11,13 @@
 Nicotine+ is a graphical client for the [Soulseek](https://www.slsknet.org/news/)
 peer-to-peer network.
 
-Nicotine+ aims to be a lightweight, pleasant, free and open source (FOSS)
-alternative to the official Soulseek client, while also providing a
-comprehensive set of features.
+Share files, chat, and find people with similar interests. Nicotine+
+is user-friendly, fast, free, and open source. It provides features
+and refinements that focus on usability, while remaining fully
+compatible with other Soulseek clients.
 
-Nicotine+ is written in Python and uses GTK for its graphical user interface.
+Nicotine+ is written in Python, and uses GTK for its graphical user
+interface.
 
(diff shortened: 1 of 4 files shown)


slook:
I think we should exchange "graphical" from the headline and put "user-friendly" there instead of in the main body paragraph, because the fact that the program does have a graphical interface is not really a groundbreaking feature anymore and this is also mentioned again in the last paragraph anyway.

Also suggest omitting " and refinements" because this unnecessary adjunct interrupts the flow of the sentence (which would otherwise not need a comma after usability).

mathiascode:
> I think we should exchange "graphical" from the headline and put "user-friendly" there instead of in the main body paragraph, because the fact that the program does have a graphical interface is not really a groundbreaking feature anymore and this is also mentioned again in the last paragraph anyway.

I'd rather not change the headline at this point. It's used in so many other places we don't control.

> Also suggest omitting " and refinements" because this unnecessary adjunct interrupts the flow of the sentence (which would otherwise not need a comma after usability).

Maybe we can remove "additional" instead? I think refinements are worth pointing out, since we put a lot of effort into those aside from features.

> Share files, chat, and connect with people who have similar interests.
> Nicotine+ is fast, user-friendly, free, and open source (FOSS). It provides
> features and refinements focused on usability while remaining fully compatible
> with other Soulseek clients.

mathiascode:
@hboetes @slook Final pass?

> Nicotine+ is a graphical client for the Soulseek peer-to-peer network.
> 
> Share files, chat, and find people with similar interests. Nicotine+
> is user-friendly, fast, free, and open source. It provides features
> and refinements that focus on usability, while remaining fully
> compatible with other Soulseek clients.
> 
> Nicotine+ is written in Python, and uses GTK for its graphical user
> interface.

PR #3618, opened 2026-01-19
Make spam filter properly case-insensitive for phrases

n6id:
added missing `.lower()` for the phrase part of the spam filter check, so that it filters regardless of casing, which seems to have been the intent. this makes the spam filter work for all entered phrases, where previously matches could only be found for phrases which had been entered all-lowercase by the user.


mathiascode:
Thanks!

PR #3673, opened 2026-03-16
Bump minimum Python version to 3.10

ernieIzde8ski:
Python 3.9 is EOL as of October last year, and is no longer supported by Python nor packaged by Ubuntu LTS[^1].

Motivation-wise, I'm interested in adding type hints (#3495, [my fork](https://github.com/ernieIzde8ski/nicotine-plus#typings)). Certain features like `TypeAlias`, the dedicated `A | B` syntax for `Union[A, B]`, and the `slots` param of `dataclass` (if desirable) would be nice to have.

[^1]: v3.10 is supported: <https://packages.ubuntu.com/questing/python3>

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.9', '3.10', '3.11', '3.12', '3.13', '3.14', '3.15-dev',
+        python: ['3.10', '3.11', '3.12', '3.13', '3.14', '3.15-dev',
                  'pypy-3.11']
     steps:
       - name: Checkout code
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -57,7 +57,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        container: ['debian:bullseye', 'debian:bookworm', 'debian:trixie', 'debian:unstable',
+        container: ['debian:bookworm', 'debian:trixie', 'debian:unstable',
                     'ubuntu:jammy', 'ubuntu:noble', 'ubuntu:devel']
     container: ${{ matrix.container }}
     steps:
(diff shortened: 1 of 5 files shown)


mathiascode:
Thanks! I looked into bumping the minimum version the other day, and came to the conclusion that the following actively supported distro versions would be affected:
- Debian 11
- Slackware 15.0

Debian 11 will reach EOL in August, so that shouldn't be a large concern. We don't know when Slackware 16 will be released, but the only Nicotine+ user on Slackware I'm aware of doesn't use it anymore, so it might not be a large concern either.

I would say we can bump the minimum Python version to 3.10. It would also make maintenance a bit easier.

Not sure about dataclasses yet. It will depend on if there's a performance hit or not compared to normal classes.

Note that we will probably not allow AI-generated contributions in the future, so if you intend to use LLMs, please limit them to personal research only (I would recommend avoiding them).

mathiascode:
Can you also bump the minimum Python version in `debian/control` and `pynicotine/__init__.py`? Hopefully I'm not forgetting more locations.

ernieIzde8ski:
Yeah, it might be worth taking a benchmark with/without dataclasses. I suspect it's not too different, especially because of the aforementioned `slots` property, but we'll see if/when I get around to it. Some classes like `Buddy` seem like they'd be great dataclasses if there isn't a performance hit.

Will take a look at those files! I based my original commit off 15ac302; hopefully that does catch all of them.

I feel like my brain rots a little upon accidental LLM exposure, so no issue there.

mathiascode:
Can you also remove Debian 11 (Bullseye) from CI? It's in `.github/workflows/tests.yml`. The Fedora CI failures are unrelated, you can ignore them.

There will probably be more questions about the type hints later, but I'd like them to generally be simple and easily readable whenever possible. Preferably fewer types per variable in general, but multiple are fine when it makes sense.

ernieIzde8ski:
Sure, makes sense

PR #3674, opened 2026-03-17
Add type hints to `core.py`

ernieIzde8ski:
File seems the most trivial. Looks like a good place to start. Two thoughts:

1. Tried to preserve the way modules are only imported as needed by wrapping all types I import in an `if TYPE_CHECKING:` block; is this desirable?
2. In `Core.send_message_to_peer`, is `username` a `str`?

### DIFF
--- a/pynicotine/core.py
+++ b/pynicotine/core.py
@@ -42,6 +42,8 @@ if TYPE_CHECKING:
     from pynicotine.userinfo import UserInfo
     from pynicotine.users import Users
 
+__all__ = ["Core", "UpdateChecker", "core"]
+
 
 class Core:
     """Core handles initialization, quitting, as well as the various components
--- a/pynicotine/core.py
+++ b/pynicotine/core.py
@@ -5,10 +5,19 @@ import os
 import sys
 import threading
 
+from threading import Thread
+from types import FrameType
+from typing import IO
+
 import pynicotine
 from pynicotine.config import config
 from pynicotine.events import events
 from pynicotine.logfacility import log
+from pynicotine.slskmessages import InternalMessage
+from pynicotine.slskmessages import PeerMessage
+from pynicotine.slskmessages import ServerDisconnect
+from pynicotine.slskmessages import ServerMessage
+from pynicotine.slskmessages import ServerReconnect
 
 
 class Core:
@@ -45,13 +54,13 @@ class Core:
         self.update_checker = None
         self._network_thread = None
 
-        self.cli_interface_address = None
-        self.cli_listen_port = None
-        self.cli_rescanning = None
+        self.cli_interface_address: str | None = None
+        self.cli_listen_port: int | None = None
+        self.cli_rescanning: bool | None = None
 
-        self.enabled_components = set()
+        self.enabled_components: set[str] = set()
 
-    def init_components(self, enabled_components=None, isolated_mode=False):
+    def init_components(self, enabled_components: set[str] | None = None, isolated_mode: bool = False):
 
         # Enable all components by default
         if enabled_components is None:
@@ -221,6 +230,17 @@ class Core:
             "version": pynicotine.__version__,
             "status": _("terminating") if signal_type == signal.SIGTERM else _("application closing")
         })
+    def quit(self, signal_type: int | None = None, _frame: FrameType | None = None):
+        log.add(
+            _("Quitting %(program)s %(version)s, %(status)s…"),
+            {
+                "program": pynicotine.__application_name__,
+                "version": pynicotine.__version__,
+                "status": _("terminating")
+                if signal_type == signal.SIGTERM
+                else _("application closing"),
+            },
+        )
 
         # Allow the networking thread to finish up before quitting
         events.emit("schedule-quit")
@@ -258,25 +278,23 @@ class Core:
         ))
 
     def disconnect(self):
-        from pynicotine.slskmessages import ServerDisconnect
         self.send_message_to_network_thread(ServerDisconnect())
 
     def reconnect(self):
-        from pynicotine.slskmessages import ServerReconnect
         self.send_message_to_network_thread(ServerReconnect())
 
-    def _server_reconnect(self, _msg):
+    def _server_reconnect(self, _msg: ServerReconnect):
         self.connect()
 
-    def send_message_to_network_thread(self, message):
+    def send_message_to_network_thread(self, message: InternalMessage):
         """Sends message to the networking thread to inform about something."""
         events.emit("queue-network-message", message)
 
-    def send_message_to_server(self, message):
+    def send_message_to_server(self, message: ServerMessage):
         """Sends message to the server."""
         events.emit("queue-network-message", message)
 
-    def send_message_to_peer(self, username, message):
+    def send_message_to_peer(self, username: str, message: PeerMessage):
         """Sends message to a peer."""
 
         message.username = username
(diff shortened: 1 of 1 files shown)


slook:
> 1. ... wrapping all types I import in an if TYPE_CHECKING: block; is this desirable?

Ideally it would not be so intrusive. Perhaps it would be sufficient that all variables holding reference to these classes are simply assigned as [Callable](https://docs.python.org/3/library/collections.abc.html#collections.abc.Callable) , then all these imports wouldn't be needed.

> 2. In Core.send_message_to_peer, is username a str?

In general our convention is wherever a variable is named `username` then it is a plain `str`, whereas a variable named `user` might sometimes be an object (although, this is not always consistent). Here, it is a `str`.

ernieIzde8ski:
I don't see how the `Callable` approach would work, given that none of the classes I've seen so far implement `__call__` (though they do implement `__new__`, being classes). It's the difference between `shares: Shares` and `shares: type[Shares]`

slook:
Sorry yes you are right I was misunderstanding that of course class definitions are not callable.

I am doubtful about the benefit gained by type checking these particular instances at all, because they are not prone to being wrongly implemented. We should instead focus our efforts on parts of the program where type checking makes the most sense, such as where it improves readability or reduces the risk of breakage.

ernieIzde8ski:
In that case, I think it makes sense to rip out the self-evident attributes from this PR and keep just the mildly ambiguous ones like `cli_interface_address` or `username`.

mathiascode:
> I am doubtful about the benefit gained by type checking these particular instances at all, because they are not prone to being wrongly implemented. We should instead focus our efforts on parts of the program where type checking makes the most sense, such as where it improves readability or reduces the risk of breakage.

I'd say we should eventually strive for complete type coverage, so we can enable strict mode in these type checkers. Otherwise there's the risk that nobody bothers adding type hints when they would make sense, because it's not required.

From previous experience, larger coverage also makes it easier to spot issues across functions/modules.

Additional imports for type hints are fine, they largely end up importing things that would be imported anyway. I don't think we have to use `TYPE_CHECKING` unless there's a problem that requires it.

mathiascode:
@ernieIzde8ski Which type checker are you using? We could add it to our CI.

ernieIzde8ski:
> @ernieIzde8ski Which type checker are you using? We could add it to our CI.

Aside from the ones mentioned in `doc/`, which I use in a `pre-commit` hook, I mostly use a Pyright fork to help figure out types. While I do find it very good for gradually discovering/introducing types, particularly because baselining [^1] suppresses pre-existing type errors, I'm not sure that I would recommend it for normal development here. I find that it assumes much stricter rules than the N+ codebase does [^2], that nearly any patch necessitates either more patches or regenerating the baseline file, and that the hundreds of suppressed little info boxes would probably cause some alert fatigue in the long run.

[^1]: https://docs.basedpyright.com/latest/benefits-over-pyright/baseline/
[^2]: For instance, a frequent "issue" I had with the `Config` class: the programmer knows that properties like `_parser` and `defaults` are initialized when `self.config_loaded is True`, but Pyright (& mypy) don't. In that vein I find a lot of "technically unsafe but obviously safe" code here

pynicotine/core.py
```diff
@@ -265,18 +269,18 @@ def reconnect(self):
         from pynicotine.slskmessages import ServerReconnect
         self.send_message_to_network_thread(ServerReconnect())
 
-    def _server_reconnect(self, _msg):
+    def _server_reconnect(self, _msg: object):
         self.connect()
 
-    def send_message_to_network_thread(self, message):
+    def send_message_to_network_thread(self, message: "InternalMessage"):
```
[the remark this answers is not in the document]
ernieIzde8ski: This crashed on my machine with version 3.10: `NameError: name 'InternalMessage' is not defined`, since it got imported within an `TYPE_CHECKING` block. Curiously, it did _not_ crash with version 3.14.

At any rate, I'm not confident that `TYPE_CHECKING` is useful here. It wouldn't cause any issues to import from `slskmessages`; maybe I'll just drop `TYPE_CHECKING` alongside the quotes?

mathiascode:
I suppose it will take a while before this kind of work stabilizes. I remember I had to wrangle for a while when adding type hints to the [tinytag](https://github.com/tinytag/tinytag) codebase, and it's pretty small.

Let me know if you run into more issues, and I can think about what to do about them. I would only focus on the backend code for now, and leave `gtkgui` for later. The GUI code will be a headache to deal with.

ernieIzde8ski:
I'll keep it in mind, though I do have plenty of work piled up for the next week or two. If I'm lucky, I might manage to sneak in a small patch—`slskmessages.py` looks long yet feasible—but no promises.

In the meantime:

1. Since there had been some debate on typing `Core`'s remaining attributes, I planned to split that commit into a new PR so that that discussion could be held separately. Should I still do that, or do I have the go-ahead to just reintroduce those changes here?

1. Is there anything else left for me to do in this PR?

mathiascode:
> I'll keep it in mind, though I do have plenty of work piled up for the next week or two. If I'm lucky, I might manage to sneak in a small patch—`slskmessages.py` looks long yet feasible—but no promises.

There's no rush. I have plenty of work to during the coming weeks, so I can't guarantee fast reviews right now.

>     1. Since there had been some debate on typing `Core`'s remaining attributes, I planned to split that commit into a new PR so that that discussion could be held separately. Should I still do that, or do I have the go-ahead to just reintroduce those changes here?

I'd say go ahead and introduce them here. We can continue the discussion if necessary.

mathiascode:
@ernieIzde8ski Before you work on `slskmessages.py` (if you do), I'm going to simplify the `unpack_` methods by removing the arguments, and storing them as variables in the message object instead. It's something I've had in mind for a while, since the current argument passing is quite verbose and error-prone.

ernieIzde8ski:
@mathiascode Any updates re: `slskmessages.py`?

mathiascode:
> @mathiascode Any updates re: `slskmessages.py`?

Sorry, I had a branch almost ready, but got distracted by other tasks. I'll open a PR soon.

mathiascode (commented):
Could you revert the formatting changes? We could perhaps think about them in the future, but right now they add more noise to the diff and deviate from our style elsewhere.

pynicotine/core.py
```diff
@@ -23,43 +57,62 @@ class Core:
...
+        self.search: "Search | None" = None
+        self.downloads: "Downloads | None" = None
+        self.uploads: "Uploads | None" = None
+        self.interests: "Interests | None" = None
+        self.userbrowse: "UserBrowse | None" = None
+        self.userinfo: "UserInfo | None" = None
+        self.buddies: "Buddies | None" = None
+        self.privatechat: "PrivateChat | None" = None
+        self.chatrooms: "ChatRooms | None" = None
+        self.pluginhandler: "PluginHandler | None" = None
```
mathiascode: We could use `from __future__ import annotations` to get deferred annotations in Python <3.14. Then we wouldn't have to explicitly specify these as strings.

pynicotine/core.py
```diff
@@ -2,13 +2,47 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import os
+import signal
 import sys
 import threading
+from threading import Thread
```
mathiascode, line 10:
```suggestion
import threading

from threading import Thread
```

pynicotine/core.py
```diff
@@ -2,13 +2,47 @@
...
+from typing import TYPE_CHECKING
 
 import pynicotine
 from pynicotine.config import config
 from pynicotine.events import events
 from pynicotine.logfacility import log
+from pynicotine.notifications import Notifications
+from pynicotine.nowplaying import NowPlaying
+from pynicotine.portchecker import PortChecker
+from pynicotine.portmapper import PortMapper
```
mathiascode: Thinking more about this, it might be a good idea to move these (including NetworkThread) under `if TYPE_CHECKING:`, since they're technically optional components.

ernieIzde8ski:
> Could you revert the formatting changes? We could perhaps think about them in the future, but right now they add more noise to the diff and deviate from our style elsewhere.

Right, I think I squashed a revert commit and didn't realize that had happened, no problem.

mathiascode (approved):

PR #3677, opened 2026-03-19
Use conda-forge instead of Homebrew for dependencies on macOS

mathiascode:
Homebrew has been a source of frustration for a few years, due to its tendency to quickly drop support for EOL macOS versions. It has forced us to drop support for older macOS versions mid-cycle, which is not ideal.

conda-forge recently gained a libadwaita package, and supports macOS >=11. Use it as a source of dependencies instead of Homebrew.

TODO:
- [x] Enable introspection in the libadwaita package on arm64 (https://github.com/conda-forge/libadwaita-feedstock/issues/10)

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -169,30 +169,34 @@ jobs:
           path: build-aux/windows/build/package
 
   macos:
+    timeout-minutes: 15
     strategy:
       matrix:
         include:
           - arch: x86_64
-            platform: macos-15-intel
+            platform: macos-26-intel
           - arch: arm64
-            platform: macos-15
+            platform: macos-26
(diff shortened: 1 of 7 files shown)


slook:
"warning: changes being made to the file will invalidate the code signature in: ..."
https://stackoverflow.com/questions/71744856/install-name-tool-errors-on-arm64

> needed to run `codesign --force -s - /path/to/dylib` on every dylib

mathiascode:
> "warning: changes being made to the file will invalidate the code signature in: ..."

Looks like that warning already occurs in master. I'd say it works as expected, since `replacing existing signature` is printed afterwards. cx_Freeze handles the signing for us.

mathiascode:
I'll backport this to 3.3.x soon.

PR #3681, opened 2026-03-21
split treeview _initialise_columns method into several parts

Kharis228:
Refactor: split TreeView._initialise_columns into smaller helpers

_initialise_columns only orchestrates the loop now; column setup is moved to dedicated methods (_apply_column_iterator_and_sort_defaults, _get_stored_column_config_root, _apply_stored_column_preferences, _create_visible_column_widgets, _setup_column_header_gestures_and_title_widget, _apply_visible_column_layout).

No behavior changes — same order of operations.

### DIFF
--- a/pynicotine/gtkgui/widgets/treeview.py
+++ b/pynicotine/gtkgui/widgets/treeview.py
@@ -242,6 +242,178 @@ class TreeView:
 
         self._column_numbers = list(self._column_ids.values())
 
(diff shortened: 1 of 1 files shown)


mathiascode:
Thanks, but there isn't much point in splitting this method right now, since most of the code will be revamped as part of the Gtk.ColumnView port in Nicotine+ 3.5.x.

PR #3684, opened 2026-03-24
SLSKPROTOCOL.md: add list of reserved client version numbers

slook:
There is ambiguity about the purpose of the "major version" and "minor version" values as to what they are supposed to represent.

We designated these names before so many other third-party clients existed in the wild, but nowadays it needs to be made clearer that project maintainers must avoid impersonating other client implementations.

+ Added: Table of reserved major versions for known clients, developers of new projects need to know this
+ Added: Table of well known client minor versions, these may become helpful for server admins
- Removed: Known clients from the Login() docstring, forcing developers to refer to the full documentation

_Other possible alternatives (request for comments @mathiascode @jpdillingham @JurgenR @bretth18):_
`major_version` -> `client_type    | client_vendor   | client_identifier | client_family ` 
`minor_version` -> `client_subtype | client_variant  | client_revision   | client_generation `

### DIFF
--- a/doc/SLSKPROTOCOL.md
+++ b/doc/SLSKPROTOCOL.md
@@ -5,7 +5,7 @@
 
 # Soulseek Protocol Documentation
 
-[Last updated on April 1, 2026](https://github.com/nicotine-plus/nicotine-plus/commits/master/doc/SLSKPROTOCOL.md)
+[Last updated on April 20, 2026](https://github.com/nicotine-plus/nicotine-plus/commits/master/doc/SLSKPROTOCOL.md)
 
 Since the official Soulseek client and server is proprietary software, this
 documentation has been compiled thanks to years of reverse engineering efforts.
@@ -222,73 +222,77 @@ missing if a file does not provide them.
     -   `{1: duration, 4: sample rate, 5: bit depth}`
 
 
-## Client Types *(Major Versions)*
+## Major Versions
 
-Projects must choose a unique type number to avoid impersonating other clients.
+Established clients have unique numbers to avoid impersonating each other.
 
 ### Reserved
 
-| Type   | Client                                  |
-|--------|-----------------------------------------|
-| `157`  | [Soulseek NS](#soulseek) and SoulseekQt |
-| `158`  | *Reserved for SoulseekQt development*   |
-| `159`  | *Reserved for SoulseekQt development*   |
-| `160`  | [Nicotine+](#nicotine)                  |
-| `169`  | seeleseek                               |
-| `170`  | [Soulseek.NET](#soulseeknet) *(API)*    |
-| `175`  | *Experimental development and testing*  |
-| `178`  | Museek+                                 |
+| Major Version | Client                                  |
+|---------------|-----------------------------------------|
+| `157`         | [Soulseek NS and SoulseekQt](#soulseek) |
+| `158`         | *Reserved for SoulseekQt development*   |
+| `159`         | *Reserved for SoulseekQt development*   |
+| `160`         | [Nicotine+](#nicotine)                  |
+| `169`         | seeleseek                               |
+| `170`         | [Soulseek.NET](#soulseeknet) *(API)*    |
+| `177`         | *Experimental development and testing*  |
 
 ### Obsolete
 
-| Type   | Client                    |
-|--------|---------------------------|
-| `156`  | Soulseek 156 *(pre 2008)* |
-| `180`  | PySoulSeek                |
-| `181`  | Nicotine+ 1.2.0 to 1.2.11 |
-| `182`  | Museek+                   |
-| `198`  | Nicotine 1.0.6 to 1.0.8   |
-| `200`  | Nicotine 1.0.4 and below  |
+| Major Version | Client                     |
+|---------------|----------------------------|
+| `139`...`155` | Soulseek® *(2002 onwards)* |
+| `156`         | Soulseek® *(2008)*         |
+| `178`         | Museek+                    |
+| `180`         | PySoulSeek                 |
+| `181`         | Nicotine+ 1.2.0 to 1.2.11  |
+| `182`         | Museek+                    |
+| `198`         | Nicotine 1.0.6 to 1.0.8    |
+| `200`         | Nicotine 1.0.4 and below   |
 
 
-## Client Subtypes *(Minor Versions)*
+## Minor Versions
 
-Projects have thier own rules for subtypes. The number can be set to anything.
+Projects have thier own rules for versioning. The number can be any **uint32**.
 
 ### Nicotine+
 
-The project maintainer increments the subtype number for new capabilities.
+The project maintainer increments the minor version when adding support for
+significant network capabilities.
 
-#### Client type `160`
+#### Major Version `160`
 
-| Subtype | Versions        |
-|---------|-----------------|
-| `1`     | 2.2.1 to 3.2.9  |
-| `2`     | 3.3.0 to 3.3.10 |
-| `3`     | 3.4.0 and above |
+| Minor Version | Release Versions |
+|---------------|------------------|
+| `1`           | 2.2.1 to 3.2.9   |
+| `2`           | 3.3.0 to 3.3.11  |
+| `3`           | 3.4.0 and above  |
 
 ### Soulseek.NET
 
-API library requires clients to choose a `minorVersion` as their subtype.
+API library requires each downstream project to choose a unique `minorVersion`.
 
-#### Client type `170`
+#### Major Version `170`
 
-| Subtype | Client                                    |
-|---------|-------------------------------------------|
-| `100`   | *All legacy clients up to API 8.5.0*      |
-| `760`   | slskd *(Reference implementation of API)* |
-| `9999`  | *Experimental development and testing*    |
+| Minor Version | Client                                    |
+|---------------|-------------------------------------------|
+| `100`         | *All legacy clients up to API 8.5.0*      |
+| `760`         | slskd *(Reference implementation of API)* |
+| `9999`        | *Experimental development and testing*    |
 
 ### Soulseek®
 
-Official client subtype number is incremented for each released version.
+Official client minor version number is incremented for each release.
 
-#### Client type `157`
+#### Major Version `157`
 
-| Subtype             | Version |
-|---------------------|---------|
-| `17` (`0x11000000`) | NS 13c  |
-| `19` (`0x13000000`) | NS 13e  |
+| Minor Version | Release                    |
+|---------------|----------------------------|
+| `17` (`0x11`) | NS 13c                     |
+| `19` (`0x13`) | NS 13e                     |
+| ...           | Qt Public Build 1 *(2011)* |
+|               | Qt *(2012 onwards)*        |
 
 
 # Server Messages
@@ -297,8 +301,8 @@ Server messages are used by clients to interface with the server over a
 connection (TCP). In Nicotine+, these messages are defined in slskmessages.py.
 
 If you want a Soulseek server, [Soulfind](https://github.com/soulfind-dev/soulfind)
-is open source and not exactly the same as the official proprietary Soulseek
-server, but it handles the protocol well enough for development and testing.
+is open source (unrelated to the official proprietary Soulseek server). It handles
+the protocol well enough for development and testing of client implementations.
 
 
 ## Server Message Format
@@ -423,10 +427,11 @@ server, but it handles the protocol well enough for development and testing.
 We send this to the server right after the connection has been established.
 Server responds with the greeting message.
 
-The server may use the client type and subtype (formally known here as
-the major and minor protocol versions) to differentiate between client
-implementations. Projects must choose a unique type number to avoid
-impersonating other clients, and have thier own rules for subtypes.
+The server uses the major and minor versions to differentiate between
+clients. Numbers are chosen that avoid impersonating clients with reserved
+[Major Versions](#major-versions). Downstream projects have their own rules
+for [Minor Versions](#minor-versions). Experimental scripts may use major
+version `177` and any minor version number they choose for each project.
 
 ### Sending Login Example
 
@@ -439,15 +444,15 @@ impersonating other clients, and have thier own rules for subtypes.
 
 #### ...continued
 
-| Data      | Client Type   | Hash Length   | Hash of Username + Password                                                                       | Client Subtype |
-|-----------|---------------|---------------|---------------------------------------------------------------------------------------------------|----------------|
-| **Human** | 175           | 32            | d51c9a7e9353746a6020f9602d452929                                                                  | 1              |
-| **Hex**   | `af 00 00 00` | `20 00 00 00` | `64 35 31 63 39 61 37 65 39 33 35 33 37 34 36 61 36 30 32 30 66 39 36 30 32 64 34 35 32 39 32 39` | `01 00 00 00`  |
+| Data      | Major Version | Hash Length   | Hash of Username + Password                                                                       | Minor Version |
+|-----------|---------------|---------------|---------------------------------------------------------------------------------------------------|---------------|
+| **Human** | 177           | 32            | d51c9a7e9353746a6020f9602d452929                                                                  | 1             |
+| **Hex**   | `b1 00 00 00` | `20 00 00 00` | `64 35 31 63 39 61 37 65 39 33 35 33 37 34 36 61 36 30 32 30 66 39 36 30 32 64 34 35 32 39 32 39` | `01 00 00 00` |
 
 #### Message as Hex Stream
 
 ```
-48 00 00 00 01 00 00 00 08 00 00 00 75 73 65 72 6e 61 6d 65 08 00 00 00 70 61 73 73 77 6f 72 64 af 00 00 00 20 00
+48 00 00 00 01 00 00 00 08 00 00 00 75 73 65 72 6e 61 6d 65 08 00 00 00 70 61 73 73 77 6f 72 64 b1 00 00 00 20 00
 00 00 64 35 31 63 39 61 37 65 39 33 35 33 37 34 36 61 36 30 32 30 66 39 36 30 32 64 34 35 32 39 32 39 01 00 00 00
 ```
 
@@ -456,12 +461,12 @@ impersonating other clients, and have thier own rules for subtypes.
     1.  **string** *username*
     2.  **string** *password*  
         A non-empty string is required
-    3.  **uint32** *client type*
-        See [Client Types](#client-types-major-versions)
-    4.  **string** *hash*
+    3.  **uint32** *major version*  
+        See [Major Versions](#major-versions)
+    4.  **string** *hash*  
         MD5 hex digest of concatenated username and password
-    5.  **uint32** *client subtype*
-        See [Client Subtypes](#client-subtypes-minor-versions)
+    5.  **uint32** *minor version*  
+        See [Minor Versions](#minor-versions)
   - Receive
     1.  **bool** *success*
     2.  If *success* is true
--- a/doc/SLSKPROTOCOL.md
+++ b/doc/SLSKPROTOCOL.md
@@ -5,7 +5,7 @@
 
 # Soulseek Protocol Documentation
 
-[Last updated on March 22, 2026](https://github.com/nicotine-plus/nicotine-plus/commits/master/doc/SLSKPROTOCOL.md)
+[Last updated on March 24, 2026](https://github.com/nicotine-plus/nicotine-plus/commits/master/doc/SLSKPROTOCOL.md)
 
 Since the official Soulseek client and server is proprietary software, this
 documentation has been compiled thanks to years of reverse engineering efforts.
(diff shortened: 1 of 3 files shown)


JurgenR:
> There is ambiguity about the purpose of the "major version" and "minor version" values as to what they are supposed to represent.

There's still some ambiguity, as it's not clear what those numbers represent or what impact they actually have on the server side. What numbers are allowed, how would I pick one? Is it just about setting a standard or is the goal to allow for different feature sets for different clients?

slook:
Thank you for your input those are good questions which give me a good idea about what kind of information could be sought from this piece documentation. Unfortunately I do not know all the answers for certain, and this document probably isn't the right place to publish some of the internal knowledge at this time.

> what impact they actually have on the server side

It is unknown exactly what happens server-side so speculation cannot be offered in our documentation, since we only include facts about client-side implementation details we know for sure are true.

There is no evidence that the server currently uses the versions for anything other than checking the number is not less than the version used by the current official client, which was a feature used to force user migration when the New Server (NS) was deployed and a new version of the client software was required due to changes made to the protocol after the old server was decommissioned.

Conceptually the same feature could be used again for similar reasons or for other purposes by the server admins such as creating firewall rules or for load-balancing but this doesn't seem to be an option for them at the present time.

The overall thinking is that different types of clients who currently impersonate eachother could be having a negative impact on the server, both in terms of security and performance, so spreading out the third-party clients onto unique numbers in a more logical fashion is intended to have a positive impact on the server.

> What numbers are allowed, how would I pick one?

As it currently stands, any major version from `157` to `177` seems to work fine, but I have not tried other numbers from different client types, maybe somebody else can confirm this, but I don't think we'll publish the knowledge about the working range in the document. It's better for this to stay a mystery because the choice is so limited, and only serious projects who aim to be fully compliant with the protocol ought be prioritized.

The **major client version type** number can be "reserved" by simply implementing it in your public repo. Then I am proposing we go one step further, and publish the major clients who are known in The document, since there is no formal approval process.

Nobody enforces any rules so it's basically just a free-for-all until they are all used, and there is no centralized record in existence so it's impossible to know if multiple clients in the wild might already be overlapping.

It seems polite to leave `158` and `159` available for use by possible future major versions of the official client, even though it has stuck to `157` since forever, nothing is ever certain. In this PR you'll be able to see what's available to pick one hopefully without cramming them in too closely or treading on anyone else's toes.

If you want to pick `175` for your project then we can amend our [Sending Login Example](https://nicotine-plus.org/doc/SLSKPROTOCOL.html#sending-login-example) (see https://github.com/nicotine-plus/nicotine-plus/commit/cc1d09905ab1b474d75701c17ee55c1f1cc2f497) if you can suggest a better number for us to reserve for "development and testing" purposes (I suggest the highest allowed value would be better for experimental use). 

Otherwise, you may pick whatever you would like your major version type to be, and if you ever need to change your mind for any reason from time to time then that's okay as needed to coordinate between us, but in general client types are expected to keep using the same major version value during its entire lifetime.

--

The **minor client version subtype** can be anything, provided that the data payload fits within the field. Relevant for API libraries is has been proposed that projects who are dependent upon a library may choose only the minor version subtype for themselves _(suggest perhaps even use a number derived from their application name/version or such to make the first two digits of the number unique to each project)_, but not be able to easily choose a different major version type other than the one which has been hard-coded into the library which they depend on (unless they fork the library, of course).

[This is what Soulseek.NET has done since their API version 9](https://github.com/jpdillingham/Soulseek.NET#about-minor-versions), and this idea makes sense because otherwise there wouldn't be enough major version type numbers to go around for all the developers using libraries who wish to write a downstream client implementation.

There are no set expectations as to how frequently or infrequently minor client version subtypes shall be changed, unless the API projects wish to enforce their own rules about what minor numbers downstream clients are allowed to use, otherwise it is realistic to expect each project has full responsibility for making their own rules for this.

> Is it just about setting a standard...

Well yes, between us it seems we are writing the book on it! The consensus that exists between our documentation is what sets the standard and improves understanding about the nature of the protocol and its usage, since [your docs](http://aioslsk.readthedocs.io/) are comprehensive and [our docs](https://nicotine-plus.org/doc/SLSKPROTOCOL.html) are authoritative, so this is already a healthy combo which makes everything verifiable for the developers who use these important references.

If the [aioslsk](https://github.com/JurgenR/aioslsk) library decides to implement a similar numbering strategy to align with the [Soulseek.NET](https://github.com/jpdillingham/Soulseek.NET) library then it will become a defacto standard for the rules about numbering API libraries and the projects based on them, which would mean this decision should then be written down so this numbering strategy will be adopted by all future API libraries as well.

If that proposal becomes widely adopted, then the "version" term shall seem inappropriate because each number will actually represent a totally different "type" of client rather than indicating any sort of sequential progression.

The terminology used for this ("version" vs "type" vs "variant" and so on) is of course up for debate, and it's understandable that stable API's cannot change this easily. However, it is confusing that these values are unrelated to the actual software version and if the "major version" never changes during the lifetime of a project due to the modern usage of this parameter then its name will obviously become outdated at some point. Future API library versions/types and third-part clients will eventually adopt whatever agreed terminology we choose to use in our documentation, and that will help the concept and usage be properly understood and properly implemented by both core developers like us as well as developers using libraries for their projects.

We don't know what any of the internal names really are, rather our developers simply invented all of the phrases based on observing the pattern of different codes and documented them to the best of their understanding at the time. As such, anything in the published documentation is always subject to amendment as we see fit.

> ... or is the goal to allow for different feature sets for different clients?

No, there isn't a goal nor any known mechanism to allow or restrict clients from having varied feature sets based on version numbers. What the server admins want to do with their network is obviously entirely different matter, but if they ever do decide to impose restrictions on buggy or malicious clients then attempting to evade them by using different version numbers would be the incorrect approach for developers to take.

This part of the protocol requires everyone involved to make some plan for the sensible allocation of the numbers otherwise it will just be chaos!

I hope that is useful information for you and your project, sorry it is long. If you have any further queries or suggestions for improving our documentation then your continued input on any related matter would be greatly appreciated.

mathiascode:
The main reason to use a unique version number is to preserve the possibility for Soulseek's owner to prevent certain client versions from connecting to the server. It would also ensure the possibility to exclude certain client versions from specific functionality, e.g. the distributed network, if it turns out the implementation was buggy later. Right now it's not being done, but we have to act in good faith and leave these possibilities open.

As for why I changed the example major version to 175 in the docs, it was mainly to prevent naive AI implementations from blindly copying Nicotine+'s version number. It's not great to have possibly buggy, unfinished client implementations identifying as us.

Why not just use any number when it comes to the major version? Because last I checked, only clients with a version number between 157 and 178 are sent the list of possible distributed parents from the server.

mathiascode:
At some point I might ask if the major version range restriction can be lifted, but I'd say it's still better to use the minor version more liberally when possible, and try to avoid using endless major versions. Especially since that's the direction the protocol has been going over the years.

slook:
> clients with a version number between 157 and 178 are sent the list of possible distributed parents

I can't get major_version `178` to connect from Nicotine+ because it causes the following error in the program:

```
[Misc] Unable to parse server message type <class 'pynicotine.slskmessages.Login'>, size 14, contents <memory at 0x7fe76b148b80>. Error: unpack_from requires a buffer of at least 13 bytes for unpacking 4 bytes at offset 9 (actual buffer size is 10)
```

> the major version range restriction ... since that's the direction the protocol has been going over the years.

At first the range was restricted to exclude the obsolete clients (Soulseek 156 and pySoulSeek/Nicotine on version 180/181) when the old server went offline, and after Nicotine+ took 160 this has screwed the official client down below that number.

So it looks like the N/N+ project is what has led to the range being clamped to those numbers in the first place which has somewhat forced this direction by effectively repurposing the major_version into a client_type, thereby creating the need for NS to add the minor_version field (or whatever they call it) into the protocol.

mathiascode:
> > clients with a version number between 157 and 178 are sent the list of possible distributed parents
> 
> I can't get major_version `178` to connect from Nicotine+ because it causes the following error in the program:
> 
> ```
> [Misc] Unable to parse server message type <class 'pynicotine.slskmessages.Login'>, size 14, contents <memory at 0x7fe76b148b80>. Error: unpack_from requires a buffer of at least 13 bytes for unpacking 4 bytes at offset 9 (actual buffer size is 10)
> ```

Looks like the server doesn't send some attributes in the login response when the version is >= 178.

> > the major version range restriction ... since that's the direction the protocol has been going over the years.
> 
> At first the range was restricted to exclude the obsolete clients (Soulseek 156 and pySoulSeek/Nicotine on version 180/181) when the old server went offline, and after Nicotine+ took 160 this has screwed the official client down below that number.
> 
> So it looks like the N/N+ project is what has led to the range being clamped to those numbers in the first place which has somewhat forced this direction by effectively repurposing the major_version into a client_type, thereby creating the need for NS to add the minor_version field (or whatever they call it) into the protocol.

SoulseekQt deciding to continue using 157 as the version happened long before Nicotine+ adopted 160 as its version. Nicotine+ used 157 since 2009, until I decided to change it to 160 a few years ago.

There was also a time where only whitelisted clients were allowed to participate in the search network:
https://web.archive.org/web/20140625180615/http://chris.schleifer.net/ssX/
https://github.com/nicotine-plus/nicotine-plus/commit/78ddb40efaa2eb98f08c729513484d604bba5274#diff-f8c18ce887a6c7908f93a52c7747beafc9f23d80cb14bd3fe13f38baebaa16a1R466
https://github.com/nicotine-plus/nicotine-plus/commit/89b6cbaf41218fdf8d9d8502350041a5ff6507af#diff-f8c18ce887a6c7908f93a52c7747beafc9f23d80cb14bd3fe13f38baebaa16a1R469-R1366
https://github.com/nicotine-plus/nicotine-plus/commit/de96913abceff34988e7983aeb237b1b8d8e4438#diff-f8c18ce887a6c7908f93a52c7747beafc9f23d80cb14bd3fe13f38baebaa16a1R521-R524

Perhaps what's left on the server today is a remnant of this?

mathiascode:
The PR generally looks good, but I'd keep the major and minor version terminology for now.

mathiascode:
> Perhaps what's left on the server today is a remnant of this?

My guess would be that this is the case, and that the version enforcement was mainly sierracat's doing (a Soulseek admin [until ~2008](https://en.wikipedia.org/wiki/Soulseek#Financing)). He also created soulseeX, which was the only whitelisted third-party client on the search network [in 2005](https://web.archive.org/web/20140625180615/http://chris.schleifer.net/ssX/), and judging by the old Nicotine+ commits going back and forth on the version number, at least until 2007-2008.

The server in general seemingly hasn't been touched much since then, considering if you try to join the "dev" room, it still tells you to contact sierracat for access.

This is mostly speculation, though. I'm sure there were some forum posts related to this back then, but I'm not sure if they're archived.

slook:
> Looks like the server doesn't send some attributes in the login response when the version is >= 178.

Well then I suggest we make the "Experimental testing and development" number `177` for use in the Login example, since that's currently at the farthest extent of the allowable working range, and I think [aioslsk](https://github.com/JurgenR/aioslsk) might like to use `175`.

> This is mostly speculation, though... I'm not sure if they're archived.

That's why I feel it's important to record everything we can find so that these historic data are not lost.

doc/SLSKPROTOCOL.md
```diff
@@ -222,14 +222,87 @@ missing if a file does not provide them.
...
+| `169`         | seeleseek                               |
+| `170`         | [Soulseek.NET](#soulseeknet) *(API)*    |
+| `177`         | *Experimental development and testing*  |
+
+### Obsolete
+
+| Major Version | Client                     |
+|---------------|----------------------------|
+| `139`...`155` | Soulseek® *(2002 onwards)* |
+| `156`         | Soulseek® *(2008)*         |
```
mathiascode: 2005 https://web.archive.org/web/20050401042245/http://www.slsknet.org/download.html

I don't think this needs to be a separate entry though.

doc/SLSKPROTOCOL.md
```diff
@@ -222,14 +222,87 @@ missing if a file does not provide them.
...
+| `160`         | [Nicotine+](#nicotine)                  |
+| `169`         | seeleseek                               |
+| `170`         | [Soulseek.NET](#soulseeknet) *(API)*    |
+| `177`         | *Experimental development and testing*  |
+
+### Obsolete
+
+| Major Version | Client                     |
+|---------------|----------------------------|
+| `139`...`155` | Soulseek® *(2002 onwards)* |
```
mathiascode: Is the starting version verified?

doc/SLSKPROTOCOL.md
```diff
@@ -222,14 +222,87 @@ missing if a file does not provide them.
...
+
+| Minor Version | Client                                    |
+|---------------|-------------------------------------------|
+| `100`         | *All legacy clients up to API 8.5.0*      |
+| `760`         | slskd *(Reference implementation of API)* |
+| `9999`        | *Experimental development and testing*    |
+
+### Soulseek®
+
+Official client minor version number is incremented for each release.
```
mathiascode: Did you verify that every release increments it?

doc/SLSKPROTOCOL.md
```diff
@@ -222,14 +222,87 @@ missing if a file does not provide them.
...
+
+API library requires each downstream project to choose a unique `minorVersion`.
+
+#### Major Version `170`
+
+| Minor Version | Client                                    |
+|---------------|-------------------------------------------|
+| `100`         | *All legacy clients up to API 8.5.0*      |
+| `760`         | slskd *(Reference implementation of API)* |
+| `9999`        | *Experimental development and testing*    |
```
mathiascode: I'd rather not document Soulseek.NET's minor versions here. I think it's better suited for the Soulseek.NET repo.

doc/SLSKPROTOCOL.md
```diff
@@ -222,14 +222,87 @@ missing if a file does not provide them.
...
+| `180`         | PySoulSeek                 |
+| `181`         | Nicotine+ 1.2.0 to 1.2.11  |
+| `182`         | Museek+                    |
+| `198`         | Nicotine 1.0.6 to 1.0.8    |
+| `200`         | Nicotine 1.0.4 and below   |
+
+
+## Minor Versions
+
+Projects have thier own rules for versioning. The number can be any **uint32**.
```
mathiascode:
```suggestion
Projects have their own rules for versioning. The number can be any **uint32**.
```

mathiascode:
> There is no evidence that the server currently uses the versions for anything other than checking the number is not less than the version used by the current official client, which was a feature used to force user migration when the New Server (NS) was deployed and a new version of the client software was required due to changes made to the protocol after the old server was decommissioned.

The server sets a minimum version allowed to connect, but I don't think it was used for the NS migration. Soulseek 156 continued working for several years on the old server, until the old server was shut down completely due to a low number of users (see https://web.archive.org/web/20120911182903/http://www.slsknet.org/news/faq-page#t10n761).

The version check existed before the new server, judging by old news posts: https://web.archive.org/web/20050330041814/http://www.slsknet.org/news/

You can also see it in action if you use major version 157 and a minor version between 0 and 11. The server rejects the login with an `INVALIDVERSION` reason. The version requirement hasn't changed since the early Soulseek NS days.

mathiascode:
Looks like the minor version can't be zero, otherwise the server puts you in the same restricted mode as major version <=156 (except it doesn't send a PM about an outdated client version).

doc/SLSKPROTOCOL.md
slook: Not exactly, no. This is my best estimate from the limited information that is available.

I think the earliest known/publicly released beta version was something like 121 or 129 in around 2000/2001 but I haven't been able to locate the resource where I originally thought I saw that.

Since the earliest data are not known for sure, I have renamed the client description to mention "up to 2004".

doc/SLSKPROTOCOL.md
slook: That certainly does appear to be the case from the information I have seen about it, but I haven't gone into any specific effort to verify it, no.

I have changed the wording "is" -> "seems to be".

doc/SLSKPROTOCOL.md
slook: These are the well known ones referencing the official upstream project, so these data are worth documenting here. When downstream ones are listed then I plan to add into or below the table the hyperlink:
https://github.com/jpdillingham/Soulseek.NET#reserved-minor-version-ranges

It being partially documented and referenced here serves as a useful example about the versioning strategy that a public API library might use as opposed to a sequential scheme of a base client implementation.

doc/SLSKPROTOCOL.md
slook: I renamed it and moved into the reserved table, because the number is still allowed so not technically obsolete.

doc/SLSKPROTOCOL.md
```diff
@@ -222,14 +222,93 @@ missing if a file does not provide them.
...
+
+Established clients have unique numbers to avoid impersonating each other.
+Obsolete clients cannot [Login](#server-code-1) to the server since Feb 17,
+2005 due to the `INVALIDVERSION` [Rejection Reason](#login-rejection-reasons).
+
+### Reserved
+
+| Major Version | Client                                  |
+|---------------|-----------------------------------------|
+| `155` / `156` | [Soulseek®](#soulseek) *(2005 to 2008)* |
```
mathiascode: These should be in the obsolete section. The server sends a PM about upgrading your client if the major version is <= 156, and doesn't allow you to do anything on the server.

doc/SLSKPROTOCOL.md
```diff
@@ -222,14 +222,93 @@ missing if a file does not provide them.
...
+| `182`         | Museek+                   |
+| `198`         | Nicotine 1.0.6 to 1.0.8   |
+| `200`         | Nicotine 1.0.4 and below  |
+
+
+## Minor Versions
+
+Projects have their own rules for versioning. Any **uint32** number may be
+chosen, but if it is `0` (zero; or if nothing is sent) then the client cannot
+participate in the [distributed search network](#distributed-messages).
```
mathiascode: The server doesn't allow you to do anything at all. It doesn't read any messages you send after the login message. It's the same mode as when you use major version <= 156, except the server doesn't send you a PM about your client being outdated.

I would just mention that zero is not allowed. I plan on making Soulfind reject your login with an `INVALIDVERSION` reason if you use it (it's not identical to the official server, but close enough).

doc/SLSKPROTOCOL.md
```diff
@@ -222,14 +222,93 @@ missing if a file does not provide them.
     -   `{1: duration, 4: sample rate, 5: bit depth}`
 
 
+## Major Versions
+
+Established clients have unique numbers to avoid impersonating each other.
+Obsolete clients cannot [Login](#server-code-1) to the server since Feb 17,
+2005 due to the `INVALIDVERSION` [Rejection Reason](#login-rejection-reasons).
```
mathiascode: `INVALIDVERSION` is only used with major version 157 and minor version 0-11, as far as I can tell. 156 and below let you log in, but send a PM about upgrading your client and don't allow you to interact with the server.

I would drop the date and just mention that 156 and below are no longer allowed to connect to the server, for simplicity. Alternatively mention it in the table.

doc/SLSKPROTOCOL.md
mathiascode: The PM is the following:
`Your connection is restricted: You cannot search or chat. Your client version is too old. You need to upgrade to the latest version. Close this client, download new version from http://www.slsknet.org, install it and reconnect.`

doc/SLSKPROTOCOL.md
slook: I think we should just leave some of the specifics as an exercise for the reader to discover by themselves. Basically I just want it to give the overall impression that major versions are tightly regulated while minor versions are freely usable to a reasonable extent as to not involve a random number generator for each instance of a client.

doc/SLSKPROTOCOL.md
slook: It would be useful to have somewhere a consolidated list of all known server PM's.

mathiascode:
I verified which minor versions SoulseekQt uses, and it's surprisingly few:

- 19 (same as NS 13e): SoulseekQt Public Build 1 to 2
- 100: SoulseekQt Public Build 3 to 2023-1-15
- 101: SoulseekQt 2024-2-1 and newer

doc/SLSKPROTOCOL.md
mathiascode: > It would be useful to have somewhere a consolidated list of all known server PM's.

I agree. We could create one at some point.

doc/SLSKPROTOCOL.md
```diff
@@ -222,14 +222,91 @@ missing if a file does not provide them.
...
+| `170`         | [Soulseek.NET](#soulseeknet) *(C# API)* |
+| `175`         | aioslsk *(Python API)*                  |
+| `177`         | *Experimental development and testing*  |
+
+### Obsolete
+
+| Major Version | Client                    |
+|---------------|---------------------------|
+| `154` or less | Soulseek® *(up to 2004)*  |
+| `155` / `156` | Soulseek® *(2005 - 2008)* |
```
mathiascode:
```suggestion
| `156` or less | Soulseek® *(pre-NS)*  |
```

doc/SLSKPROTOCOL.md
```diff
@@ -222,14 +222,91 @@ missing if a file does not provide them.
...
+### Soulseek®
+
+Official client minor version only increments occasionally. It may signify
+different server functionality.
+
+#### Major Version `157`
+
+| Minor Version | Release                     | Date                   |
+|---------------|-----------------------------|------------------------|
+| `0` to `11`   | NS Alpha (`INVALIDVERSION`) |                        |
```
mathiascode:
```suggestion
| `1` to `11`   | NS Test (`INVALIDVERSION`) |                        |
```

I checked some old versions, and none of them start from 0.

doc/SLSKPROTOCOL.md
```diff
@@ -222,14 +222,91 @@ missing if a file does not provide them.
...
+| `17` (`0x11`) | NS 13c                      |                        |
+| `19` (`0x13`) | NS 13e                      |                        |
+| `19`          | Qt Public Build 1 and 2     | *from 2011-04-19*      |
+| `100`         | Qt Public Build 3           | *to 2023-01-15*        |
+| `101`         | Qt                          | *2024-02-01 and newer* |
```
mathiascode: I think a separate date column is a bit confusing. The date is the build's version number in SoulseekQt's case.

doc/SLSKPROTOCOL.md
```diff
@@ -222,14 +222,91 @@ missing if a file does not provide them.
...
 
 Server messages are used by clients to interface with the server over a
 connection (TCP). In Nicotine+, these messages are defined in slskmessages.py.
 
-If you want a Soulseek server, check out [Soulfind](https://github.com/soulfind-dev/soulfind).
-Soulfind is obviously not exactly the same as the official proprietary Soulseek
-server, but it handles the protocol well enough (and can be modified).
+If you want a Soulseek server, [Soulfind](https://github.com/soulfind-dev/soulfind)
+is open source (unrelated to the official proprietary Soulseek server). It handles
+the protocol well enough for development and testing of client implementations.
```
mathiascode:
```suggestion
If you want a Soulseek server, [Soulfind](https://github.com/soulfind-dev/soulfind)
is an open source implementation (unrelated to the official proprietary Soulseek
server). It handles the protocol well enough for development and testing of client
implementations.
```

doc/SLSKPROTOCOL.md
slook: It wasn't called "NS". It was a client called "test"...
https://web.archive.org/web/20110803190214/http://www.slsknet.org/changelog.html

mathiascode:
Thanks! Merging, since there have been no other comments.

mathiascode (approved):

PR #3690, opened 2026-03-27
uploads.py: differentiate 'started' vs 'resumed' uploads

slook:
@mathiascode This is just an idea I haven't tested properly. Basically I think it will help if we could somehow identify when an upload is resumed vs started from the beginning.

I think the Transfer() object already contains most data we would need to identify if a non-compliant peer never got the beginning of a file (that's assuming auto-clear uploads list isn't enabled) and just went in for a chunk.

So it's just a matter of presenting a warning in the interface or at least in the log so we start investigating what action should be taken to gracefully handle this kind of scenario (like not sending the speed nor incrementing the stats unless it's known that all of the entire file bytes had actually been transferred).

+ Added: Identify resumed uploads in the transfer logs

Draft PR, not for merge.

### DIFF
--- a/pynicotine/uploads.py
+++ b/pynicotine/uploads.py
@@ -1165,9 +1165,10 @@ class Uploads(Transfers):
         sock = upload.sock = msg.sock
         need_update = True
         upload_started = False
+        is_resume = bool(upload.current_byte_offset)
 
-        log.add_transfer("Initializing upload with token %s for file %s to user %s",
-                         (token, virtual_path, username))
+        log.add_transfer("%s upload with token %s for file %s to user %s",
+                         ("Resuming" if is_resume else "Initializing", token, virtual_path, username))
 
         real_path = core.shares.virtual2real(
             virtual_path,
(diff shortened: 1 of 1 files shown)


mathiascode:
I don't think we can identify this reliably (e.g. if an upload resumes after a restart), but it could still make sense to be more selective about when to send the upload speed to the server.

slook:
The client restarting is the only case when we couldn't track this reliably. During the session, we could verify most of this behaviour by knowing if a transfer suddenly appeared half way in the middle during the session or if it was already in the history like that at startup.

Such as we could store `Transfer().initial_byte_offset` == `0` or `None` respectively (restart it'd be unset), so if set the value should normally be equal to zero, more than this would indicate something fishy occurred.

slook:
If the log line wording is changed as suggested in this PR, then you could manually examine the transfer log file to spot if the first ever instance of a particular transfer says "Resuming" then this is easier to find than hunting down a non-zero offset value in the protocol message log file.

PR #3698, opened 2026-04-03
CI: add experimental AppImage package

### LINKED ISSUE #2991: Appimage download
Given the closed source client is downloadable as an appimage, this FOSS client should compete with them rather than restrict users to various package managers only

It's not even easy to build an appimage myself from the package list because most listed repos don't provide binary downloads of their packages, and the ones that do provide binaries that are too new for my glibc

The closed source appimage was bult in 2018, so it'll run for me, but I would like to use a client I can trust

Appimages are very easy to build, just compile to a directory (`nicotine.AppDir/`), and write an AppRun:

```sh
#!/usr/bin/sh

HERE=${0%/*} # bash sucks, if this doesn't work, readlink isn't a portable solution either
export PATH="${HERE}:${HERE}/usr/bin/:${HERE}/usr/sbin/:${HERE}/usr/games/:${HERE}/bin/:${HERE}/sbin/${PATH:+:$PATH}"
export LD_LIBRARY_PATH="${HERE}/usr/lib/:${HERE}/usr/lib/i386-linux-gnu/:${HERE}/usr/lib/x86_64-linux-gnu/:${HERE}/usr/lib32/:${HERE}/usr/lib64/:${HERE}/lib/:${HERE}/lib/i386-linux-gnu/:${HERE}/lib/x86_64-linux-gnu/:${HERE}/lib32/:${HERE}/lib64/${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
DEFAULT_XDG_DATA_DIRS='./share/:/usr/share/gnome:/usr/local/share/:/usr/share/'
export XDG_DATA_DIRS="${HERE}/usr/share/:${XDG_DATA_DIRS:-$DEFAULT_XDG_DATA_DIRS}"

# Other env locations may need to be defined for GTK and the like

nicotine "$@" # nothing else works well
```

You may also need to provide a .DirIcon before the actual appimage may build

**Additional context:**

1: I don't use package managers because they've repeatedly proven to be broken and unreliable, every last one of them

2: I won't use a Linux kernel >= 6.0 because I don't agree with the Rust movements and forced additions (memory safety takes skill, and Rust is not a solution for that)

Currently the kernel I use is 5.15 from Artix with glibc 2.35
but I'm looking to switch to SliTaz 5 which is 3.16 allegedly

(older means less bloat (no additional nonsense), which may mean less exploits if the kernel is selected and other exploits patched with care)

mathiascode:
Recent developments have paved the way for this, and we can mostly reuse our Windows and macOS packaging here.

Closes #2991


.github/workflows/packaging.yml
```diff
@@ -114,6 +114,49 @@ jobs:
...
+    defaults:
+      run:
+        shell: bash -el {0}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+
+      - name: Set up Miniforge
+        run: |
+          wget "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-$(uname)-$(uname -m).sh"
```
slook, line 131: `latest` -> `tag/26.1.1-3`

```sh
wget "https://github.com/conda-forge/miniforge/releases/tag/26.1.1-3/download/Miniforge3-$(uname)-$(uname -m).sh"
```

Note these aren't shell scripts they've got big blobs in them.

slook (changes requested):

.github/workflows/packaging.yml
```diff
@@ -114,6 +114,50 @@ jobs:
           name: snap-${{ matrix.arch }}-package
           path: nicotine-plus_*.snap
 
+  appimage:
+    timeout-minutes: 15
+    strategy:
+      matrix:
+        include:
+          - arch: x86_64
+            platform: ubuntu-latest
```
slook, line 123: Upstream is only tested to work on up to Ubuntu 24.04 https://github.com/conda-forge/miniforge#testing

They only bother testing with LTS releases, so the next one will be [Ubuntu 26.04 LTS (Resolute Raccoon)](https://www.releases.ubuntu.com/26.04/)

mathiascode: GitHub Actions only has LTS Ubuntu releases, and it takes a while for them to point `ubuntu-latest` at the latest release anyway.

.github/workflows/packaging.yml
mathiascode: Not sure if there's a way for Dependabot to bump this version the same way it does in other cases. I'd lean towards just using the latest tag, since I don't think we gain much by pinning the version if the packages are rolling anyway.

mathiascode:
@slook Can you try the AppImage if you have time? Just so I know it at least works on someone else's system before I merge.

slook:
yes it does run

.github/workflows/packaging.yml
slook: Given that it works satisfactorily now with the version it's been built with, then I don't see what benefits would be gained (without reviewing the change log first) by using something different.

PR #3706, opened 2026-04-09
GUI: inhibit logout until state is saved

### LINKED ISSUE #3383: Nicotine doesn't save the config when closing from a general computer shutdown
Nicotine+ version:  Nicotine+ 3.3.9.dev1
Operating System: EndeavourOS 
KDE Plasma Version: 6.3.5
KDE Frameworks Version: 6.14.0
Qt Version: 6.9.0
Kernel Version: 6.14.6-arch1-1 (64-bit)
Graphics Platform: X11
Processors: 12 × AMD Ryzen 5 5600X 6-Core Processor
Memory: 31.3 Gio of RAM
Graphics Processor: AMD Radeon RX 6600 XT
Manufacturer: TOP ACHAT
Product Name: CUSTOM
System Version: V1

### Describe the bug
When closing Nicotine manually, no problems occur and the config file gets saved upon exit.
However, when I shut down my computer, the app cuts without saving the files preemptively.

The bug has been verified with logs enabled : no logs show any attempt at writing the config file when I shutdown my computer, only when I close nicotine manually.

### Expected behavior
The config file should be saved when the computer shuts down. Maybe X11 is behaving weirdly and doesn't send a `SIGTERM` to close apps.

Shut down the computer with nicotine still open.

### Additional context

```
I discussed on IRC a little, the logs I get when closing manually are as follows:
[2025-05-14 18:32:57] Quitting Nicotine+ 3.3.9.dev1, application closing…
[2025-05-14 18:32:57] [Misc] NAT-PMP: Binding socket to local IP address 192.168.1.60
[2025-05-14 18:32:57] [Misc] NAT-PMP: Portmap request attempt 1 of 2 to gateway 192.168.1.254, port 5351: b'\x00\x02\x00\x00\x08\xba\x00\x00\x00\x00\x00\x00'
[2025-05-14 18:32:57] Disconnected from server server.slsknet.org:2242
[2025-05-14 18:32:57] [Misc] Backed up and saved file /home/yves/.config/nicotine/config
[2025-05-14 18:32:57] [Misc] Backed up and saved file /home/yves/.local/share/nicotine/uploads.json
[2025-05-14 18:32:57] [Misc] Backed up and saved file /home/yves/.local/share/nicotine/downloads.json
[2025-05-14 18:32:57] [Misc] Backed up and saved file /home/yves/.config/nicotine/config
[2025-05-14 18:32:57] Quit Nicotine+ 3.3.9.dev1!
```
No logs at all show up when I shutdown my computer.

mathiascode:
Closes #3383

### DIFF
--- a/pynicotine/gtkgui/application.py
+++ b/pynicotine/gtkgui/application.py
@@ -56,7 +56,7 @@ class Application:
 
     def __init__(self, start_hidden, ci_mode, isolated_mode, multi_instance):
 
-        self._instance = Gtk.Application(application_id=pynicotine.__application_id__)
+        self._instance = Gtk.Application(application_id=pynicotine.__application_id__, register_session=True)
         GLib.set_application_name(pynicotine.__application_name__)
         GLib.set_prgname(pynicotine.__application_id__)
 
@@ -80,6 +80,7 @@ class Application:
         self.tray_icon = None
         self.spell_checker = None
 
+        self.inhibit_logout_cookie = None
         self.previous_download_folder = None
         self.previous_file_download_folder = None
 
@@ -91,6 +92,7 @@ class Application:
         Gtk.Widget.set_default_direction(Gtk.TextDirection.LTR)
 
         self._instance.connect("activate", self.on_activate)
+        self._instance.connect("query-end", self.on_query_end)
         self._instance.connect("shutdown", self.on_shutdown)
 
         for event_name, callback in (
@@ -129,6 +131,23 @@ class Application:
 
     def remove_window(self, window):
         self._instance.remove_window(window)
+
+    def inhibit_logout(self, reason):
+
+        if self.inhibit_logout_cookie:
+            return
+
+        self.inhibit_logout_cookie = self._instance.inhibit(
+            self.window.widget, Gtk.ApplicationInhibitFlags.LOGOUT, reason
+        )
+
+    def uninhibit_logout(self):
+
+        if not self.inhibit_logout_cookie:
+            return
+
+        self._instance.uninhibit(self.inhibit_logout_cookie)
+        self.inhibit_logout_cookie = None
 
     def get_accels_for_action(self, action_name):
         return self._instance.get_accels_for_action(action_name)
@@ -1015,6 +1034,10 @@ class Application:
 
         core.confirm_quit()
 
+    def on_query_end(self, *_args):
+        self.inhibit_logout(_("Saving configuration…"))
+        core.quit()
+
     def on_shutdown(self, *_args):
 
         if self.about is not None:
@@ -1053,4 +1076,5 @@ class Application:
         if self.tray_icon is not None:
             self.tray_icon.destroy()
 
+        self.uninhibit_logout()
         self.__dict__.clear()


pynicotine/gtkgui/application.py
```diff
@@ -981,6 +983,10 @@ def on_activate(self, *_args):
         self.chat_history = ChatHistory(self)
         self.room_list = RoomList(self)
 
+        self.inhibit_cookie = self._instance.inhibit(
+            self.window.widget, Gtk.ApplicationInhibitFlags.LOGOUT, _("Quitting…")
+        )
```
slook, line 988: It shouldn't be necessary to inhibit for the entire duration of the N+ session, the docs say that this can be set in `on_query_end()`...

_https://docs.gtk.org/gtk4/signal.Application.query-end.html_

> _Emitted when the session manager is about to end the session._
> 
> _Applications can connect to this signal and call [gtk_application_inhibit()](https://docs.gtk.org/gtk4/method.Application.inhibit.html) with [GTK_APPLICATION_INHIBIT_LOGOUT](https://docs.gtk.org/gtk4/flags.ApplicationInhibitFlags.html#logout) to delay the end of the session until state has been saved._

`"Quitting…"` -> `"Quit Nicotine+"`

Consider making `inhibit()` and `uninhibit()` methods, because these might also be useful to call for other reasons such as "Wait for uploads to finish", during disk ops while moving finished downloads, and so on.

mathiascode: > `"Quitting…"` -> `"Quit Nicotine+"`

Quitting would be the correct form, looking at what other applications do here. It's a description that appears in the GUI, telling the user why the application is waiting.

slook: > telling the user why the application is waiting

There needs to be some good reason clearly specified to tell the user about the potential consequences of force-killing the process if something goes wrong at this time. It cannot be generic. Suggest:

Quitting Nicotine+ 3.3.11rc1, saving configuration…

```
self.inhibit_logout(_("saving configuration"))

...

_("Quitting %(program)s %(version)s, %(status)s…") % {
    "program": pynicotine.__application_name__,
    "version": pynicotine.__version__,
    "status": reason
}
```

_or..._
"saving configuration"
"writing config to disk"

But adding that better explanation can be addressed at a later time if desired. The main thing now is that this mechanism is working in the correct way to close the bug being addressed.

mathiascode: Maybe we can just go with "Saving configuration…"? The reason appears next to the application name and icon anyway.

For the 3.3.11 backport, I'll use "Quitting...", since we have a translated string for that.

slook: It would be useful if we had a template string, i.e. `"Saving %(file)s…" % {"file": "configuration"}`, because in the future we might want to inhibit the logout during other, potentially much longer, filesystem operations such as "downloaded folder" (perhaps even including the path) in cases where the program is in the middle of moving a batch of finished files out of the incomplete location and into the final destination (that's equivalent to the CD/DVD burning example given in the Gtk docs), or perhaps the "shares database" after finishing a rescan, and so on.

> For the 3.3.11 backport, I'll use "Quitting..."

That will be fine for now (since the config file is so small that it's unlikely to inhibit for very long), but ultimately this negates the entire purpose of having a logout inhibit reason, which is to identify why this program is causing their system to hang, i.e. what operation is deemed so important that intentionally hanging the entire OS by blocking a privileged system action is considered better than data loss.

If there is no apparent good reason given, then the user will just force kill whatever process has caused their system to hang, without first waiting for the problem with their disk to be correctly resolved.

mathiascode: > It would be useful if we had a template string, i.e. `"Saving %(file)s…" % {"file": "configuration"}`, because in the future we might want to inhibit the logout during other, potentially much longer, filesystem operations such as "downloaded folder" (perhaps even including the path) in cases where the program is in the middle of moving a batch of finished files out of the incomplete location and into the final destination (that's equivalent to the CD/DVD burning example given in the Gtk docs), or perhaps the "shares database" after finishing a rescan, and so on.

Adding such a short template string isn't worth it, it just makes it more difficult for translators to piece things together. We can just add more translatable strings later if needed, whether they start with "Saving" or something else.

> > For the 3.3.11 backport, I'll use "Quitting..."
> 
> That will be fine for now (since the config file is so small that it's unlikely to inhibit for very long), but ultimately this negates the entire purpose of having a logout inhibit reason, which is to identify why this program is causing their system to hang, i.e. what operation is deemed so important that intentionally hanging the entire OS by blocking a privileged system action is considered better than data loss.
> 
> If there is no apparent good reason given, then the user will just force kill whatever process has caused their system to hang, without first waiting for the problem with their disk to be correctly resolved.

Yes, but we have to use something for 3.3.11.

PR #3707, opened 2026-04-11
tests.yml: add Xvfb display server for FreeBSD CI

radiosilence:
The FreeBSD test job consistently fails on `test_gui_startup` because no display server is available. The Broadway backend isn't found, and without a fallback display, GTK crashes immediately (`GDK_IS_DISPLAY (display)` assertion failure).

This mirrors the existing OpenBSD job which already installs and starts Xvfb. Adds `xorg-vfbserver` to the FreeBSD package list and starts `Xvfb :99` before running tests.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -168,7 +168,11 @@ jobs:
           version: '15.0'
           run: |
             sudo pkg install -y dejavu gettext gtk4 python3 devel/py-build devel/py-pygobject \
-              devel/py-setuptools devel/py-wheel
+              devel/py-setuptools devel/py-wheel xorg-vfbserver
+
+            export DISPLAY=:99
+            Xvfb :99 > /dev/null 2>&1 &
+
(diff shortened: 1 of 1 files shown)


slook:
Thank you for your PR pointing out a correct solution to the problem with this broken runner.

If it would be okay, I suggest the maintainer ought to commit it directly instead of merging this PR, because I notice your fork has recently had it's own tests actions runners configured, which is fine of course, but our repo workflows are not yet hardened enough for approving workflow runs against AI assisted workflow configuration changes from contributors' forks.

radiosilence:
Yeah feel free to just copy paste it dudes :) My fork is purely so I can have personal nightly builds + no-quarantine casks.

mathiascode:
We're now using Wayland for the FreeBSD CI job in https://github.com/nicotine-plus/nicotine-plus/commit/fc696db86ff43200658f43cc8fffbb61a7bd99a5.

PR #3720, opened 2026-04-23
Type hint string utils

ernieIzde8ski:
I'm currently trying to break up my patches into smaller commits. Hopefully that makes it easier both to write & to review them.

### DIFF
--- a/pynicotine/utils.py
+++ b/pynicotine/utils.py
@@ -6,8 +6,10 @@
 # SPDX-FileCopyrightText: 2001-2003 Alexander Kanavin
 # SPDX-License-Identifier: GPL-3.0-or-later
 
+from collections.abc import Iterable
 import os
 import sys
+from typing import Literal
 
 UINT32_LIMIT = 4294967295
 UINT64_LIMIT = 18446744073709551615
@@ -131,7 +133,7 @@ REPLACEMENTCHAR = "_"
 TRANSLATE_PUNCTUATION = str.maketrans(dict.fromkeys(PUNCTUATION, " "))
 
 
-def clean_file(basename):
+def clean_file(basename: str) -> str:
 
     for char in ILLEGALFILECHARS:
         if char in basename:
@@ -146,7 +148,7 @@ def clean_file(basename):
     return basename
 
 
-def clean_path(path):
+def clean_path(path: str) -> str:
 
     path = os.path.normpath(path)
 
@@ -171,7 +173,7 @@ def clean_path(path):
     return path
 
 
-def encode_path(path, prefix=True):
+def encode_path(path: str, prefix: bool = True):
     """Converts a file path to bytes for processing by the system.
 
     On Windows, also append prefix to enable extended-length path.
@@ -188,7 +190,7 @@ def encode_path(path, prefix=True):
     return path.encode("utf-8")
 
 
-def human_duration_approx(seconds):
+def human_duration_approx(seconds: float) -> str:
     """Returns a string for an approximatation of any amount of time in a
     sensible unit of measurement, for use where precision isn't required."""
 
@@ -212,7 +214,7 @@ def human_duration_approx(seconds):
     return ngettext("%(num)s day", "%(num)s days", days) % {"num": humanize(days)}
 
 
-def human_length(seconds):
+def human_length(seconds: float) -> str:
     """Returns a string for exact ISO 8601 timestamp for track playing length
     when appropriate, approximate length otherwise."""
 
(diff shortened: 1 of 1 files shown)


pynicotine/utils.py
```diff
@@ -229,7 +231,7 @@ def human_length(seconds):
     return f"{minutes}:{seconds:02d}"
 
 
-def _human_speed_or_size(number, unit=None):
+def _human_speed_or_size(number: float, unit: Literal["B"] | None = None) -> str:
```
slook: The number is specified in Bytes which is always an `int`

ernieIzde8ski: I don't think I follow; does Bytes refer to `GLib.Bytes`?

slook: No. If it is a local file then it might be the result of something like a [os.stat_result.st_size](https://docs.python.org/3/library/os.html#os.stat_result.st_size) call, or if raw data comes in from a peer's file listing then it's size is an unsigned long long integer value returned by [struct.unpack_from()](https://docs.python.org/3/library/struct.html#struct.unpack_from) ...
https://github.com/nicotine-plus/nicotine-plus/blob/563b69f69bf2ba8bb584441bd1d08fb686b6aa45/pynicotine/slskmessages.py#L35
https://github.com/nicotine-plus/nicotine-plus/blob/563b69f69bf2ba8bb584441bd1d08fb686b6aa45/pynicotine/slskmessages.py#L377-L379
So in any case `result` is then stored and processed as just a plain old integer of zero or more. We don't quite get all the way down to level of dealing with individual bits!

When a size or speed gets displayed in a Gtk widget then it's rendered into a string alongside other "human" readable text. We don't depend on `GLib` for processing since the program can run in `--headless` mode.

pynicotine/utils.py
```diff
@@ -246,19 +248,21 @@ def _human_speed_or_size(number, unit=None):
...
 
-def humanize(number):
+def humanize(number: float) -> str:
     return f"{number:n}"
 
 
-def factorize(filesize, base=1024):
+def factorize(
+    filesize: str, base: int = 1024
+) -> tuple[None, None] | tuple[int | None, int]:
```
mathiascode:
```suggestion
) -> tuple[int | None, int | None]:
```

pynicotine/utils.py
```diff
@@ -6,8 +6,10 @@
 # SPDX-FileCopyrightText: 2001-2003 Alexander Kanavin
 # SPDX-License-Identifier: GPL-3.0-or-later
 
+from collections.abc import Iterable
```
mathiascode: `from x import y` imports should be moved after `import x` imports, separated by a space. You can see other files for examples.

pynicotine/utils.py
Sourcerer2: I set this one more strictly to `tuple[int, int] | tuple[None, int] | tuple[None, None]` as `tuple[int, None]` is not actually possible, which is implicitly acknowledged when this function is called and only `digit` is checked: https://github.com/nicotine-plus/nicotine-plus/blob/3f9ec565c863490fadd28c50789ad0982a9d9158/pynicotine/gtkgui/search.py#L1198-L1205

However, even after this, my type checker thinks `factor` is `int | None`, as it can't deduce from the previous `digit is None` check that that's actually impossible, so I've changed it to this:
```python
result = factorize(digit)
if result[0] is None:
    # Invalid Size unit
    continue
digit, factor = result
```
and now it knows that factor is `int` and nothing else.

slook: We prefer to expand by unpacking each returned parameter rather than relying on indexes.

This difficulty we are having is a symptom of the fact that this whole mechanism is just a giant bodge up. See:
https://github.com/nicotine-plus/nicotine-plus/discussions/2145#discussioncomment-7104598

I am the one who is to blame for making it even worse than it was before, because it was me who added the Duration filter entry into the pre-existing filesize/bitrate logic and made factorize() into a function using code that used to be inline. Whereas really, each different type of entry deserves totally separate treatments, at Mat rightfully pointed out in that discussion, it seems needlessly complex to have 3 levels of trial and error just to figure out what filtering operation the user wishes to do, yet there are already 3 perfectly good separate text entry fields to start with in the beginning. It's a very bad design.

Furthermore, validating the user input in the GUI here is what currently prevents us from moving the guts of the filtering logic into the core where it belongs (because other user interfaces (like the headless mode) ideally shouldn't have to re-implement the entire search filtering system from scratch for each and every user interface that may be added in the future to operate the program).

Maybe the ValueError should be caught by the caller instead of handling it inside the function. Then the function would either return a pair of valid values or raise an exception if the user input that is entered into the TextEntry was syntactically incorrect.

PR #3724, opened 2026-04-28
slskmessages.py: refactor message parsing

mathiascode:
Passing message contents and offsets between methods is error-prone, since it's easy to forget to update them. Store them as instance attributes instead.

### DIFF
--- a/pynicotine/slskmessages.py
+++ b/pynicotine/slskmessages.py
@@ -274,9 +274,17 @@ class RemoveAllowedResponse(InternalMessage):
 class SlskMessage:
     """This is a parent class for all protocol messages."""
 
-    __slots__ = ()
+    __slots__ = ("_message", "_offset")
     __excluded_attrs__ = {}
     msg_type = None
+
+    def __init__(self, msg_content=None):
+        self._message = msg_content
+        self._offset = 0
+
+    def finish_parsing(self):
+        self._message = None
+        self._offset = 0
 
     @staticmethod
     def pack_bytes(content):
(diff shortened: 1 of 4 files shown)


mathiascode:
@ernieIzde8ski This is what it looks like right now. Let me know how this works for you.

ernieIzde8ski:
Doesn't look like I'll have any issues here, though I'm curious: why `ServerMessage.__init__` and not `super().__init__`?

pynicotine/slskproto.py
```diff
@@ -676,7 +676,8 @@ def _unpack_network_message(msg_class, msg_content, msg_size, conn_type, sock=No
             if allowed_responses is not None:
                 msg.allowed_responses = allowed_responses
 
-            msg.parse_network_message(msg_content)
+            msg.parse_network_message()
+            msg.finish_parsing()
```
slook: Do this in a `finally:` block so as to clear contents even in case of Exception.

pynicotine/slskmessages.py
```diff
@@ -3144,12 +3199,56 @@ class PeerMessage(SlskMessage):
...
 
-    def __init__(self):
+    def __init__(self, msg_content=None):
+        SlskMessage.__init__(self, msg_content)
         self.username = None
         self.sock = None
         self.addr = None
         self.allowed_responses = set()
 
+    def parse_file_size(self):
```
slook: Consider renaming this to `unpack_file_size()`

slook (changes requested):

mathiascode:
> Doesn't look like I'll have any issues here, though I'm curious: why `ServerMessage.__init__` and not `super().__init__`?

Mainly because the former is slightly faster, but it it causes problems, we can change it.

mathiascode:
Merging for further testing.

PR #3675, opened 2026-03-18
Type windows/setup.py

ernieIzde8ski:
I lied, this file is more trivial than `core.py`. I don't anticipate any problems this time, given the lack of any logical changes aside from two `typing` imports & the instantialization of a `TypeVar`.

PR #3729, opened 2026-05-01
Type annotations for pynicotine/*.py

Sourcerer2:
Hi, this is my first contribution here!

I started with the simplest type annotations first, most of which were for methods returning `None` that weren't explicitly annotated.

Let me know if it's okay to submit like this with lots of small commits.


mathiascode:
Thanks, but @ernieIzde8ski is doing similar work, some of which is yet to be submitted. Please don't put too much effort into this just yet.

@ernieIzde8ski What have you done so far? Is there anything left aside from the GUI that @Sourcerer2 could work on?

ernieIzde8ski:
Aside from #3720, I do have [an experimental branch](https://github.com/ernieIzde8ski/nicotine-plus/tree/typings), but given that it exists in a somewhat broken state, I've mostly only kept it there for reference. In particular `__builtins__.pyi` is useful with Pyright:

```py
import gettext

_ = gettext.gettext
ngettext = gettext.ngettext
```

mathiascode:
> Aside from #3720, I do have [an experimental branch](https://github.com/ernieIzde8ski/nicotine-plus/tree/typings), but given that it exists in a somewhat broken state, I've mostly only kept it there for reference.

Are there any PRs you're planning to submit in the near future? I'm fine with two contributors doing this work, but it would be good if we could avoid duplicate work.

@Sourcerer2 You can look at the comments in previous PRs to get an idea of what to expect:
https://github.com/nicotine-plus/nicotine-plus/pull/3674
https://github.com/nicotine-plus/nicotine-plus/pull/3675
https://github.com/nicotine-plus/nicotine-plus/pull/3720

> Let me know if it's okay to submit like this with lots of small commits.

The PRs will be squash merged, so it doesn't matter much.

Sourcerer2:
Thanks. I've read through all those PRs and used them as a guide when making my changes.

ernieIzde8ski:
> Are there any PRs you're planning to submit in the near future? I'm fine with two contributors doing this work, but it would be good if we could avoid duplicate work.

@mathiascode I have no immediate plans, no.

slook:
I don't understand what benefit is gained by annotating methods that don't contain any return statements with `-> None`.

Surely doing so would only be useful in methods that actually return some value?

Sourcerer2:
> I don't understand what benefit is gained by annotating methods that don't contain any return statements with `-> None`.
> 
> Surely doing so would only be useful in methods that actually return some value?

My main reason is to make it clear that the return type annotation wasn't simply forgotten about, but that it's actually `None`. You can also tell at a glance that the return type is `None` without having to look at its implementation, whether or not it has a `return` somewhere in its body, but it also warns you if you add a non-`None` return to the function, which would pass without warnings if the return type wasn't annotated.

To quote The Zen of Python: "Explicit is better than implicit."

Note that PEP484 even recommends adding `-> None` to `__init__`.

ernieIzde8ski:
> To quote The Zen of Python: "Explicit is better than implicit."

I don't see it as an issue, but I've never really understood why this would apply to `-> None`. A language like Rust typically enforces much more explicit rules than Python, but even they flag it as a warning when you return the unit type:

```
    Checking playground v0.0.1 (/playground)
warning: unneeded unit return type
 --> src/main.rs:1:10
  |
1 | fn main() -> () { println!("Hello, world"); }
  |          ^^^^^^ help: remove the `-> ()`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.95.0/index.html#unused_unit
  = note: `#[warn(clippy::unused_unit)]` on by default
```

I suppose it would fix editor issues if you use an LSP that doesn't actually check function bodies, though this shouldn't be an issue with pyright or mypy.

ernieIzde8ski:
Hum, on further thought, it does make a little more sense for that in Rust than in Python. The default return type of a Rust function _is_ the unit type, whereas if I were to describe a default return type for Python functions, it would probably be `Any`, not `None`.

ernieIzde8ski:
@Sourcerer2 I'm gonna go ahead and make some small PRs for stuff outside `pynicotine/*.py`, you should be clear to keep going

pynicotine/utils.py
```diff
@@ -249,21 +250,19 @@ def _human_speed_or_size(number: int, unit: Literal["B"] | None = None) -> str:
...
 
 
 def humanize(number: float) -> str:
     return f"{number:n}"
 
 
-def factorize(
-    filesize: str, base: int = 1024
-) -> tuple[int | None, int | None]:
+def factorize(filesize: str, base: int = 1024) -> tuple[None, None] | tuple[int, int] | tuple[None, int]:
```
Sourcerer2: See https://github.com/nicotine-plus/nicotine-plus/pull/3720#discussion_r3172932668.

pynicotine/utils.py
```diff
@@ -382,8 +383,8 @@ def replace_text(text: str, replacements: dict[str, str]) -> str:
     return text
 
 
-def execute_command(command, replacement=None, background=True, returnoutput=False,
-                    hidden=False, placeholder="$"):
+def execute_command(command: str, replacement: str | None = None, background: bool = True,
+                    returnoutput: bool = False, hidden: bool = False, placeholder: str = "$") -> bytes | Literal[True]:
```
Sourcerer2, line 388: Is there a reason it returns `True` instead of returning `b''`? Wherever the return value is captured, it's assumed it's `bytes`, never `bool`

pynicotine/users.py
```diff
@@ -137,10 +137,10 @@ def request_ip_address(self, username, notify=False):
     def request_set_status(self, status) -> None:
         core.send_message_to_server(SetStatus(status))
 
-    def request_user_stats(self, username) -> None:
+    def request_user_stats(self, username: str) -> None:
         core.send_message_to_server(GetUserStats(username))
 
-    def watch_user(self, username, context=None, is_implicit=False) -> None:
+    def watch_user(self, username: str, context=None, is_implicit: bool = False) -> None:
```
Sourcerer2, line 154: The parameter `context` is marked as optional, but it seems to always be provided when calling this method, and not providing it is deprecated, so I think it should be made mandatory.
https://github.com/nicotine-plus/nicotine-plus/blob/3f9ec565c863490fadd28c50789ad0982a9d9158/pynicotine/users.py#L165

pynicotine/slskmessages.py
Sourcerer2: Both of these parameters are always provided, so there's no reason to keep them optional.

pynicotine/search.py
Sourcerer2, line 56: Are any of these parameters actually optional?

Sourcerer2:
Variable `search_term` appears 83 times, and `searchterm` appears 36 times, but they both seem to represent the same thing. Should I rename `searchterm` to `search_term` for consistency?

Variable `username` is always `str`, but `user` is only mostly `str`. There's also `user_data` which seems to be `UserData`.
My suggestions are:
`username` should always be `str`.
Either `user_data` should always be `UserData`, and we get rid of `user`, or we keep `user`, but rename `UserData` class to `User`, and get rid of `user_data`.

Sourcerer2:
There are several instances where class attributes are used instead of enums.https://github.com/nicotine-plus/nicotine-plus/blob/3f9ec565c863490fadd28c50789ad0982a9d9158/pynicotine/slskmessages.py#L130-L133

Accessing enum values seems to take only a few hundred nanoseconds longer than accessing class attributes.
The upside is that using enums is more readable and helps with type checkng: `ParentStatus` vs. `Literal[0, 1, 2]`.

pynicotine/utils.py
slook: It looks like the intention is to provide a truthy value in case the caller doesn't want an output but wants to know if the execution was successful.

pynicotine/users.py
slook: It will be depreciated throughout 3.4.x and obsoleted in 3.5.0.

> it seems to always be provided when calling this method

We cannot know if or how third-party custom plugins might be calling any method.

pynicotine/search.py
slook: `room` and `users` are always optional.

Even when mode is "rooms" then None is all joined rooms, and when mode is "users" then None means all buddies.

pynicotine/slskmessages.py
slook: Yes we must make sure to send valid strings over the network and never null or zero-length values. This needs a deeper consideration that warrants a separate PR.

In the future I'd like to add stricter validation to avoid faulty inputs from bad peers. For example the user name cannot be longer than 30 ASCII-only characters and so on.

slook:
> Should I rename searchterm to search_term for consistency?

Not in this PR, any functional changes would each have to be made in a smaller PR in case it causes a bug which may have to be reverted.

> `username` is always `str`, but `user` is only mostly `str`. There's also `user_data` which seems to be `UserData`.

I agree that consistency about this is lacking, but again any variable renaming or special cases where there is ambiguity requiring particular attention or debate should be done in separate PR's.

> Accessing enum values seems to take only a few hundred nanoseconds longer than accessing class attributes

Thank you for assessing the performance implications, since optimizing runtime speed is going to be one of the highest priorities for the project during the next development cycle.

pynicotine/__init__.py
```diff
@@ -5,9 +5,9 @@
 __application_id__ = "org.nicotine_plus.Nicotine"
 __version__ = "3.4.0.dev1"
 __author__ = "Nicotine+ Team"
-__copyright__ = """© 2004–2026 Nicotine+ Contributors
-© 2003–2004 Nicotine Contributors
-© 2001–2003 PySoulSeek Contributors"""
+__copyright__ = """© 2004-2026 Nicotine+ Contributors
+© 2003-2004 Nicotine Contributors
+© 2001-2003 PySoulSeek Contributors"""
```
slook: Revert

pynicotine/portmapper.py
```diff
@@ -173,14 +173,14 @@ def _request_port_mapping(self, public_port, private_port, lease_duration):
                     response = self.PortmapResponse(message=sock.recv(16))
                     return response.result
 
-                except socket.timeout:
+                except TimeoutError:
```
slook: Make this change in a separate PR.

pynicotine/portmapper.py
```diff
@@ -409,7 +409,7 @@ def get_services(private_ip):
                         UPnP.SSDP.add_service(
                             services, locations, UPnP.SSDPResponse(message.decode("utf-8", "replace")))
 
-                    except socket.timeout:
+                    except TimeoutError:
```
slook: Make these changes in a separate PR.

pynicotine/shares.py
```diff
@@ -546,11 +549,11 @@ def is_hidden(cls, folder, filename=None, entry=None):
             else:
                 entry_stat = entry.stat()
 
-            return entry_stat.st_file_attributes & stat.FILE_ATTRIBUTE_HIDDEN
+            return bool(entry_stat.st_file_attributes & stat.FILE_ATTRIBUTE_HIDDEN)
```
slook: Our convention is to just use parenthesis to visually delineate the evaluation without bool.

pynicotine/slskmessages.py
slook, line 1: Rebase from latest master into your branch, since recent commit https://github.com/nicotine-plus/nicotine-plus/commit/3f9ec565c863490fadd28c50789ad0982a9d9158 this module has significantly changed.

pynicotine/userbrowse.py
```diff
@@ -323,16 +323,16 @@ def save_shares_list_to_disk(self, username):
         except Exception as error:
             log.add(_("Cannot save shares for user %(user)s: %(error)s"), {"user": username, "error": error})
 
-    def download_file(self, username, folder_path, file_data, download_folder_path=None):
+    def download_file(self, username: str, folder_path: str, file_data, download_folder_path=None) -> None:
 
         _code, basename, file_size, _ext, file_attributes, *_unused = file_data
-        file_path = "\\".join([folder_path, basename])
+        file_path = f"{folder_path}\\{basename}"
```
slook: Revert. Make these string formatting changes in a separate PR.

pynicotine/userinfo.py
```diff
@@ -130,24 +132,22 @@ def show_user(self, username=None, refresh=False, switch_page=True):
...
 
     @staticmethod
-    def save_user_picture(file_path, picture_bytes):
+    def save_user_picture(file_path: str, picture_bytes: bytes) -> None:
 
         try:
-            with open(encode_path(file_path), "wb") as file_handle:
-                file_handle.write(picture_bytes)
-
+            Path(encode_path(file_path)).write_bytes(picture_bytes)
```
slook: Revert. Make this change in a separate PR and explain rationale.

pynicotine/utils.py
```diff
@@ -166,15 +168,15 @@ def clean_path(path: str) -> str:
         if char in path:
             path = path.replace(char, REPLACEMENTCHAR)
 
-    path = "".join([drive, path])
+    path = f"{drive}{path}"
```
slook: Make these path string formatting changes in a separate PR.

pynicotine/utils.py
```diff
@@ -232,7 +234,7 @@ def human_length(seconds: float) -> str:
     return f"{minutes}:{seconds:02d}"
 
 
-def _human_speed_or_size(number: int, unit: Literal["B"] | None = None) -> str:
+def _human_speed_or_size(number: float, unit: Literal["B"] | None = None) -> str:
```
slook: Revert. Only whole bytes are accepted as input.

pynicotine/utils.py
```diff
@@ -249,21 +251,19 @@ def _human_speed_or_size(number: int, unit: Literal["B"] | None = None) -> str:
...
 
 
 def humanize(number: float) -> str:
     return f"{number:n}"
 
 
-def factorize(
-    filesize: str, base: int = 1024
-) -> tuple[int | None, int | None]:
+def factorize(filesize: str, base: int = 1024) -> tuple[int, int] | tuple[None, int] | tuple[None, None]:
```
slook: Revert for now. As discussed in the other PR this is special case requiring wider refactoring.

pynicotine/utils.py
```diff
@@ -470,15 +470,15 @@ def execute_command(command, replacement=None, background=True, returnoutput=Fal
 
     # subcommands is now: [['C:\Program Files\WinAmp\WinAmp.exe', '--xforce', '--title=My Title', '$'], ['flite', '-t']]
     if replacement:
-        for i, _ in enumerate(subcommands):
+        for i in range(len(subcommands)):
```
slook: Revert and make this change in a separate PR.

slook (changes requested):

pynicotine/shares.py
Sourcerer2: Here I'm converting it to `bool()` because the method is supposed to return `bool`, not `int`, which is what it did before

slook: Oh right yes you're right, good find. I don't think that was intentional but you best leave it the way it was and better to fix the bug in a separate PR.

Most of the code in this function is unreachable anyway because `entry` from the caller is never None since we always have the stat entry from the `scandir()`. So the whole implementation of this old function is rubbish and so it needs to be reviewed at a later date.

You can add a `# TODO:` code comment or make a note for things like this in the https://github.com/nicotine-plus/nicotine-plus/discussions/3686 discussion thread that is where we keep our TODO list of minor nitpicks for each development cycle.

mathiascode:
> Variable `search_term` appears 83 times, and `searchterm` appears 36 times, but they both seem to represent the same thing. Should I rename `searchterm` to `search_term` for consistency?
> 
> Variable `username` is always `str`, but `user` is only mostly `str`. There's also `user_data` which seems to be `UserData`. My suggestions are: `username` should always be `str`. Either `user_data` should always be `UserData`, and we get rid of `user`, or we keep `user`, but rename `UserData` class to `User`, and get rid of `user_data`.

Variable naming is something we've been slowly fixing over the years, but it's not done yet. Especially most of the instance variable names in slskmessages.py are very old.

We want to use `search_term` and `username`, but please make those changes in a separate PR. Not quite sure about `user_data` yet.

mathiascode:
> There are several instances where class attributes are used instead of enums.
> 
> https://github.com/nicotine-plus/nicotine-plus/blob/3f9ec565c863490fadd28c50789ad0982a9d9158/pynicotine/slskmessages.py#L130-L133
> 
> Accessing enum values seems to take only a few hundred nanoseconds longer than accessing class attributes. The upside is that using enums is more readable and helps with type checkng: `ParentStatus` vs. `Literal[0, 1, 2]`.

It's something we could consider, but could you do it in a separate PR? It would make it easier to compare performance and test if it's an isolated changed.

Sourcerer2:
> It's something we could consider, but could you do it in a separate PR? It would make it easier to compare performance and test if it's an isolated changed.

Yes, I haven't made those kinds of changes here.
In some places where I've seen the usage of a class as `enum`, I've set the type annotation to be some kind of `Literal` instead of a generic `int` or `str`.

pynicotine/interests.py
```diff
@@ -154,7 +160,7 @@ def _similar_users(self, msg):
             # Request user status, speed and number of shared files
             core.users.watch_user(user.username, context="interests")
 
-    def _item_similar_users(self, msg):
+    def _item_similar_users(self, msg: ItemSimilarUsers) -> None:
         """Server code 112."""
 
         self._similar_users(msg)
```
Sourcerer2, line 166: Why is this sending `ItemSimilarUsers` to a method that accepts `SimilarUsers`?

pynicotine/slskproto.py
```diff
@@ -1940,7 +1945,7 @@ def _calc_upload_limit_by_transfer(self):
     def _calc_upload_limit_none(self):
         return self._calc_upload_limit(limit_disabled=True)
```
Sourcerer2, line 1980: This `return` is unnecessary as the underlying function never returns anything.

pynicotine/userbrowse.py
```diff
@@ -88,26 +93,28 @@ def _show_user(self, username, path=None, new_request=False, switch_page=True):
...
             self.remove_user(username)
 
-    def _parse_local_shares(self, username, msg):
+    def _parse_local_shares(self, username: str, msg) -> None:
         """Parse a local shares list and show it in the UI."""
 
         msg.parse_network_message(memoryview(msg.make_network_message()))
         msg.username = username
 
         events.emit_main_thread("shared-file-list-response", msg)
```
Sourcerer2, line 117: What type is `msg` here? The way it's called suggests `SharedFileListResponse`, but that doesn't seem right based on the next line.

pynicotine/interests.py
slook: They both have the same (or similar) structure...

- `SimilarUsers` is a list of users ranked by the number of matching liked things to our own list of liked interests.
- `ItemSimilarUsers` is a list of users who like (or hate) a specific recommendation (or unrecommendation) item that is selected from a server-provided global list (if we don't like anything) or personal list (if we like things).

pynicotine/slskproto.py
slook: Make a note to fix in a separate PR.

pynicotine/userbrowse.py
slook: Isn't it a `FolderContentsResponse` or `FileListMessage`? It's a zlib compressed list which gets made when we finish rescanning our shares (we keep this to avoid having to re-compress it each time a peer requests it), which is the same kind of data we receive from a peer when they respond to a browse request from us.

Sourcerer2: On line 137, it's called with 
```python
msg = core.shares.compressed_shares.get(current_permission_level)
Thread(target=self._parse_local_shares, args=(username, msg), name="LocalShareParser").start()
```
and those are defined in shares.py:
```python
self.compressed_shares = {
    PermissionLevel.PUBLIC: SharedFileListResponse(permission_level=PermissionLevel.PUBLIC),
    PermissionLevel.BUDDY: SharedFileListResponse(permission_level=PermissionLevel.BUDDY),
    PermissionLevel.TRUSTED: SharedFileListResponse(permission_level=PermissionLevel.TRUSTED),
    PermissionLevel.BANNED: SharedFileListResponse(permission_level=PermissionLevel.BANNED)
}
```

pynicotine/interests.py
Sourcerer2: That's true, but that's only relying on the fact that they both have the same `users` member. Maybe the hint of `_similar_users()` should be expanded to `SimilarUsers | ItemSimilarUsers`.

Sourcerer2:
I think this is ready.

I have intentionally left some identifiers unannotated because it makes them easier to find when fixing the naming inconsistencies, and some were initialized to `None`, but they don't seem to be optional or `None`, so I left those as they are instead of annotating them with `type | None`, as that would introduces even more warnings down the line because they're never checked for `None` before being used.

pynicotine/userbrowse.py
```diff
@@ -391,10 +398,10 @@ def upload_folder(self, username, requested_folder_path, local_browsed_user, rec
             requested_folder_path, browsed_user=local_browsed_user, recurse=recurse
         ):
             for _code, basename, *_unused in files:
-                file_path = "\\".join([folder_path, basename])
+                file_path = f"{folder_path}\\{basename}"
```
slook: Revert

pynicotine/userbrowse.py
```diff
@@ -369,20 +376,20 @@ def download_folder(self, username, requested_folder_path, download_folder_path=
...
                     core.downloads.enqueue_download(
                         username, file_path, folder_path=destination_folder_path, size=file_size,
                         file_attributes=file_attributes)
 
-    def upload_file(self, username, folder_path, file_data):
+    def upload_file(self, username: str, folder_path: str, file_data) -> None:
 
         _code, basename, *_unused = file_data
-        file_path = "\\".join([folder_path, basename])
+        file_path = f"{folder_path}\\{basename}"
```
slook: Revert

pynicotine/userbrowse.py
```diff
@@ -369,20 +376,20 @@ def download_folder(self, username, requested_folder_path, download_folder_path=
 
             if files:
                 for _code, basename, file_size, _ext, file_attributes, *_unused in files:
-                    file_path = "\\".join([folder_path, basename])
+                    file_path = f"{folder_path}\\{basename}"
```
slook: Revert

slook (changes requested):

pynicotine/userbrowse.py
ernieIzde8ski: Shouldn't this be trivial to check? `print(type(msg))`

Sourcerer2: This code has since changed, probably due to treating `msg` as the wrong type, causing a crash. https://github.com/nicotine-plus/nicotine-plus/commit/9c6be82825f981f733c7855da5192a996ca9af2e

mathiascode:
Thanks! I'm focusing on getting the Nicotine+ 3.3.11 release ready right now, but I'll take a closer look at this PR after that.

mathiascode:
@Sourcerer2 @ernieIzde8ski Sorry for the delay. I wanted to give an update on where we're at right now.

Nicotine+ 3.3.11 is right around the corner now (famous last words...). Once it's out, I want to focus on getting these type hints into Nicotine+ 3.4.x, still. I think type hints for gtkgui/ will have to wait for 3.5.x, since we'll drop GTK 3 support then, which makes things a little easier.

I've probably caused a whole bunch of merge conflicts in this PR. If they're too hard to untangle, I can do it when I get to reviewing and working on this.

PR #3730, opened 2026-05-01
Add `__builtins__` type stub

ernieIzde8ski:
Pyright-related plugins support extending builtins via a `__builtins__.pyi` file or a `typings/__builtins__.pyi` file[^1][^2]. I opted for `__builtins__.pyi` since I didn't introduce any other new type stubs.

[^1]: Pyright (& Pylance): https://microsoft.github.io/pyright/#/builtins
[^2]: basedpyright: https://docs.basedpyright.com/v1.30.1/usage/builtins/


slook:
Is it not possible to have this inside the `pynicotine` module rather than it being in the project root? Considering that the only python code in the root is the `nicotine` script and `setup.py`, neither of which make use of `gettext`.

Also, it would be good for uneducated folks like me to have a brief comment in the code about why this file's existence is required, and about what the `gettext` entry in it is for. Otherwise, it seems rather abstract.

ernieIzde8ski:
> Is it not possible to have this inside the `pynicotine` module rather than it being in the project root? Considering that the only python code in the root is the `nicotine` script and `setup.py`, neither of which make use of `gettext`.

IIRC, no, but I'll give it a shot. `typings/__builtins__.pyi` is definitely an option though.

> Also, it would be good for uneducated folks like me to have a brief comment in the code about why this file's existence is required, and about what the `gettext` entry in it is for. Otherwise, it seems rather abstract.

Should I do the same in #3731?

slook:
> Should I do the same in https://github.com/nicotine-plus/nicotine-plus/pull/3731?

That's slightly more self-explanatory, but yes I think it makes the repo look untidy having empty files laying about for no apparent reason.

ernieIzde8ski:
> Is it not possible to have this inside the `pynicotine` module rather than it being in the project root? Considering that the only python code in the root is the `nicotine` script and `setup.py`, neither of which make use of `gettext`.

…Huh, it's not explicitly stated, but yeah, works perfect! With `__builtins__.pyi` in `pynicotine/`, it even raises type errors if you try to use `ngettext` in `setup.py` (without just importing it of course).

pynicotine/__builtins__.pyi
```diff
@@ -0,0 +1,12 @@
+"""
+`__builtins__.pyi` informs Pyright-based type checkers of additional builtins.
+See other scripts in `pynicotine/` for uses of `_` & `ngettext`.
+"""
```
slook: I think the docstring is supposed to go below the copyright header, and we don't normally use `` codeblocks in them.

```
"""This stub informs Pyright-based type checkers of additional builtins. See the
modules in pynicotine for uses of _() and ngettext() translation functions."""
```

slook (changes requested):

pynicotine/__builtins__.pyi
ernieIzde8ski: > I think the docstring is supposed to go below the copyright header

Confirmed, there's another in `tinytag.py`.

mathiascode (approved):

PR #3731, opened 2026-05-01
Add `py.typed` marker file

ernieIzde8ski:
The `py.typed` file declares a module to be typed: https://peps.python.org/pep-0561/#packaging-type-information

In my experience, adding `py.typed` to a library means both pyright & mypy get better at providing type annotations to third parties, even when a program is mostly untyped. That is, no effect on N+ itself, but hopefully useful for plugin developers.

### DIFF
--- a/pynicotine/py.typed
+++ b/pynicotine/py.typed
@@ -1,0 +1,2 @@
+# SPDX-FileCopyrightText: 2026 Nicotine+ Contributors
+# SPDX-License-Identifier: GPL-3.0-or-later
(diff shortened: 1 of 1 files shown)


pynicotine/py.typed
```diff
@@ -1,2 +1,6 @@
 # SPDX-FileCopyrightText: 2026 Nicotine+ Contributors
 # SPDX-License-Identifier: GPL-3.0-or-later
+
+This file marks the pynicotine module as supporting typing. See:
```
slook: Is "pynicotine" technically considered as a "package" rather than a "module"?

_https://peps.python.org/pep-0561/#definition-of-terms_
> _“Package” a directory or directories that namespace Python modules._

ernieIzde8ski: Yeah, I'd think so, good catch.

PR #3741, opened 2026-05-11
PeerMessage: parse_virtual_path() to validate directory name components

slook:
This is a draft PR intended to aid development and testing to try out if an approach to impose limits upon what names and how many files can be shared, searched and browsed at the protocol level, as to if it would be feasible and not be too restrictive.

- Removed: Don't replace forward slashes of incoming directory entries. No peer should transmit improper path separators in their shared folder listings. As a side effect of not replacing these, files that contain forward-slashes "`/`" in their names could be better supported (i.e. so they don't fracture the tree view nodes, and so that such files could be looked up by the peer).

+ Added: Validate virtual path and file name components to aim for partial UDF naming compliance when:
    - Scanning shares - to prevent from the client from sharing faulty directory listings to other peers on the network
    - Receiving search results
    - Receiving list of a folder contents from a peer
    - Receiving list of shared files from a browsed peer

+ Changed: Use unordered sets for ILLEGAL constants instead of ordered lists, because sets are immutable and hashable.

<details><summary>Restrictions imposed BOTH ways, on what cannot be listed for sharing AND what cannot be received...</summary>
    - No folder or file name that contains an illegal CTRL character IN it (names with illegal PRINTABLE chars ARE still allowed)</br>
    - No folder or file basename that is only 1 character long AND is an illegal file character</br>
    - No folder or file basename that is, or would strip down to, a relative directory path traversal identifier (ie, `.`, `..`, `~`, `/`)</br>
    - No folder or file basename that is, or would strip down to, an absolute zero-length string (ie, ` ` -> `\\`)</br>
    - No folder or file base stem that is, or would strip down to, a Windows system reserved name (ie, `NUL`, `CON.ext`, etc)</br>
    - No more than `4194304` total number of folders in a shared directory listing (silently drop entry or log scanner errors)</br>
    - No more than `65535` file entries within any single shared folder (silently drop entry or log scanner errors)</br>
    - _(todo note: there is no limit on the total number of files in all folders, suggest max count 65536*65536=`UINT32`)_</br>
</details>

<details><summary>Restrictions imposed on what cannot be listed by other peers when receiving responses FROM them...</summary>
    - No directory entry containing any single folder or file name component that is larger than 4 KiB (`4096` Bytes)</br>
    - No directory entry with a concatenated path+filename larger than a sum total of 32 KiB (`32768` Bytes)</br>
    - No directory entry containing a folder path that is nested deeper than `1024` parent levels in a shared tree listing</br>
    - _(note: above errors will result in a `__INVALID_SHARE__<>` sentinel being prepended to invalidate only that directory entry)_</br>
    - No more than `25000` search result entries in any single search response from a peer (any excess are silently dropped)</br>
</details>

<details><summary>Additional restrictions imposed on what cannot be listed for sharing out TO other peers...</summary>
    - No directory entry containing any single folder or file name component that is larger than `254` Bytes</br>
    - No directory entry with a concatenated path+filename larger than a sum total of 4 KiB (`4096` Bytes)</br>
    - No folder that is nested more than `512` levels deep in the shared directory tree listing</br>
    - No folder or file name that ends with a space or dot (` `, `.`)</br>
    - _(note: such errors will result in a logged message raised when rescanning shares, and that entry won't be shared)_</br>
</details>

Disallowing such shared directory entries is not intended to be total tyranny, since entries caught by the parser as implemented in this PR would probably lead to an unresolvable path later on, or it would not be downloadable by some other clients anyway.

This doesn't prevent illegal characters being within a name (provided that that character is not the only character in the name), so a further implementation to normalize path component names when saving files to disk would still be required as well.

UDF 1.02 is used a reference because it uses defacto standards relevant to multimedia filesystem interoperability. See link:
https://dev.ecma-international.org/wp-content/uploads/ECMA_TR-112-7_1st_edition_december_2023.pdf

### DIFF
--- a/pynicotine/shares.py
+++ b/pynicotine/shares.py
@@ -570,10 +570,14 @@ class Scanner:
 
             file_list = []
             has_filtered_files = False
+            has_parsed_folder = False
             virtual_folder_path_lower = virtual_folder_path.lower()
             virtual_folder_words = virtual_folder_path_lower.translate(TRANSLATE_PUNCTUATION).split()
 
             try:
+                SharedFileListResponse.parse_virtual_path(virtual_folder_path, dir_n=self.current_folder_count)
+                has_parsed_folder = True
+
                 with os.scandir(encode_path(folder_path, prefix=False)) as entries:
                     for entry in entries:
                         basename = basename_escaped = entry.name.decode("utf-8", "replace")
@@ -603,6 +607,9 @@ class Scanner:
                             if self.is_hidden(folder_path, basename, entry):
                                 continue
 
+                            SharedFileListResponse.parse_virtual_path(
+                                basename_escaped, dir_n=self.current_folder_count, file_n=len(file_list) + 1
+                            )
                             virtual_file_path = f"{virtual_folder_path}\\{basename_escaped}"
 
                             if (self.file_filter_regex
@@ -636,7 +643,7 @@ class Scanner:
 
                             self.current_file_index += 1
 
-                        except OSError as error:
+                        except (OSError, RuntimeError) as error:
                             self.writer.send(
                                 ScannerLogMessage(
                                     _("Error while scanning file %(path)s: %(error)s"),
@@ -644,7 +651,7 @@ class Scanner:
                                 )
                             )
 
-            except OSError as error:
+            except (OSError, RuntimeError) as error:
                 self.writer.send(
                     ScannerLogMessage(
                         _("Error while scanning folder %(path)s: %(error)s"),
@@ -652,7 +659,7 @@ class Scanner:
                     )
                 )
 
-            if not has_filtered_files or file_list:
+            if has_parsed_folder and (not has_filtered_files or file_list):
                 self.streams[virtual_folder_path] = self.get_folder_stream(file_list)
 
             self.current_folder_count += 1
--- a/pynicotine/slskmessages.py
+++ b/pynicotine/slskmessages.py
@@ -21,6 +21,7 @@ from socket import inet_ntoa
 from struct import Struct
 
 from pynicotine.utils import UINT32_LIMIT
+from pynicotine.utils import encode_name
 from pynicotine.utils import human_length
 
 # This module contains message classes, that networking and UI thread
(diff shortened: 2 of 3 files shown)


mathiascode:
I can see a benefit of applying some of these checks when scanning our own files, to reduce the number of problematic paths on the network. I don't like enforcing limits on incoming files that silently drop them though. It will just make users confused and think Nicotine+ is broken.

My approach has always been to make as much content as possible available, and sanitize it if necessary.

mathiascode:
Especially protocol-level restrictions on lengths and counts are problematic, because once you commit to them, you're effectively stuck with them. You will always run into problems with old clients.

slook:
The name limit of 254 bytes exists on most platforms to some degree, even on POSIX, as does the path limit of 4096. It is also problematic handling tree depths of greater 512 levels deep, not least from a user point of view. Some filesystems only allow up to 127 levels of depth. These sort of length and count restrictions are very well established cross-platform.

I think allowing 4096 Bytes for filenames and 32768 Bytes for pathnames at a depth of up to 1024 subfolder levels is more than reasonable enough, since nobody wants to navigate a tree structure like that anyway. Likewise, nobody wants to wait for 65536 files in a folder to load into the view while browsing, and many filesystems won't allow that many files in one folder.

The problem without any depth limit is that a 2GB payload consisting of only single (or even zero) character path and filenames can take as much up to about 50GB of memory once it is uncompressed, because of all the file attribute tuples. That is easy to achieve with a directory structure that is, say, 16.7 million levels deep. Whereas silently dropping excess results with my proposed limits, such a directory structure would only consume less than half of a GigaByte, thereby effectively neutralizing such a payload and allowing the program to continue running.

I also think it's reasonable to require that every component in a directory entry has to have a name that is greater than 0 characters long, since no filesystem that I'm aware of supports handling files with empty names anyway.

The purpose of my draft PR is to allow us to view what get picked up by my parser. You can use the Include filter in search or browse find for the `__INVALID_SHARE__` string to view what comes up. So far I have found it mostly only picks up peers that transmit leading backslashes in all their directory entries (so that case obviously needs special handling), and a few with forward slashes surrounded by spaces (which hits the dot or space condition) and can't currently be downloaded anyway (due to the slash replacement causing a mismatch). None of the other conditions are hardly triggered.

> make as much content as possible available, and sanitize it if necessary.

My intent is not to reduce availability but rather increase compatibility between clients (like better handling of slashes) and make more files downloadable that would otherwise not be.

PR #3749, opened 2026-05-17
Implement grouping of plugin settings

mathiascode:
Allows plugins to group their settings under headings, similar to our main preferences, by specifying the group name under the "group" key in metasettings. Grouping is especially useful when the list of plugin settings is long.


pynicotine/plugins/plugin_debugger/__init__.py
```diff
@@ -26,6 +26,7 @@ def __init__(self, *args, **kwargs):
             self.settings[function_name] = (function_name not in verbose_events)
             self.metasettings[function_name] = {
                 "description": f"{function_name}()",
+                "group": "Logged Events",
```
slook: `"Logged Events" if function_name.endswith("_event") else "Logged Notifications"`

PR #3760, opened 2026-05-28
chatrooms.py: append new room to end of autojoin list

slook:
+ Fixed: If the chat tabs were never reordered during the last session then a newly added room appeared one position to the left of the last tab whereas it should be at the very end.


mathiascode:
This again... Hopefully we don't have more bugs like this.

PR #3766, opened 2026-06-02
preferences.py: inconsistent toggle to indicate failure

slook:
+ Added: Warning icon to indicate that a plugin has failed whenever it is enabled but unloaded

image

The icon also appears so that visual feedback is provided for trying to toggle the checkbox but the plugin won't initialize.

### DIFF
--- a/pynicotine/gtkgui/dialogs/preferences.py
+++ b/pynicotine/gtkgui/dialogs/preferences.py
@@ -3421,6 +3421,12 @@ class PluginsPage:
                     "toggle_callback": self.on_toggle_plugin,
                     "hide_header": True
                 },
+                "loaded": {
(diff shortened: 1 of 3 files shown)


mathiascode:
It's one of those cases where the gap looks a bit strange when there are no failed plugins. How does the icon column look when placed at the end, similar to the download dialog?

slook:
It does not look good at the far end because it is crammed all the way on the right edge. I've also tried it on the far left side which was sort of okay except the name got truncated when its a narrow width yet there is a large gap that made it look off-centre.

So I decided its more appropriate to set a partial inconsistent toggle instead of a warning icon, since a failed state makes the toggle widget inoperable anyway.

image

PR #3781, opened 2026-06-12
Implement safe path joining to prevent path traversal

mathiascode:
Adds a new safe_path_join() function that removes illegal characters and path traversal components from the final path, and ensures it remains within the base path.


mathiascode:
Doing a 3.3.11rc2 soon after this lands, and hopefully a final 3.3.11 release this month or early July.

PR #3804, opened 2026-07-07
downloads.py: retry 'Download folder error' downloads without connecting

### LINKED ISSUE #3769: Downloads in incomplete get stuck there if they temporarily can't be moved to completed downloads folder
Nicotine+ version:  3.3.10
Operating System/Distribution: Nixos

### Describe the bug
I had accidentally set permissions for the completed downloads such that nicotine+ couldn't access it. Downloads in the incomplete download folder finished, and I saw the error that it couldn't move them to the completed folder. I investigated, realized there was a permissions issue, and fixed the permissions issue with the completed downloads folder. But the finished downloads just stayed in the incomplete downloads folder (with their prepended bit `INCOMPLETEc6674bc98610fdff807dbb7ab405920a<original filename>` filename.

### Expected behavior
Once permissions are set properly for the completed downloads folder, the downloads should be moved from the incomplete downloads folder to the complete downloads folder (and placed in their proper subfolders, where appropriate). Or there should be a way to signal the app to try again to move them.

Set permissions for the completed downloads folder such that the app's user doesn't have write access. Download something. Wait until it finishes. Fix the permissions issue (give the app's user write access). See the file remain in the incomplete downloads folder.

### Additional context

mathiascode:
Since these downloads are technically finished, but not moved to the final folder destination yet, there's no need to tie them to the transfer connection flow.

Fixes #3769

### DIFF
--- a/pynicotine/downloads.py
+++ b/pynicotine/downloads.py
@@ -190,6 +190,83 @@ class Downloads(Transfers):
                 # Mark transfer as failed in order to resume it when connected
                 self._fail_transfer(transfer)
 
+    # File Actions #
+
+    def _open_incomplete_transfer(self, transfer):
+
+        username = transfer.username
+        virtual_path = transfer.virtual_path
+        incomplete_folder_path = self.get_incomplete_download_folder()
+        offset = 0
+
+        try:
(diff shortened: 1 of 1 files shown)


slook:
Causes TypeError crash `start_time` is None see issue #3877

PR #3812, opened 2026-07-12
pluginsystem.py: add method to check if plugin failed to load

mathiascode:
Use it to keep track of failed plugins consistently in the plugin list in the GUI.

### DIFF
--- a/pynicotine/gtkgui/dialogs/preferences.py
+++ b/pynicotine/gtkgui/dialogs/preferences.py
@@ -3458,8 +3458,6 @@ class PluginsPage:
 
         self.application.preferences.set_widgets_data(self.options)
 
-        plugins_active = self.enable_plugins_toggle.get_active()
-
         for plugin_name in core.pluginhandler.list_installed_plugins():
             try:
                 info = core.pluginhandler.get_plugin_info(plugin_name)
@@ -3467,27 +3465,31 @@ class PluginsPage:
                 continue
 
             plugin_human_name = info.get("Name", plugin_name)
-            enabled = (plugin_name in config.sections["plugins"]["enabled"])
-            failed = (plugins_active and enabled and plugin_name not in core.pluginhandler.enabled_plugins)
+            enabled = plugin_name in config.sections["plugins"]["enabled"]
+            failed = core.pluginhandler.is_plugin_failed(plugin_name)
 
             self.plugin_list_view.add_row([enabled, plugin_human_name, plugin_name, failed], select_row=False)
 
         self.plugin_list_view.unfreeze()
 
     def get_settings(self):
-
-        return {
-            "plugins": {
-                "enable": self.enable_plugins_toggle.get_active()
-            }
-        }
+        return {}
 
     def check_plugin_settings_button(self, plugin_name):
         self.plugin_settings_button.set_sensitive(bool(core.pluginhandler.get_plugin_metasettings(plugin_name)))
 
+    def update_failed_plugin_states(self):
+
+        for plugin_name, iterator in self.plugin_list_view.iterators.items():
+            failed = core.pluginhandler.is_plugin_failed(plugin_name)
+            self.plugin_list_view.set_row_value(iterator, "inconsistent_data", failed)
+
     def on_failed_tooltip(self, treeview, iterator):
+
+        plugin_name = treeview.get_row_value(iterator, "name_data")
         failed = treeview.get_row_value(iterator, "inconsistent_data")
-        return _("Failed") if failed else ""
+
+        return _("Error Loading Plugin %(name)s") % {"name": plugin_name} if failed else ""
 
     def on_plugin_popup_menu(self, menu, _widget):
 
@@ -3540,9 +3542,14 @@ class PluginsPage:
     def on_toggle_plugin(self, list_view, iterator):
 
         plugin_name = list_view.get_row_value(iterator, "name_data")
-        was_loaded = (plugin_name in core.pluginhandler.enabled_plugins)
-        enabled = core.pluginhandler.toggle_plugin(plugin_name)
-        failed = (not was_loaded and not enabled)
+
+        if list_view.get_row_value(iterator, "enabled"):
+            core.pluginhandler.disable_plugin(plugin_name)
+        else:
+            core.pluginhandler.enable_plugin(plugin_name)
+
+        enabled = plugin_name in config.sections["plugins"]["enabled"]
+        failed = core.pluginhandler.is_plugin_failed(plugin_name)
 
         list_view.set_row_value(iterator, "enabled", enabled)
         list_view.set_row_value(iterator, "inconsistent_data", failed)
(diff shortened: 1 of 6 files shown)


pynicotine/pluginsystem.py
```diff
@@ -394,13 +394,14 @@ class InstallException(Exception):
...
 class PluginHandler:
-    __slots__ = ("plugin_folders", "enabled_plugins", "command_source", "commands",
+    __slots__ = ("plugin_folders", "enabled_plugins", "failed_plugins", "command_source", "commands",
                  "internal_plugin_folder", "user_plugin_folder", "_load_now_playing_sender")
 
     def __init__(self, isolated_mode=False):
 
         self.plugin_folders = []
         self.enabled_plugins = {}
+        self.failed_plugins = set()
```
slook: Consider a read-only function instead of a variable...

```python
        failed = core.pluginhandler.is_plugin_failed(plugin_name)
```

```python
    def is_plugin_failed(plugin_name):

        if not config.sections["plugins"]["enable"]:
            return False

        return (plugin_name in config.sections["plugins"]["enabled"] and plugin_name not in self.enabled_plugins)

```

slook (changes requested):

pynicotine/pluginsystem.py
mathiascode: We could probably add a `is_plugin_loaded` function too, and rename `enabled_plugins` to `loaded_plugins`. The current naming is confusing.

mathiascode:
A plugin is now always considered enabled if it ends up in the plugin list in the config, and `enable/disable_plugin()`'s return value follows that rule.

The checkbox makes more sense this way IMO. It's never checked for plugins that have invalid characters in their folder name, since those aren't allowed in the config, but otherwise we use the indeterminate checkbox.

pynicotine/gtkgui/dialogs/preferences.py
```diff
@@ -3458,36 +3458,35 @@ def set_settings(self):
...
+    def update_failed_plugin_states(self):
+
+        for plugin_name, iterator in self.plugin_list_view.iterators.items():
+            failed = core.pluginhandler.is_plugin_failed(plugin_name)
+            self.plugin_list_view.set_row_value(iterator, "inconsistent_data", failed)
+
     def on_failed_tooltip(self, treeview, iterator):
         failed = treeview.get_row_value(iterator, "inconsistent_data")
-        return _("Failed") if failed else ""
+        return _("Error while Loading Plugin") if failed else ""
```
slook: Either sentence case or capitalize all words "Error Loading Plugin".

It could also be helpful to include the module name in the tooltip, such as is in the log string, to make diagnostics easier.

pynicotine/pluginsystem.py
```diff
@@ -712,19 +715,15 @@ def enable_plugin(self, plugin_name):
...
-            self.enabled_plugins[plugin_name] = plugin
+            self.loaded_plugins[plugin_name] = plugin
             plugin.loaded_notification()
 
             log.add(_("Loaded plugin %s"), plugin.human_name)
 
         except Exception:
             from traceback import format_exc
-            log.add(_("Unable to load plugin %(module)s\n%(exc_trace)s"),
+            log.add(_("Failed to load plugin %(module)s\n%(exc_trace)s"),
```
slook: Consider removing the exception trace from this translatable string, instead append it afterwards.

pynicotine/pluginsystem.py
```diff
@@ -655,15 +655,18 @@ def enable_plugin(self, plugin_name):
         # Our config file doesn't play nicely with some characters
         if "=" in plugin_name:
             log.add(
-                _("Unable to load plugin %(name)s. Plugin folder name contains invalid characters: %(characters)s"), {
+                _("Cannot enable plugin %(name)s. Plugin folder name contains invalid characters: %(characters)s"), {
```
slook: Consider making two separate strings for each sentence, and make the first part the same as the other "Failed to load plugin" string. In a such a way, more types of errors could be handled in the future.

pynicotine/pluginsystem.py
```diff
@@ -766,9 +768,8 @@ def disable_plugin(self, plugin_name, is_permanent=True):
 
         except Exception:
             from traceback import format_exc
-            log.add(_("Unable to unload plugin %(module)s\n%(exc_trace)s"),
+            log.add(_("Failed to unload plugin %(module)s\n%(exc_trace)s"),
```
slook: It is not necessary for the exception trace to be included within this translatable string, it might as well be appended afterwards.

slook (changes requested):

pynicotine/pluginsystem.py
mathiascode: I split out the reason, but I'd like to keep the strings separate, since one is a deliberate restriction that doesn't attempt to load the plugin.

pynicotine/pluginsystem.py
```diff
@@ -712,19 +715,15 @@ def enable_plugin(self, plugin_name):
...
 
         except Exception:
             from traceback import format_exc
-            log.add(_("Unable to load plugin %(module)s\n%(exc_trace)s"),
-                    {"module": plugin_name, "exc_trace": format_exc()})
-            return False
+            log.add(_("Failed to load plugin %(name)s") + "\n%(trace)s",
+                    {"name": plugin_name, "trace": format_exc()})
 
         return True
```
slook: `return False` on failure, or change final `return is_plugin_loaded()`

pynicotine/pluginsystem.py
```diff
@@ -766,9 +768,8 @@ def disable_plugin(self, plugin_name, is_permanent=True):
 
         except Exception:
             from traceback import format_exc
-            log.add(_("Unable to unload plugin %(module)s\n%(exc_trace)s"),
-                    {"module": plugin_name, "exc_trace": format_exc()})
-            return False
+            log.add(_("Failed to unload plugin %(name)s") + "\n%(trace)s",
+                    {"name": plugin_name, "trace": format_exc()})
```
slook, line 771: `return False` on failure, or change final `return not is_plugin_loaded()`

slook (changes requested):

pynicotine/pluginsystem.py
mathiascode: Do we need a return value? To me the enabled/loaded split makes a boolean feel odd, because it's not clear which one it would indicate.

pynicotine/pluginsystem.py
mathiascode: `return not is_plugin_loaded()` would always return true anyway, due to the `finally:` cleanup.

pynicotine/pluginsystem.py
mathiascode: I tried something else in the latest commit. Not sure if it's an improvement or not.

slook: > Do we need a return value?

If none of the callers use it then it's not needed.

> I tried something else in the latest commit. Not sure if it's an improvement or not.

The new class seems a tad overboard really, and it has the disadvantage of an import being required if a plugin wants to read the state of another plugin. The meaning of "PENDING" in unclear in this context for "STANDBY". I preferred the simpler boolean function.

mathiascode: Alright, reverted.

pynicotine/gtkgui/dialogs/preferences.py
```diff
@@ -3535,9 +3542,15 @@ def on_select_plugin(self, list_view, iterator):
     def on_toggle_plugin(self, list_view, iterator):
 
         plugin_name = list_view.get_row_value(iterator, "name_data")
-        was_loaded = (plugin_name in core.pluginhandler.enabled_plugins)
-        enabled = core.pluginhandler.toggle_plugin(plugin_name)
-        failed = (not was_loaded and not enabled)
+        enabled = list_view.get_row_value(iterator, "enabled")
+
+        if enabled:
```
slook: `if list_view.get_row_value(iterator, "enabled"):`

There is no need to assign the checkbox toggle state to a variable here, and it's confusing to have two different variables called the same thing within the method.

slook (changes requested):

slook:
https://github.com/nicotine-plus/nicotine-plus/blob/4c3e31cb2489db5600131f6ba55d6e0b0fb5bc0b/pynicotine/plugins/core_commands/__init__.py#L608

`enabled_plugins` -> `loaded_plugins` or `self.parent.is_plugin_loaded(basename)`

PR #3813, opened 2026-07-12
treeview.py: add column menu items to move a coumn left or right

mathiascode:
Reordering columns by dragging and dropping their headers is broken in GTK 4.22, so use this solution for now. It's useful to have an alternative way to reorder columns in any case.


mathiascode:
> Reordering columns by dragging and dropping their headers is broken in GTK 4.22

Upstream issue is https://gitlab.gnome.org/GNOME/gtk/-/work_items/8027. Regardless if it's fixed soon or not, I'm going to backport this PR to Nicotine+ 3.3.11 (with reused Left and Right translatable strings), since GTK fixes will take some time to trickle down to distros.

PR #3824, opened 2026-07-21
Improve presentation of private room system messages

mathiascode:
Translate the messages and show them in a more appropriate form when needed.


pynicotine/privatechat.py
```diff
@@ -160,24 +161,148 @@ def _private_message_queue_add(self, msg):
...
+                ),
+                (
+                    r"The room you are trying to enter \((.*?)\) is moderated\. Please contact one of these "
+                    r"moderators if you are interested in being added to the room's member list:",
+                    _("The room you are trying to enter (%s) is moderated. Please contact one of these moderators "
+                      "if you are interested in being added to the room's member list:")
+                ),
+                (
+                    r"Room \((.*?)\) is registered as public\.",
+                    _("Room (%s) is registered as public.")
```
slook: It would be better if all the translated string formatters used the long `%(room)s` form, rather than just `%s`, like you've done below for the membership messages.

slook (changes requested):

PR #3832, opened 2026-07-25
slskproto.py: reject invalid usernames in PeerInit messages

mathiascode:
Reject very large usernames and usernames containing non-printable characters.

### DIFF
--- a/pynicotine/slskproto.py
+++ b/pynicotine/slskproto.py
@@ -350,6 +350,8 @@ class NetworkThread(Thread):
     MAX_INCOMING_MESSAGE_SIZE_16K = 16384        # 16 KiB
     TCP_BUFFER_SIZE_MEDIUM = 208896              # 204 KiB, maximum limit NetBSD accepts by default
     TCP_BUFFER_SIZE_SMALL = 16384                # 16 KiB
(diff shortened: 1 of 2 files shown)


pynicotine/slskproto.py
```diff
@@ -1706,13 +1706,19 @@ def _process_peer_init_message(self, conn, msg_type, msg_size, msg_content):
             conn_type = msg.conn_type
             addr = conn.addr
 
-            log.add_conn("Received incoming direct connection of type %s from user "
-                         "%s, address %s", (conn_type, username, addr))
+            if not username or len(username) > 255 or username == "server" or not username.isprintable():
```
slook: Room for expansion in the future isn't required. It is what it is.

`255` -> `USERNAME_MAX_LENGTH`

`core.users.USERNAME_MAX_LENGTH = 30`

mathiascode: I'm not comfortable using that limit at the protocol level. The server currently checks string lengths in bytes, and if it ever allowed unicode characters in usernames, a strict limit in the client would leave even less room for those (since they're usually multi-byte).

Maybe that will never happen, but I think it's better to have some flexibility, especially when the server makes it so easy to bump limits.

PR #3837, opened 2026-07-28
SLSKPROTOCOL.md: Login does handle an empty password

slook:
The Login message (server code 1) does support an empty string in the *password* field, because the server will respond appropriately with an `EMPTYPASSWORD` rejection reason. "A non-empty string" is therefore not "required" for sending.

Also added a note to encourage reusing the same login credentials and a warning about randomly generated usernames, with a hyperlink to the [official server rules](https://www.slsknet.org/news/node/681) page.

### DIFF
--- a/doc/SLSKPROTOCOL.md
+++ b/doc/SLSKPROTOCOL.md
@@ -5,7 +5,7 @@
 
 # Soulseek Protocol Documentation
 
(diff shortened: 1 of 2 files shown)


mathiascode:
Remember to update the comment in slskmessages.py too.

slook:
Of course, I forgot. Thanks.

mathiascode:
Thanks!

PR #3845, opened 2026-08-09
Improve text output of certain core commands

mathiascode:
Add some missing translatable strings, make certain strings more friendly for future RTL languages, and add a command to list plugins to match shares.

### DIFF
--- a/pynicotine/plugins/core_commands/__init__.py
+++ b/pynicotine/plugins/core_commands/__init__.py
@@ -9,6 +9,7 @@ from pynicotine.slskmessages import UserStatus
 class _CommandGroup:
     CHAT = _("Chat")
     CHAT_ROOMS = _("Chat Rooms")
+    PLUGINS = _("Plugins")
     PRIVATE_CHAT = _("Private Chat")
     NETWORK_FILTERS = _("Network Filters")
     SEARCH_FILES = _("Search Files")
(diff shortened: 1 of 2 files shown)


pynicotine/plugins/core_commands/__init__.py
```diff
@@ -563,8 +586,14 @@ def share_command(self, args, **_unused):
...
 
-        self.output(_("Added %(group_name)s share \"%(virtual_name)s\" (rescan required)") % {
-            "group_name": permission_level,
+        permission_levels = {
+            PermissionLevel.PUBLIC: _("public"),
+            PermissionLevel.BUDDY: _("buddy"),
+            PermissionLevel.TRUSTED: _("trusted")
+        }
+
+        self.output(_("Added %(permission_level)s share \"%(virtual_name)s\" (rescan required)") % {
```
slook, line 595: Consider using single quotes for the outers, so that the inners shall not require escaping.

`'Added %(permission_level)s share "%(virtual_name)s" (rescan required)'`

pynicotine/plugins/core_commands/__init__.py
```diff
@@ -541,10 +552,22 @@ def list_shares_command(self, args, **_unused):
             if not num_shares or args and permission_level not in args.lower():
```
slook, line 552: Check if the filtering condition would fail to work because the `parameters` choice strings are untranslated in `args`.

slook (changes requested):

PR #3855, opened 2026-08-16
ChatEntry: focus chat view before making active entry insensitive

mathiascode:
Makes sure the widget focus ends up somewhere sensible, and avoids a 'GtkText - did not receive a focus-out event.' warning in GTK 4.


mathiascode:
Two ways to test this:
- When in a joined chatroom with the chat entry focused, press Ctrl-Shift-D to disconnect, and then hold Alt.
- Switch between a joined room tab and a failed room tab (e.g. the 'dev' room).

PR #3859, opened 2026-08-18
SLSKPROTOCOL.md: risk of file size drift in TransferRequest

slook:
Researching the age old "99% Finished" problem and corrupted audio tracks that have digital glitches in them from bit rot...

Introduction:
+ Changed: Careful coordination between clients is "necessary" -> "absolutely essential"
+ Changed: compatibility with "other clients" - "the server and other peers"

TransferRequest (peer code 40):
+ Added: Procedure to follow for what to do about getting a different file size than originally requested.
+ Added: Note importance of reporting an accurate file size value.

UploadFailed (peer code 46):
+ Removed: Mention of Soulseek NS specific handling for unreadable files, because it would seem the original author's intent was about observing its behaviour as a model that other clients can follow rather than being an issue to work around.
+ Removed: The notion that a peer may ignore a failure sign and decide for themselves it's "finished" when it's stuck at 100%.
+ Changed: Clarify wording in broader terms about who is responsible for indicating failures and on what basis they happen.

File Offset (file connections):
+ Added: Note that it is an error to be sending this message if the offset is equal to the expected file size. This one is aspirational since Nicotine+ wrongly marks the transfer as finished in this case rather than trying to negotiate a failure (to be in this situation means that the peer thinks it should have more data but we no longer have enough bytes to send them).
+ Added: Note that it is an error to send this message more than once over the same F connection. Wrong retry strategy.

downloads._transfer_request_downloads():
+ Changed: I'm assuming recent versions of "the SoulseekQt client" are now capable of handling files larger than 2 GB correctly, although I haven't actually verified whether or not this is no longer the case. -> "old SoulseekQt clients"
+ Removed: "correct" because there's no guarantee the real file size is still going to be the same if we use our cached value.


mathiascode:
> to be in this situation means that the peer thinks it should have more data but we no longer have enough bytes to send them.

Not necessarily, clients can also send it when it has failed to move a finished download to the final destination. Nicotine+ 3.3.11 changes this to always happen locally, though.

I don't think there's much that can be done about an error message here? We only have the offset to tell if a transfer is incomplete or not. If something is stuck in someone's download queue, then it's also just stuck. Not much we can do to nudge it, unfortunately.

slook:
It's quite a complicated conundrum isn't it.

> I don't think there's much that can be done about an error message here?

I was thinking in this uncertain situation we could send them an UploadFailed (peer code 46 doesn't contain an error reason) and this might be enough to invoke them to re-queue it, which would later cause us to re-read the file stats size from our disk, thereby alerting them whether or not our file has changed, and if it has then they should wipe it and start over from the beginning.

> If something is stuck in someone's download queue, then it's also just stuck.

This is all assuming that a differing file size is at least sometimes the plausible reason for this and in the hope that next time they shall end up getting a consistent file of the expected size which won't get stuck.

However indeed as you say there could be other local reasons for this phenomenon, and if either peer handles it wrong then it would be easy to get caught in an endless loop of retries if were not careful.

slook:
| **Uploading** | **Downloading** |
|---|---|
| **1.** The server sent us a [FileSearch](#server-code-26) or [DistribSearch](#distributed-code-3) with a query from the peer. | **1.** We send a [FileSearch](#server-code-26) to the server, which propagates the query to every peer. |
| **2.** Having found a match, we establish a `P` connection and send a [FileSearchResponse](#peer-code-9) listing our results. | **2.** Any peers that find a match for our query establish a `P` connection with us, and we get a [FileSearchResponse](#peer-code-9) with their results. |
| **3.** The peer sent us a [QueueUpload](#peer-code-43). If we’re busy, they wait. | **3.** We send a [QueueUpload](#peer-code-43) to the peer with our selected choice, and wait for our turn. |
| **4.** We send a [TransferRequest](#peer-code-40), that includes our current file size on disk. Giving the wrong size value of an existing transfer could corrupt their data if our file has changed since a previous attempt. | **4.** The peer sends a [TransferRequest](#peer-code-40) to us. We store the size for later comparison and must erase all partial file data if that size ever changes between transfer attempts. |
| **5.** The peer replies with a [TransferResponse](#peer-code-41-a) to accept our transfer request. We load the file and get it ready to go. | **5.** We reply with a [TransferResponse](#peer-code-41-a) to accept the peer's transfer request and indicate that we're ready. |
| **6.** We establish a new `F` connection with the peer by sending a [File Transfer Init](#file-transfer-init). | **6.** The peer establishes a new `F` connection with us once they have loaded the file and got it ready to go. |
| **7.** The peer sends a [FileOffset](#file-offset) indicating how many bytes they have already downloaded. This tells us to start (or resume) from the next byte after the offset they specified. For example, if they tell us `0` we'll send them the first byte. | **7.** We send a [FileOffset](#file-offset) to the peer, telling them exactly how many bytes we've already downloaded so far. If we're starting from the beginning, the offset is `0`. |
| **8.** We keep streaming sequential byte chunks until the final byte, unless the connection drops abruptly before it's finished (in which case we should send them an [UploadFailed](#peer-code-46)). Failures are not retried by us. | **8.** The peer keeps streaming sequential byte chunks until the final byte reaches us, unless their file becomes unreadable or the connection drops (in which case we may or may not get an [UploadFailed](#peer-code-46) from them). |
| **9.** After sending all data, we must wait for a while until the downloader closes the `F` connection. We only close it if the transfer gets interrupted or has gone wrong. | **9.** After receiving all data chunks and aligning the end to its expected size, we must close the `F` connection to signal success. If the uploader ends up closing it then a failure has to be dealt with. |
| **10.** Once the peer has successfully finished, we mark the upload as `Complete` and may notify the server of our average [SendUploadSpeed](#server-code-121) for this transfer. | **10.** Upon successfully finishing, we mark the download as `Complete` and permanently save the file with its original name to make it available for opening. |

slook:
```
| **Uploading** | **Downloading** |
|---|---|
| **1.** The server sent us a [FileSearch](#server-code-26) or [DistribSearch](#distributed-code-3) with a query from the peer. | **1.** We send a [FileSearch](#server-code-26) to the server, which propagates the query to every peer. |
| **2.** Having found a match, we establish a `P` connection and send a [FileSearchResponse](#peer-code-9) listing our results. | **2.** Any peers that find a match for our query establish a `P` connection with us, and we get a [FileSearchResponse](#peer-code-9) with their results. |
| **3.** The peer sent us a [QueueUpload](#peer-code-43). If we’re busy, they wait. | **3.** We send a [QueueUpload](#peer-code-43) to the peer with our selected choice, and wait for our turn. |
| **4.** We send a [TransferRequest](#peer-code-40), that includes our current file size on disk. Giving the wrong size value of an existing transfer could corrupt their data if our file has changed since a previous attempt. | **4.** The peer sends a [TransferRequest](#peer-code-40) to us. We store the size for later comparison and must erase all partial file data if that size ever changes between transfer attempts. |
| **5.** The peer replies with a [TransferResponse](#peer-code-41-a) to accept our transfer request. We load the file and get it ready to go. | **5.** We reply with a [TransferResponse](#peer-code-41-a) to accept the peer's transfer request and indicate that we're ready. |
| **6.** We establish a new `F` connection with the peer by sending a [File Transfer Init](#file-transfer-init). | **6.** The peer establishes a new `F` connection with us once they have loaded the file and got it ready to go. |
| **7.** The peer sends a [FileOffset](#file-offset) indicating how many bytes they have already downloaded. This tells us to start (or resume) from the next byte after the offset they specified. For example, if they tell us `0` we'll send them the first byte. | **7.** We send a [FileOffset](#file-offset) to the peer, telling them exactly how many bytes we've already downloaded so far. If we're starting from the beginning, the offset is `0`. |
| **8.** We keep streaming sequential byte chunks until the final byte, unless the connection drops abruptly before it's finished (in which case we should send them an [UploadFailed](#peer-code-46)). Failures are not retried by us. | **8.** The peer keeps streaming sequential byte chunks until the final byte reaches us, unless their file becomes unreadable or the connection drops (in which case we may or may not get an [UploadFailed](#peer-code-46) from them). |
| **9.** After sending all data, we must wait for a while until the downloader closes the `F` connection. We only close it if the transfer gets interrupted or has gone wrong. | **9.** After receiving all data chunks and aligning the end to its expected size, we must close the `F` connection to signal success. If the uploader ends up closing it then a failure has to be dealt with. |
| **10.** Once the peer has successfully finished, we mark the upload as `Complete` and may notify the server of our average [SendUploadSpeed](#server-code-121) for this transfer. | **10.** Upon successfully finishing, we mark the download as `Complete` and permanently save the file with its original name to make it available for opening. |
```

PR #3861, opened 2026-08-19
slskproto.py: truncate rejected username for representing in conn log

slook:
+ Added: Only show a truncated part of a malformed `username` in the conn log when rejecting a `PeerInit`
+ Added: Only show a truncated part of a unknown `conn_type` in the conn log when rejecting a `PeerInit`

The problem with using `repr()` here is that is makes another copy of the string before the string formatter makes yet another copy, so logging the rejection causes the total memory usage of faulty username to be tripled.

Related to recent PR #3832


mathiascode:
I forgot %r exists, so I switched to it instead in https://github.com/nicotine-plus/nicotine-plus/commit/051d01fef0f85a89309767d3632835231ff01bcc.

PR #3862, opened 2026-08-19
textview.py: only compile URL regex once for all widget instances

slook:
+ Changed: Don't compile the URL regex again for each new instance of a chat tab


mathiascode:
It should already be compiled only once, since it's a class variable.

PR #3864, opened 2026-08-21
privatechat.py: adopt new CTCP config option

slook:
+ Fixed: Disabling CTCP responses had no effect since PR #3816


mathiascode:
Thanks, no idea how I forget this...

PR #3867, opened 2026-08-22
fix: detect portable mode in frozen state on Windows

NoPlagiarism:
fix small issue when frozen executable can't detect portable folder (mode)

### DIFF
--- a/pynicotine/config.py
+++ b/pynicotine/config.py
@@ -57,8 +57,13 @@ class Config:
         - the data folder
         """
 
-        script_folder_path = os.path.dirname(sys.argv[0])
-        portable_folder_path = os.path.join(script_folder_path, "portable")
+        is_frozen = getattr(sys, "frozen", False)
+
+        if is_frozen and sys.platform == "win32":
+            root_folder_path = os.path.dirname(sys.executable)
+        else:
+            root_folder_path = os.path.dirname(sys.argv[0])
+        portable_folder_path = os.path.join(root_folder_path, "portable")
 
         if os.path.isdir(portable_folder_path):
             data_folder_path = os.path.join(portable_folder_path, "data")
--- a/pynicotine/config.py
+++ b/pynicotine/config.py
@@ -58,11 +58,7 @@ class Config:
         """
 
         is_frozen = getattr(sys, "frozen", False)
-
-        if is_frozen and sys.platform == "win32":
-            root_folder_path = os.path.dirname(sys.executable)
-        else:
-            root_folder_path = os.path.dirname(sys.argv[0])
+        root_folder_path = os.path.dirname(sys.executable if is_frozen else sys.argv[0])
         portable_folder_path = os.path.join(root_folder_path, "portable")
 
         if os.path.isdir(portable_folder_path):


mathiascode:
Thank you for your PR. How are you starting Nicotine+ (or what are you doing to run into the issue you're having)? If there's a bug here, I'd guess it needs to be a more broad fix rather than special-casing it for frozen Windows builds.

NoPlagiarism:
> or what are you doing to run into the issue you're having

Let me reiterate this issue

When downloading latest portable release version of Nicotine+ for Windows and even after creating folder "portable" for data. It still chooses AppData as config folder. It was due to `get_user_folders()` lacking handling being launched from .exe.

> a more broad fix rather than special-casing it for frozen Windows builds.

I suspect that running from source code escaped such problem, and that's why it wasn't detected. But what about other platforms? Does Nicotine+ provide portable executables for other systems? I only see AppImage and I can try to test that. I understand that most installation methods do not provide portable way, but I still can't test how they will all run after removing check for specifically Windows builds

slook:
What happens if you just use the directory path of `sys.executable` for all cases?

NoPlagiarism:
I couldn't run AppImage in portable mode. My theory is `sys.executable` points to something else than AppImage (maybe to executable inside AppImage). So I, currently, do not see a way to make portable mode for AppImage

Other distributions won't benefit from this anyway, since all default ways to install do not provide a portable way to do it. Currently "portable" ways to launch program is source code, Windows zip and AppImage. Source code should recognize portable folder, Windows zip is fixed with this PR

mathiascode:
Looks like AppImage has its own optional portable mode built-in, so we don't have to worry about it: https://docs.appimage.org/user-guide/portable-mode.html

pynicotine/config.py
```diff
@@ -57,8 +57,13 @@ def get_user_folders():
...
         """
 
-        script_folder_path = os.path.dirname(sys.argv[0])
-        portable_folder_path = os.path.join(script_folder_path, "portable")
+        is_frozen = getattr(sys, "frozen", False)
+
+        if is_frozen and sys.platform == "win32":
+            root_folder_path = os.path.dirname(sys.executable)
+        else:
+            root_folder_path = os.path.dirname(sys.argv[0])
```
mathiascode:
```suggestion
        is_frozen = getattr(sys, "frozen", False)
        root_folder_path = os.path.dirname(sys.executable if is_frozen else sys.argv[0])
```

Let's just use sys.executable for all frozen builds for now, to simplify things. If people want a true portable mode on macOS in the future, we can figure something out then.

NoPlagiarism:
> AppImage has its own optional portable mode built-in

Confirmed. `-.AppImage.config` is the one working, so it means every way of using program in portable should work

slook:
The problem with [sys.executable](https://docs.python.org/3/library/sys.html#sys.executable) is that it may be `None` unless on systems "where it makes sense".

It therefore might have been better to consider using something like [sys.exec_prefix](https://docs.python.org/3/library/sys.html#sys.exec_prefix) or [sys.prefix](https://docs.python.org/3/library/sys.html#sys.prefix) instead, or perhaps [sys.base_exec_prefix](https://docs.python.org/3/library/sys.html#sys.base_exec_prefix) or [sys.base_prefix](https://docs.python.org/3/library/sys.html#sys.base_prefix) whatever the difference between them is.

https://docs.python.org/3/library/sysconfig.html#installation-paths

"If this sounds a trifle unusual, it is" https://docs.python.org/3/library/sysconfig.html#prefix-scheme

mathiascode:
We know sys.executable is always available in our frozen builds though. There's a bunch of other initialization code using it that would fail otherwise.

Issue #3877, opened 2026-09-01
Keeps on crashing after sitting idle (unknown amount of time)


grigzy28:
Nicotine+ version:  3.4.0-dev1 - snap 2064
Operating System/Distribution:  Kubuntu 24.04

Nicotine+ Version: 3.4.0.dev1
GTK Version: 4.18.6
Python Version: 3.12.3 (linux)

Type: <class 'TypeError'>
Value: unsupported operand type(s) for -: 'float' and 'NoneType'
Traceback:   File "/snap/nicotine-plus/2064/lib/python3.12/site-packages/pynicotine/gtkgui/application.py", line 1011, in on_process_thread_events
    return events.process_thread_events()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/snap/nicotine-plus/2064/lib/python3.12/site-packages/pynicotine/events.py", line 338, in process_thread_events
    self.emit(event.event_name, *event.args, **event.kwargs)
  File "/snap/nicotine-plus/2064/lib/python3.12/site-packages/pynicotine/events.py", line 281, in emit
    raise error
  File "/snap/nicotine-plus/2064/lib/python3.12/site-packages/pynicotine/events.py", line 273, in emit
    function(*args, **kwargs)
  File "/snap/nicotine-plus/2064/lib/python3.12/site-packages/pynicotine/downloads.py", line 1339, in _file_download_progress
    self._update_transfer_progress(
  File "/snap/nicotine-plus/2064/lib/python3.12/site-packages/pynicotine/transfers.py", line 441, in _update_transfer_progress
    transfer.time_elapsed = time_elapsed = (time.monotonic() - transfer.start_time)
                                            ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~



### Expected behavior
No errors

### Steps to reproduce the bug
Close, and reopen the application and wait for undetermined amount of time

slook:
Regression since 3246f41067 `start_time` is unset because `_file_transfer_init()` isn't yet invoked in the "Download folder error" case.

mathiascode:
I think the problem is `sock` is set too late on the transfer object in `_file_transfer_init()`, so it's not closed when the download folder error happens.

slook:
`download.sock` can be something different from `msg.sock` so maybe the other `token` for that same `username` was closed.

slook:
[...]

PR #3878, opened 2026-09-01
test_downloads.py: verify that progress does not raise TypeError

slook:
Regression test for PR #3804 replicates crash reported in issue #3877.

```
$ python3 -m unittest -v
...
Verify that progress does not raise TypeError. ... ERROR
...

======================================================================
ERROR: test_download_progress_without_connecting (pynicotine.tests.unit.transfers.test_downloads.DownloadsTest.test_download_progress_without_connecting)
Verify that progress does not raise TypeError.
----------------------------------------------------------------------
  File "/home/user/Git/slook/nicotine-plus/pynicotine/tests/unit/transfers/test_downloads.py", line 353, in test_download_progress_without_connecting
    core.downloads._file_download_progress(username, token, bytes_left=0)
  File "/home/user/Git/slook/nicotine-plus/pynicotine/downloads.py", line 1339, in _file_download_progress
    self._update_transfer_progress(
  File "/home/user/Git/slook/nicotine-plus/pynicotine/transfers.py", line 441, in _update_transfer_progress
    transfer.time_elapsed = time_elapsed = (time.monotonic() - transfer.start_time)
                                            ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'

----------------------------------------------------------------------
Ran 68 tests in 14.721s

FAILED (errors=1, skipped=1)
Error: Process completed with exit code 1.
```

### DIFF
--- a/pynicotine/tests/unit/transfers/test_downloads.py
+++ b/pynicotine/tests/unit/transfers/test_downloads.py
@@ -8,6 +8,9 @@ from unittest import TestCase
 
 from pynicotine.config import config
 from pynicotine.core import core
+from pynicotine.slskmessages import increment_token
+from pynicotine.slskmessages import initial_token
+from pynicotine.transfers import Transfer
 from pynicotine.transfers import TransferStatus
 from pynicotine.userbrowse import BrowsedUser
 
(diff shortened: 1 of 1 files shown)


mathiascode:
It would be nice to have some file transfer tests that involve the networking thread eventually, because the interaction between the transfer manager and networking thread is probably where most things can go subtly wrong.

slook:
I agree. However the problem with offline unit testing of individual methods is that it seems like they would be too artificially constructed to be of any real world value. Really the only thing it proves is that the caller's arguments are valid, which of course is breaks every time something is changed.

The more realistic solution is end-to-end testing against an actual server instance, such as soulseek-rs does, see:
- https://github.com/michel/soulseek-rs/blob/master/soulseek-rs-lib/tests/e2e.rs
- https://github.com/michel/soulseek-rs/blob/master/soulseek-rs/tests/cli_e2e.rs
- https://github.com/michel/soulseek-rs/blob/master/.github/workflows/ci.yml#L74-L129
- https://github.com/michel/soulseek-rs/actions/runs/33738460801/job/100594536904

The advantage that CI suite has is that multiple client instances can be tested simultaneously, which would be useful for testing queue candidate selection or speed limiting behaviours and so on, and it also means you can do interesting things such as https://github.com/michel/soulseek-rs/tree/master/benchmarks, which is how I discovered a bug in the server's handling of backlogged login attempts which is addressed by PR https://github.com/soulfind-dev/soulfind/pull/104, since I probably would never have known about that problem without such a testing suite highlighting the issue.

mathiascode:
For sure. One thing I like when working on tinytag is that I can refactor the internals as much as I want, and the tests just work. It would be great to get as close to that as possible in Nicotine+ and Soulfind, and only test through public functions/methods if we can.

slook:
The beauty of the tinytag CI suite is that it's based on real world input data of `samples`, and the outputs are known data which are deterministic. That is to say, the end results are either correct or they're not, or in other words the program either works properly or it doesn't, regardless of what goes on internally.

That's a good idea actually, since you already have shares library right there in the tinytag CI suite so why not just use those data in the samples folder for testing N+ transfers as well. Another advantage of orchestrating such a runner would be that you could run checksums on all the downloads to make sure that the files stay intact in the face of whatever combination of simulated failure modes like abort/retry, dropped connections and so on.

I'd also want to test what happens when files are mutated mid-transfer as well though, and tinytag can't do that, but it would be possible to rename files for swapping them out half way or directly edit the ID3 tags and file content with a script between queue attempts and during active F connections to simulate all the kinds of misbehavior that many of the experimental clients that have recently been slopped out into the wild seem to exhibit nowadays.

> only test through public functions/methods if we can.

You could do it that way, but since the pynicotine package was never really designed to be a callable API it might be sufficient to simply populate `uploads.json` and `downloads.json` with the test cases, and just let each headless instance of the program do what it normally does when it is launched then terminate them upon reaching an end result as measured by the contents (or lack thereof) in the Incomplete/Finished Downloads folders, since that is the final output that is actually sought after for test validation purposes, rather than depending on any particular outputs produced by the program itself.

slook:
Closing, as discussed the program will be testing the unittest moreso than the other way around.

PR #3884, opened 2026-09-04
slskproto.py: check root_username in DistribBranchRoot (D code 5)

slook:
+ Added: Validate the `root_username` string reported by a peer ~is a valid Soulseek username~ before we consider trying them as a possible parent, so that there can never be any risk of us later forwarding a faulty raw payload onto all our children.
+ ~Changed: A name of a class constant to avoid ambiguity about what usernames are actually "accepted" on the network.~
+ Changed: Slight adjustments to wording of Conn log lines ~and added explanatory text to the class docstring.~

This matches the same sort of checking that's done for PeerInit, ~but stricter~ because a remote possibility does exist that this value can end up being published to our children on the distributed network, whose reasonable expectation is that this string would be validated by us as their parent, whom they are likely assume to be a trustworthy source of such pivotal information.

It is warranted to check the peer input of this message because it is only a single string with no codepoint or check bit, so any arbitrary data could end up being put there by a faulty peer, or even a foreign entity from an alien network.

## For a ridiculous example:

Lets say you are having a bad day and plugged your router ports in the wrong way around and in doing so you accidentally connected a Napster server into the Soulseek distributed network by mistake, as you do, then according to the documentation for message Type `5` of that foreign protocol...

```markdown
**Client-Server protocol**

Napster uses TCP for client to server communication.  Typically the servers
run on ports 8888 and 7777.  Note that this is different from the `metaserver'
(or redirector) which runs on port 8875.

Each message to/from the server is in the form of <length><type><data>
where <length> and <type> are 2 bytes each.  <length> specifies the length in
bytes of the <data> portion of the message.

Note that unlike the IRC protocol, each line does NOT end in \r\n.  The
<length> field specifies exactly how much data you should read.

[You may ask yourself, what could possibly go wrong? -ed]

**Message Types**

The following section describes the format of the <data> section for each
specific message type.  Each field is denoted with <>.  The fields in a
message are separated by a single space character (ASCII 32).  Where
appropriate, examples of the <data> section for each message are given.

...
5       "auto-upgrade" [SERVER]

        Format: <version> <hostname:filename>

        Napster is out of date, get a new version.
        Or also known as gaping security hole.

        <version>  = string, the new version number.
        <hostname> = string, hostname of upgrade (http) server
        <filename> = string, filename

        Connections are always made to port 80.

        The HTTP Request:
                GET <filename> HTTP/1.0
                Connection: Keep-Alive
                Host: <hostname>

        Expected HTTP Response.
                Content-Length: <size>
                Content-Type: <doesn't matter>
                <data>

        Upgrade file is save as "upgrade.exe".
        And executed as: upgrade.exe UPGRADE "<current.exe>"

        As far as I can tell no confirmation is requested by Napster when it
        receives this message.  And immediately start to "auto-upgrade". To keep
        users informed that Napster is doing something potentially very harmful
        to their computer it displays a message saying it's "auto-upgrading".

        I think this pretty bad, since all someone has to do is hack a napster
        server et voila a zillion clients at your disposal.

        As far as I known this only affects the Win32 2.0b5a Napster client.
        Other clients/versions I don't know.

	[This section was contributed by Thomas van der Heijden <thom@bart.nl> -ed]
...
```

... all their payloads are crudely space-delimited, so such a horrendous misconnection event could theoretically result in our root user's name being stored and forwarded as:
- A UTF or Latin-1 string-encoded version number of an obscure client software package, followed by space,
- followed by an URL to page hosting an executable file that can be downloaded from some random place on the internet,
- followed by a colon, followed by the file name of that server's Napster installer ("upgrade.exe").
 
This would almost certainly lead to undefined behaviour. For instance, unless it happened to be a shortened URL this string probably wouldn't resolve to a valid Soulseek username and it would be very unlikely, although not impossible, to match the name of a registered account.

Thankfully, slsk isn't vulnerable to its clients automatically executing arbitrary executable binary payloads such as what routinely happens on naps, but nevertheless this PR reduces the potential maximum size our forwarded payload outputs from ~16000 UTF or Latin-1 characters down to ~30 ASCII characters~ 256 bytes or fewer, thereby substantially reducing the chances of us announcing to everyone the we adopted an alien parent who told us that our new distributed network branch root is a 32-bit Windows executable, or any other kinds of unbounded string payload encoding mishap of a similarly ridiculous nature, are now rejected.

### DIFF
--- a/pynicotine/slskmessages.py
+++ b/pynicotine/slskmessages.py
@@ -4156,17 +4156,19 @@ class DistribBranchRoot(DistribMessage):
     message regardless of our status in the distributed network.
     """
 
-    __slots__ = ("root_username",)
+    __slots__ = ("root_username", "root_username_size")
 
     def __init__(self, root_username=None, *, msg_content=None):
         DistribMessage.__init__(self, msg_content)
         self.root_username = root_username
+        self.root_username_size = None  # in bytes
 
     def make_network_message(self):
         return self.pack_string(self.root_username)
 
     def parse_network_message(self):
         self.root_username = self.unpack_string()
+        self.root_username_size = self._offset - 4
 
 
 class DistribChildDepth(DistribMessage):
--- a/pynicotine/slskproto.py
+++ b/pynicotine/slskproto.py
@@ -333,9 +333,10 @@ class NetworkThread(Thread):
     Soulseek server and peers. Communication with the core is done through
     events.
 
-    The server and peers send each other small binary messages that
-    start with length and message code followed by the actual message
-    data.
+    The server and peers send each other small binary messages that start with
+    length and message code followed by the actual message data payload. Each
+    client should be tolerant of receiving bad inputs, but must be strict about
+    only sending valid outputs, with respect to clients that are less capable.
     """
 
     CONNECTION_INIT_TIMEOUT = 2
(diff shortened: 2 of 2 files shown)


mathiascode:
I would prefer to use the same username check we do in PeerInit.

PR #3887, opened 2026-09-06
mainwindow.py: OK button in Soulseek Announcement message dialog

slook:
"Close" doesn't exit the application, nor should it, but in context with the "going offline" wording in this particular announcement one could wrongly think that the entire session is going to be ending on pressing the button...
image

+ Changed: "OK" makes more sense here, and the suggested action highlight makes it look more prominent too:
image

+ ~Added: Also include the `title` within logged message line:~

    - ~Status bar...~
`Soulseek Announcement: [Test] Server Message Code 66.`

    - ~Log pane...~
`2026-09-06 16:20:50 Soulseek Announcement: [Test] Server Message Code 66.`

    - ~Console...~
`[2026-09-06 16:20:50] Soulseek Announcement: [Test] Server Message Code 66.`

### DIFF
--- a/pynicotine/gtkgui/mainwindow.py
+++ b/pynicotine/gtkgui/mainwindow.py
@@ -1164,7 +1164,9 @@ class MainWindow(Window):
     def update_log(self, timestamp_format, msg, title, level):
 
         if title:
-            MessageDialog(application=self.application, title=title, message=msg, selectable=True).present()
+            MessageDialog(
+                application=self.application, title=title, message=msg, buttons=[("ok", _("_OK"))], selectable=True
+            ).present()
 
         # Keep verbose debug messages out of statusbar to make it more useful
         if level not in {"transfer", "connection", "message", "miscellaneous"}:


mathiascode:
Unsure about using OK as the label, since it would also end up being used by some other dialogs, and I feel it's more common if there's an action to apply (save or similar).

The title being omitted was intentional. The idea was that the actual log message would be complete, and the title would be used for environments (e.g. dialogs) that need some more structure.

I think it would be better to prepend a string before the announcement, something like "Announcement from the Soulseek server on <date>: <announcement>". That would also make it harder to send arbitrary, confusing text.

slook:
> Unsure about using OK as the label

What else would suit, perhaps "Dismiss", "Continue" or something like that?

> I feel it's more common if there's an action to apply

Considering adding a "Don't show again for the rest of this session" checkbox option, because this might be a good idea just in case announcements ever start popping up very frequently for some reason.

I'm happy to revert the log title, it's mainly the button label I'm bothered about.

PR #3891, opened 2026-09-12
Add more warnings before installing/enabling third-party plugins

mathiascode:
Make it clear that plugins can be malicious.


mathiascode:
Backporting this to 3.3.x with English strings, because there are currently no warnings there, and people are becoming way too comfortable installing random plugins on the internet.

PR #3896, opened 2026-09-18
CI: pin debian packaging runner to ubuntu-24.04 image

slook:
To avoid **mk-origtargz error** that shall be encountered in `ubuntu-26.04` which is `ubuntu-latest` soon, see runner output...

_https://github.com/slook/nicotine-plus/actions/runs/35291192688/job/105436654535#step:4:889_
```
Creating tar archive
removing 'nicotine_plus-3.4.0.dev2' (and everything under it)
Successfully built nicotine_plus-3.4.0.dev2.tar.gz
mk-origtargz error: Could not read dist/nicotine-plus-*.tar.gz: No such file or directory
Error: Process completed with exit code 1.
##[debug]Finishing: Build Debian package
```

Unknown why different behaviour of [mk-origtargz](https://manpages.ubuntu.com/manpages.gz/resolute/man1/mk-origtargz.1.gz) command in [devscripts](https://launchpad.net/ubuntu/+source/devscripts) package;
- Provided by: [devscripts (Version: 2.23.7)](https://launchpad.net/ubuntu/noble/+package/devscripts) (Noble) https://launchpad.net/ubuntu/+source/devscripts/2.23.7
- Provided by: [devscripts (Version: 2.26.7)](https://launchpad.net/ubuntu/resolute/+package/devscripts) (Resolute) https://launchpad.net/ubuntu/+source/devscripts/2.26.7

The `ubuntu-24.04` (Noble LTS) image is supported until May 2029.


mathiascode:
Looks like it's the difference in the file name (`nicotine-plus` with hyphen vs `nicotine_plus` with underscore). I forget where it was mentioned, but the switch to underscores happened somewhere in the Python tooling in recent years.

mathiascode:
> Support [PEP 625](https://peps.python.org/pep-0625/) by canonicalizing package name and version in filenames per [the spec](https://packaging.python.org/en/latest/specifications/source-distribution-format/#source-distribution-file-name). Projects whose names contain uppercase characters, dashes, or periods will now see their sdist names normalized to match the standard and the format previously seen in wheels. For example:
> 
>     zope.interface -> zope_interface
> 
>     CherryPy -> cherrypy
> 
>     foo-bar_baz -> foo_bar_baz
> 
> Projects are encouraged to adopt this change to align with standards and other backend build systems. ([#3593](https://github.com/pypa/setuptools/issues/3593))

https://setuptools.pypa.io/en/latest/history.html#v69-3-0

slook:
I see why now thank you.

Well then I suggest we still pin only the `3.3.x` branch to use `ubuntu-24.04` (in case there might be other legacy semantics for backwards compatibility), and keep using the `ubuntu-latest` runner so the master branch is migrated changing the command argument to `dist/nicotine*.tar.gz` or simply `dist/*.tar.gz` what do you reckon?

mathiascode:
`dist/nicotine*.tar.gz` sounds fine to me.

PR #3899, opened 2026-09-19
CI [3.3.x]: packaging runners pinned to ubuntu-24.04 image

slook:
`3.3.x` branch. Keeps everything working the same after `ubuntu-latest` becomes `ubuntu-26.04` to help avoid any surprises, such as the mk-origtargz error noticed in PR #3896, in case there is a future need to make another release in the 3.3.x series.

Except the `python-dist` job already runs with the 26.04 image, so the source build runner will be available until May 2032.

The `ubuntu-24.04` (Noble LTS) image is supported until May 2029.


mathiascode:
I'll merge this for now, but I may or may not upgrade some of these to 26.04 at some point. I've found that having CI changes drift too much between the master and 3.3.x branches makes it more difficult to test and remember things.

PR #3902, opened 2026-09-20
CI: migrate debian packaging runner to ubuntu-26.04 image

slook:
PEP 625 requires normalizing package name without-dashes in sdist filename:
https://setuptools.pypa.io/en/latest/history.html#v69-3-0
https://peps.python.org/pep-0625/

This implements the solution for `master` branch agreed upon in unmerged PR #3896 also see discussion https://github.com/nicotine-plus/nicotine-plus/discussions/1537#discussioncomment-18489020 to prepare in advance for [ubuntu-latest is migrating to Ubuntu 26.04](https://github.blog/changelog/2026-09-17-ubuntu-26-generally-available-and-latest-migration/#ubuntu-latest-is-migrating-to-ubuntu-26-04) from Oct 2026. It is thereafter supported until May 2031.

The `3.3.x` branch would also need this fix if it's ever desired to migrate its CI is from `ubuntu-24.04` has pinned in PR #3899 to evade breakage due to **mk-origtargz error**, but doing so would cause the source tarball filename to change ("nicotine-plus" with hyphen vs "nicotine_plus" with underscore) yet current stable EOL isn't until May 2029.


mathiascode:
Remember to update the instructions in PACKAGING.md as well.

